PR #12, opened 2015-01-19
EclSimulator: make compile with dune-alugrid

dr-robertk:
This patch makes the ebos simulator compile with ALUGrid (and by that assure that the code still follows the DUNE grid interface). 
For that reasons the identification to the Cartesian cell id has been moved to the grid manager. 
Some minor other fixes have been made to make the code compile. 
Also, I added the FindZOLTAN.cmake check but for some reasons it is not working correctly yet. 

The switch between CpGrid and ALUGrid needs to be put into the ewoms way by @andlaus.

### DIFF
--- a/applications/ebos/eclgridmanager.hh
+++ b/applications/ebos/eclgridmanager.hh
@@ -266,6 +266,29 @@ public:
 #endif
     }
 
+    /*!
+     * \brief Extract Cartesian index triplet (i,j,k) of an active cell.
+     *
+     * \param [in]   c   active cell index.
+     * \param [out] ijk  Cartesian index triplet
+    */
+    void getIJK(const int c, std::array<int,3>& ijk) const
+    {
+        assert( c < int(cartesianCellId().size()) );
+        int gc = cartesianCellId()[ c ];
+        ijk[0] = gc % cartesianSize_[0];  gc /= cartesianSize_[0];
+        ijk[1] = gc % cartesianSize_[1];
+        ijk[2] = gc / cartesianSize_[1];
+
+#if not defined NDEBUG && HAVE_DUNE_ALUGRID == 0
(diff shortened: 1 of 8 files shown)


andlaus:
that's good stuff, thanks. The only issue minor I've found is that the zoltan cmake module and the corresponding change to `FindAlugrid.cmake` should be split out to a separate patch and added to opm-core as well. Having said this, I have no reservation to merge this PR this as-is and do the opm-core build system changes for you (whatever if you prefer)...

dr-robertk:
I'll do the split. Why should one add this patch to the core?

dr-robertk:
I merged the Zoltan Patch into the other one.  Was that it?

dr-robertk:
I found a problem with the getIJK method. I needed to fix this before this is merged.

andlaus:
> Why should one add this patch to the core? 

that's basically an ideosyncrasy of the build system: since everything needs to be copied to every module, it is easier to have a "central" module which is used for the syncronization. this central module happens to be opm-core.

> I found a problem with the getIJK method. I needed to fix this before this is merged.

ok, no problem. Just state when I should push the Big Green Guy.

andlaus:
> I merged the Zoltan Patch into the other one. Was that it?

in principle, yes. (If you mean to include all other changes to the `cmake` directory in this patch, too.) But it seems like something went wrong, as to me it seems like the first patch of this PR still contains build system and non-buildsystem changes...

dr-robertk:
I fixed the getIJK issue. ALUGrid does not run yet, but CpGrid still runs ;-)

dr-robertk:
Now it also runs with ALUGrid.

andlaus:
cool. in it goes, thanks.

PR #13, opened 2015-01-21
In dune 2.4 and later it will be necessary to avoid iterator dereferencing.

dr-robertk:
The reason is that entities will be copied. To be on the save side, we just dereference the iterator once. 
This PR fixed some of the occurrences, but probably not all of them.


andlaus:
thanks. looks good and works as advertised on my machine. merging.

PR #15, opened 2015-03-19
External cmake

joakim-hove:
This is not properly tested - but follows exactly the same pattern as for the other modules.


andlaus:
if `make ctests && ctest` works, you have my blessing to push The Green One...

PR #16, opened 2015-04-16
update debian packaging

akva2:
- this changes to package using the cmake buildsystem
- explicitly disables tests during packaging as the 'make test' target fails

if merged, please backport to release branch.


andlaus:
thanks

PR #17, opened 2015-04-24
update redhat packaging

akva2:


andlaus:
thx. do you need an RC4 tag after the merge?

akva2:
nah i have tested with these files and are not bothering with pushing builds public before final.

andlaus:
ok

PR #19, opened 2015-05-21
add the eWoms specific cmake files

andlaus:
this is a demonstration of the federated builds feature (OPM/opm-cmake#1)


andlaus:
OPM/opm-common#1 has been closed without being merged long ago, so let's close this here, too.

PR #21, opened 2015-05-27
changed: use config-mode

akva2:
- use config-mode for opm-cmake system installs
- look for opm-cmake in sibling directory


andlaus:
thanks. merging.

PR #24, opened 2015-07-27
Use ParseMode argument.

joakim-hove:
Followup to: https://github.com/OPM/opm-parser/pull/527


andlaus:
Looks good. Once you merge OPM/opm-parser#527, please also merge this one.

PR #26, opened 2015-08-05
Ewoms can use dune fem.

dr-robertk:
This PR adds dune-fem and thus adaptive components to ewoms. 
This code also compiles with CpGrid, as the necessary compatibility layer has been added in 
ewoms/aux/compatibilty.hh 

The code is a clone of andlaus/implement_adaptivity merged with the current master.


dr-robertk:
I added some more fixes that now make all tests compile again, with or without dune-fem.

dr-robertk:
After the merge with andlaus/grid_adaptivity_fixes I forgot to update opm-material. Now everything should compile with the newest master.

dr-robertk:
I reimplemented FingerProblem such that PersistentContainer is used. For that reasons I needed to implement a method entity( dofIdx ) on the stencil class that returns the corresponding entities for a dof index. For EcFV this is simply the elements and it was already available, for VcFV I added the method returning the corresponding vertex entities. Both finger problems seem to work.

dr-robertk:
Everything seems to work now, so I think it could be merged.

andlaus:
in general it looks good, but with this branch it seems like it does not compile anymore if dune-fem is unavailable. (also, I have some troubles testing it with dune-fem because the dune-fem master expects Dune 3.0-git and I've only got Dune 2.4 because I need to use Dune's autotools build system.)

dr-robertk:
Please use the dune-fem releases/2.4 branch. I will fix the other issues.

dr-robertk:
The above mentioned issue is fixed now.

andlaus:
ok, I tested this (only without dune-fem because of the build system issues) and did not find any regressions. merging; thanks for the work!

PR #27, opened 2015-09-08
changed: sync with renaming to opm-common

akva2:
*Do not merge* before renaming with friends has happened.


andlaus:
whoever merges OPM/opm-common#35, please merge this one too!

andlaus:
`dune.module` and `configure` also need to be changed. also, I suppose it would be a good idea to add `opm-common` to the prereqs because in the future there is c++ code to be added to it. do you want to do this yourself, or do you prefer if I do it?

akva2:
i welcome all help.

atgeirr:
> I suppose it would be a good idea to add opm-common to the prereqs because in the future there is c++ code to be added to it

It should be done, but in the spirit of taking small steps and making sure they are successful, should we not wait until the dust settles on this change first?

akva2:
yes, step by step.

who use these damn configure scripts anyways? this is cmake, not autotools. learn, adapt...

PR #34, opened 2015-10-20
update debian packaging

akva2:


debian/control
```diff
@@ -10,7 +10,7 @@ Build-Depends: debhelper (>= 9), cmake,
  doxygen, texlive-latex-extra, texlive-latex-recommended, ghostscript,
  libboost-filesystem-dev, libboost-system-dev, libboost-date-time-dev,
  libboost-test-dev, libsuperlu3-dev, gfortran, libsuitesparse-dev, libert.ecl-dev,
- libopm-parser-dev, libboost-iostreams-dev
+ libopm-parser-dev, libboost-iostreams-dev, libopm-common-dev
```
atgeirr: Does this not depend on opm-material?

akva2: yes it does, and it always did (or at least it did in 2015.04)

atgeirr: Again, sorry for missing that it was already there (misunderstood order etc.)

andlaus:
seems fine. thanks for your work!

PR #36, opened 2015-10-20
EclEquilInitializer: make it work for ALUGrid.

dr-robertk:
This PR fixes the EQUIL initialization for ALUGrid. However, there seems to be a problem with space filling curve ordering. That is either an ALUGrid problem or a problem of initStateEquil.

This was tested with SPE9_CP.DATA

### DIFF
--- a/applications/ebos/eclequilinitializer.hh
+++ b/applications/ebos/eclequilinitializer.hh
@@ -76,7 +76,7 @@ public:
         : simulator_(simulator)
     {
         const auto& gridManager = simulator.gridManager();
-        const auto& grid = gridManager.grid();
+        const auto& equilGrid   = gridManager.equilGrid();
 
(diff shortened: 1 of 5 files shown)


dr-robertk:
I just checked, the problem is the reordering of cells which destroys the one to one relation between the CpGrid and ALUGrid. I will fix this in a little bit. For now it works as is.

andlaus:
I'm currently working on a more generic (albeit less efficient) solution which also works for parallel runs and arbitrary cell ordering. I hope to get this into a working condition soon.

dr-robertk:
I have pushed an update that includes the general initialization of PolyhedralGrid given a DUNE grid.
Please take a look. It works but the Eclwriter does not yet respect the cell reordering.

dr-robertk:
This PR is superseded by #37.

PR #38, opened 2015-10-23
Bump version and unify format of dune.module.

atgeirr:
After this, all code modules have the same formatting in dune.module, and the correct version name (for the master branch): "2016.04-pre".

### DIFF
--- a/dune.module
+++ b/dune.module
@@ -1,15 +1,17 @@
-################################ 
-# Dune module information file. This file gets parsed by dunecontrol
+####################################################################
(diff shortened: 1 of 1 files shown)


andlaus:
thanks, but I pushed something similar half an hour ago. If I forgot something there, please reopen this PR.

atgeirr:
> If I forgot something there

You probably did not, but consider a modification since this PR would make the formatting identical across OPM (order of fields etc, initial comment etc.) which helps a lot when trying to do things more or less automated.

andlaus:
ok, done.

PR #39, opened 2015-10-26
ebos: fix issues with parallelization

dr-robertk:
This PR fixes a couple of issues with parallelization. Now, at least SPE1 can be run in parallel. So we can finally start investigating the parallelization of the wells.

### DIFF
--- a/applications/ebos/eclproblem.hh
+++ b/applications/ebos/eclproblem.hh
@@ -421,8 +421,6 @@ public:
         auto& linearizer = this->model().linearizer();
         int episodeIdx = simulator.episodeIndex();
 
-        std::cout << "Episode " << episodeIdx + 1 << " finished.\n";
-
         bool wellsWillChange = wellManager_.wellsChanged(eclState, episodeIdx + 1);
         linearizer.setLinearizationReusable(!wellsWillChange);
 
--- a/applications/ebos/eclwellmanager.hh
+++ b/applications/ebos/eclwellmanager.hh
@@ -91,14 +91,26 @@ public:
     {
         const auto &deckSchedule = eclState->getSchedule();
 
+        typedef std::array<int, Grid::dimension> CartesianCoordinate;
+
+        // vector containing wells used for numerics
+        // this list can differ from the overall number of wells
+        // due to partitioning issues
+        std::map< const std::string, std::pair< std::shared_ptr<Well>, CartesianCoordinate> > wells;
+
         // create the wells
-        for (size_t deckWellIdx = 0; deckWellIdx < deckSchedule->numWells(); ++deckWellIdx) {
+        for (size_t deckWellIdx = 0; deckWellIdx < deckSchedule->numWells(); ++deckWellIdx)
+        {
             Opm::WellConstPtr deckWell = deckSchedule->getWells()[deckWellIdx];
             const std::string &wellName = deckWell->name();
+            CartesianCoordinate cartesianCoord;
+            cartesianCoord.fill( 0 );
+            cartesianCoord[ 0 ] = deckWell->getHeadI();
+            cartesianCoord[ 1 ] = deckWell->getHeadJ();
 
             std::shared_ptr<Well> well(new Well(simulator_));
-            wellNameToIndex_[wellName] = wells_.size();
-            wells_.push_back(well);
+            // insert well into well map
+            wells[ wellName ] = std::make_pair( well, cartesianCoord );
 
             // set the name of the well but not much else. (i.e., if it is not completed,
             // the well primarily serves as a placeholder.) The big rest of the well is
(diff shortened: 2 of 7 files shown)


applications/ebos/eclproblem.hh
```diff
@@ -421,7 +421,8 @@ public:
         auto& linearizer = this->model().linearizer();
         int episodeIdx = simulator.episodeIndex();
 
-        std::cout << "Episode " << episodeIdx + 1 << " finished.\n";
+        //if( simulator.gridManager().grid().comm().rank() == 0 )
+        //    std::cout << "Episode " << episodeIdx + 1 << " finished.\n";
```
andlaus: please remove that line completely (or add the `if`, whatever you prefer)

applications/ebos/eclwellmanager.hh
```diff
@@ -693,10 +756,16 @@ protected:
...
             const std::string& wellName = deckWell->name();
 
-            if (!hasWell(wellName)) {
-                std::cout << "Well '" << wellName << "' suddenly appears in the completions "
-                          << "for the report step, but has not been previously specified. "
-                          << "Ignoring.\n";
+            if (!hasWell(wellName))
+            {
+#ifndef NDEBUG
+                if( simulator_.gridManager().grid().comm().size() == 1 )
```
andlaus: did you mean`rank()==0`?

ewoms/common/cartesianindexmapper.hh
```diff
@@ -200,18 +200,23 @@ namespace Dune
...
+          if( dimension == 3 )
           {
-              for( int d=0; d<dimension-2; ++d )
-              {
-                coords[d] = gc % cartesianDimensions()[d];  gc /= cartesianDimensions()[d];
-              }
-
-              coords[dimension-2] = gc % cartesianDimensions()[dimension-2];
-              coords[dimension-1] = gc / cartesianDimensions()[dimension-1];
+             coords[0] = gc % cartesianDimensions()[0];  gc /= cartesianDimensions()[0];
```
andlaus: two lines here, please.

ewoms/disc/common/fvbasediscretization.hh
```diff
@@ -402,7 +402,7 @@ public:
                                                               asImp_().dofMapper());
         gridView_.communicate(*sumHandle,
                               Dune::InteriorBorder_All_Interface,
-                              Dune::BackwardCommunication);
+                              Dune::ForwardCommunication);
```
andlaus: hm, not sure if `ForwardCommunication` is right here: it is supposed to add the parts of a DOF which are located on a different rank. it does not matter for ECFV and VCFV, though. (the reasons are different, though.)

andlaus:
I commented uppon a few nitpicks, but in general this looks good.

applications/ebos/eclwellmanager.hh
dr-robertk: no, I meant size == 1 ;-) This comment makes sense in serial, but we need to improve the wells distribution anyway.

ewoms/disc/common/fvbasediscretization.hh
dr-robertk: BackwardCommunication is wrong, because for dune-cornerpoint and alugrid it communicates from ghosts to interior, when you want to communicate from interior (where the volume was computed) to ghost.

dr-robertk:
Any white spaces by chance ;-)?

dr-robertk:
I think it would be good if could run the tests. I'm not sure if the communication bug creates problems for the vcvf scheme.

dr-robertk:
Fixed the style issues.

applications/ebos/eclwellmanager.hh
andlaus: hm, I see: in parallel there is no process which has a global view on all wells...

ewoms/disc/common/fvbasediscretization.hh
andlaus: my intention was to account the volume of the part of a DOF which is on a ghost to the rank where the DOF is in the interior or on the border. this statement works with both possibilites though, because for ECFV, it does not do anything and for VCFV it is actually bi-directional on the border vertices (and the others don't matter) ;)

andlaus: I see what may have confused you, the code above is a bit tricky: interior elements may still contain non-interior DOFs. (i.e. everything which is on the border, which is relevant for VCFV.)

andlaus:
> Any white spaces by chance ;-)?

only a missing new line, but that's because I'm too stupid to expect two statements on the same line ;)

andlaus:
> I think it would be good if could run the tests. I'm not sure if the communication bug creates problems for the vcvf scheme.

it will probably work both ways. I'll test tomorrow.

ewoms/disc/common/fvbasediscretization.hh
dr-robertk: I know. Thats why I said this needs to be tested for vcvf. But for ecfv the backward communication just copies a bunch of zeros to some interior elements. As an example, if you have interior and ghost cells, then communication pattern (InteriorBorder_All, Backward) packs everything on ghosts and sends it to everything interior on the other side. Backward really means that you flip InteriorBorder and All.  I admit that this is really confusing.

dr-robertk:
Btw, for vcfv it should not matter whether you do Backward or Forward, because what you really want to do is summing up all border entities. That's done either way.

dr-robertk:
Arrgh. I messed up the branch. wait, I have to fix this.

ewoms/disc/common/fvbasediscretization.hh
andlaus: > But for ecfv the backward communication just copies a bunch of zeros to some interior elements.

doesn't it add zeros?

dr-robertk: Ja, thats more like it. But it leaves some zeros for the ghost dofs. In eclproblem.hh the porosity is divided by dofTotalVolume (which is zero for ghosts).

andlaus: > Backward really means that you flip InteriorBorder and All. I admit that this is really confusing.

yeah, true. Since Dune 3.0 is going to come with only limited backwards compatibility, wouldn't it be feasible to change that API to something like

```c++
gridView.communicate(handle, Dune::InteriorBorder, Dune::All);
```

(i.e. specifying the "from" and "to" entities directly while skipping the direction.) the old method can probably be emulated with this...

andlaus: also, `Dune::All` should probably be better `Dune::Any`...

dr-robertk: Sounds good. We should suggest this at the DUNE bug tracker.

andlaus:
alright. I've tested that this works as advertised and that it does not regress any tests. I'm happy to merge it as-is if you think `ForwardCommunication` is less confusing than `BackwardCommunication`.

dr-robertk:
I do since it works ;-).

andlaus:
the other direction works too. want to bet a glass of beer? ;)

ewoms/disc/common/fvbasediscretization.hh
andlaus: cool, I suppose you should do this because whenever I suggested something like this on the DUNE bugtracker, it was either plainly rejected or it was a _long_ argument.

dr-robertk:
Ok, when I switch to BackwardCommunication I get:
ebos: /home/robertk/work/OPM/trunk/ewoms/ewoms/models/blackoil/blackoilintensivequantities.hh:235: void Ewoms::BlackOilIntensiveQuantities<TypeTag>::update(const ElementContext&, int, int) [with TypeTag = Ewoms::Properties::TTag::EclProblem; Ewoms::BlackOilIntensiveQuantities<TypeTag>::ElementContext = Ewoms::FvBaseElementContext<Ewoms::Properties::TTag::EclProblem>]: Assertion `std::isfinite(Toolbox::value(porosity_))' failed.

I take cash only :-D

ewoms/disc/common/fvbasediscretization.hh
andlaus: hm, something along the lines of

```c++
gridView.template communicate<Dune::InteriorBorder, Dune::Any>(handle);
```

would possibly be even better: that way grids have a chance to let the compiler produce an error message for communication types which they do not support. (on the other hand, it may be better to only throw at runtime because the method may never be called in some instances...)

andlaus:
> I take cash only :-D 

I suppose you only won a glass of beer ;)

PR #40, opened 2015-10-27
Implement storage cache

andlaus:
this is a bit hacky because it assumes that an Euler time discretization is used and that the solution for the time indices 0 and 1 are identical at the beginning of a time step. Nevertheless, it saves us from updating the intensive variables for solutions other than the one of the current iteration.

on my machine, this accelerates the linearization with ebos by about 30% for SPE9 (6.4 instead of 8.7 seconds, but the linearization still takes up the largest time chunk with about 50% of the total time spend).

### DIFF
--- a/applications/ebos/eclproblem.hh
+++ b/applications/ebos/eclproblem.hh
@@ -152,6 +152,9 @@ SET_BOOL_PROP(EclBaseProblem, EnableEclSummaryOutput, true);
 // decent speedup...
 SET_BOOL_PROP(EclBaseProblem, EnableIntensiveQuantityCache, true);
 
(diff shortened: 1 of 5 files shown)


dr-robertk:
Looks good. I'll merge, when I confirm that all test are still running ;-).

andlaus:
ok, I rebased this on top of the current master and tested this more thoroughly: it turned out that the globalStorage() and the checkConservativeness() methods did not work if the  storage cache was enabled. this is now fixed, and the PR is probably ready to be merged.

andlaus:
okay, merging my own PR as I used it for a while now and I haven't found any regressions recently.

PR #44, opened 2016-01-11
Fvbasediscretization: fix double free coruption

dr-robertk:
This PR fixed the double free corruption at the end of an adaptive run. The problem was wrong appearance of variables (solution and restrictProlong).

### DIFF
--- a/ewoms/disc/common/fvbasediscretization.hh
+++ b/ewoms/disc/common/fvbasediscretization.hh
@@ -328,6 +328,7 @@ class FvBaseDiscretization
     typedef Dune::Fem::AdaptationManager<Grid, RestrictProlong  > AdaptationManager;
 #else
     typedef BlockVectorWrapper  DiscreteFunction;
+    typedef size_t              DiscreteFunctionSpace;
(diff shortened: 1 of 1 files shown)


ewoms/disc/common/fvbasediscretization.hh
```diff
@@ -328,6 +328,7 @@ class FvBaseDiscretization
     typedef Dune::Fem::AdaptationManager<Grid, RestrictProlong  > AdaptationManager;
 #else
     typedef BlockVectorWrapper  DiscreteFunction;
+    typedef size_t              DiscreteFunctionSpace;
```
andlaus: I'm confused: is this just a dummy value or is it actually used somewhere?

dr-robertk: It's use in the constructor for the solution_. There has to ba an ifdef
somewhere. I prefer it that way, because having two different lines of
initialization of the solution is not a good idea.

andlaus:
ok. merging, thanks.

PR #45, opened 2016-01-12
VtkMultiWriter: make compile with current dune-grid master.

dr-robertk:
This PR make the code compile with the current dune-grid master.


dr-robertk:
@andlaus: You email seems broken.

andlaus:
thanks. (Note that I do not currently use the dune master, so if anything breaks, I'm eager to get a fix). merging.

> @andlaus: You email seems broken. 

Yep. it was since monday (due to server problems). now it should work again, so if you send me any messages, please re-send.

PR #50, opened 2016-04-01
Jenkins (re)boot

akva2:
This is the build scripts for this module, ref OPM/opm-common#102 and OPM/opm-parser#743

This is not really a reboot since there is no ewoms build job currently. I see no reason to keep ewoms in the dark ages of software development.

### DIFF
--- a/jenkins/README.md
+++ b/jenkins/README.md
@@ -1,0 +1,25 @@
+ewoms jenkins build scripts:
+--------------------------------
+
+[B]build-ewoms.sh[/B]:
+This is a helper script which contains functions for building,
+testing and cloning ewoms and its dependencies.
+
+[B]build.sh[/B]:
+This script will build dependencies, then build ewoms and execute its tests.
+It is intended for post-merge builds of the master branch.
+
+[B]build-pr.sh[/B]:
+This script will build dependencies, then build ewoms and execute its tests.
+It inspects the $ghbPrBuildComment environmental variable to obtain a pull request
+to use for opm-common, opm-parser, opm-material, opm-core and opm-grid (defaults to master)
+and then builds $sha1 of ewoms.
+
+It is intended for pre-merge builds of pull requests.
+
+You can optionally specify a given pull request to use for opm-common, opm-parser, opm-material,
+opm-core and dune-cornerpoint through the trigger.
+The trigger line needs to contain opm-common=&lt;pull request number&gt; 
+and/or opm-parser=&lt;pull request number&gt;
+and/or opm-material=&lt;pull request number&gt; and/or opm-core=&lt;pull request number&gt;
+and/or opm-grid=&lt;pull request number&gt;.
--- a/jenkins/build-ewoms.sh
+++ b/jenkins/build-ewoms.sh
@@ -1,0 +1,73 @@
+#!/bin/bash
+
+function build_ewoms {
+  # Build ERT
+  pushd .
+  mkdir -p $WORKSPACE/deps/ert
+  cd $WORKSPACE/deps/ert
+  git init .
+  git remote add origin https://github.com/Ensembles/ert
+  git fetch --depth 1 origin $ERT:branch_to_build
+  test $? -eq 0 || exit 1
+  git checkout branch_to_build
+  popd
+
+  pushd .
+  mkdir -p serial/build-ert
+  cd serial/build-ert
+  cmake $WORKSPACE/deps/ert/devel -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=$WORKSPACE/serial/install
+  test $? -eq 0 || exit 1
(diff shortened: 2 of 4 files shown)


andlaus:
like for the opm-material PR: please merge this  yourself once you consider it to be ready.

about the test suite: the tests in ewoms are primarily meant to be integration tests, not unit tests. I suppose the reason why many of them are not run on your system is that they rely on optional dependencies like dune-alugrid, dune-fem, available quadruple precision math, etc. If want to reduce their runtime compiling them in optimized mode yields a factor 10 speedup and they can also be run in parallel via `ctest -j$NUM_CORES`. Once the these prerequisites are met, the test suite is rather okay, IMO: coverage is at > 90% according to `gcov`, and I can complete a full run of the test suite in less than three minutes on my machine...

akva2:
heh, sure, if you can own a umpteen core machine you can do that. i cannot do that on a shared ci resource. in any case;

13/56 Test #13: co2injection_flash_ni_ecfv ..........***Failed  533.61 sec

that's 9 min-ish for a single failing test on my laptop, so even if i ran in parallel it would minumum take 9 min here.

integration tests do not have to be this huge. i extensively use integration tests in my projects, but none of them run longer than 5 secs and coverage is not harmed in any way.

as for the dummy test stuff, according to your comment it 's because you want them to show up on cdash no matter. we could do this  https://github.com/akva2/opm-common/commit/7c24dcfac1f70b8e2481ade5760e64924451fac2 to shut it up a little..

andlaus:
yeah, I think you're right that for the purposes of a CI system like Jenkins, these tests should not be run. In that context, it would be cool if tags (like `run-during-ci`) could be provided, but I'm not aware of ctest supporting this...

> that's 9 min-ish for a single failing test on my laptop, so even if i ran in parallel it would minumum take 9 min here.

This is caused by a breakdown of the model for the tested scenario. I'm aware of this and I even know what it is caused by: that's a precision problem within opm-material's `NcpFlash` class. The trouble with fixing this is that it turned out to be a "whack-a-mole" issue: if you fix the solver for the `NcpFlash` solver for the eWoms flash model, it has trouble to converge for other things like the test for the peng-robinson EOS in `opm-material`. Also, this issue does not seem to appear if quadruple precision math is used. I'm planning to fix this before 2016.04, though it may be require larger changes to `NcpFlash` (nested AD-evaluations anyone?) so I might even fail to fix it in time.

> integration tests do not have to be this huge. i extensively use integration tests in my projects, but none of them run longer than 5 secs and coverage is not harmed in any way. 

true, but I'd like to them to be somewhat realistic simulations where something actually happens, i.e., I guess they need a certain grid size and they need to run for enough time steps. (BTW: IIRC the tests of opm-upscaling also take a relatively long time, don't they?)

> as for the dummy test stuff, according to your comment it 's because you want them to show up on cdash no matter. we could do this akva2/opm-common@7c24dcf to shut it up a little..

yeah, I wanted them to show up even if they were not compiled. (because I always forget that they exist if they are silently ignored.) I agree that akva2/opm-common@7c24dcf is the better way to go; the only reason I used the current hack is that I could not find another way to tell ctest "this test exists but is currently unavailable" when I wrote that macro.

andlaus:
ok, I've fixed the test failures: on my machine the full test suite now passes in less than 22 seconds in optimized mode (that requires libquadmath headers to be not available. if it is around, the longest running test takes aboout 9 minutes.)

I suppose that you can now press the green button with a good conscience...

akva2:
okay, i will get to this towards the end of the day, please don't merge before i have checked that everything runs fine after restructuring in upstreams.

akva2:
jenkins build this opm-common=109 please

akva2:
jenkins build this opm-common=109 please

akva2:
i cannot honor the request until dune-localfunctions is installed on the jenkins box. i do not have shell access and thus this is as far as can take it today. @alfbr 

otherwise it works and have been tested locally. note that the post-builder is also disabled until opm-common/109 is merged since it would bog down the machine running the quadmath enabled tests (plus it still wouldn't find dune-localfunctions). also i cannot do the neater test-skipping since we need to be compatible with cmake 2.8 (mapping an return code to a skipped test was added in 3.0.something). you might have to inspect the results manually as i cannot completely rely on the error codes (i cannot find any documentation for them so i'm just guessing that 8 = 'skipped tests but nothing failed'. thus it can in principle skip more tests than it did earlier for whatever reason. unlikely, sure, but in the name of full disclosure and my clean hands and all that.)

the skipped test handling is the reason for the additional code duplication in the helper function.

andlaus:
> i cannot honor the request until dune-localfunctions is installed on the jenkins box

That's fine with me. In any case: Thanks for the work!

akva2:
jenkins build this please

akva2:
all good. merging.

PR #51, opened 2016-04-06
Renaming dune-cornerpoint -> opm-grid.

atgeirr:
This updates code to refer to opm-grid instead of dune-cornerpoint.

Requires OPM/opm-common#104.


andlaus:
thanks. you have permission to merge when this is needed.

PR #60, opened 2016-06-20
fixed: broken array loop in jenkins script

akva2:
python, bash, c++, fortran. sometimes you get lost in the syntax jungle.


andlaus:
sound like autotools ;)

thanks.

PR #61, opened 2016-06-28
fixed: restore build of ebos after opm-parser changes

akva2:
the WellConstPtr typedef is gone


akva2:
@joakim-hove i think that was the change right (ditched the shared ptr and the typedef in the process ?)

akva2:
jenkins build this please

jokva:
Seems likely, yes.

andlaus:
thanks

andlaus:
one thing I'm wondering about: isn't there _a lot_ of potential with `struct Wells` from opm-core? (that one is located in `opm/core/wells.h`.)

jokva:
There is.

PR #64, opened 2016-08-16
add the control files for the travis CI system

andlaus:
this patch is based on the travis files from opm-material, so it may be that they work. (I've no experience with travis yet, nor do I know a way to test this besides the live system, so bear with me.)

this will not break anything outside travis, so I'll be self-merging immediately.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,29 @@
+language: cpp
+
+compiler:
+  - gcc
+
+addons:
+  apt:
+    sources:
+      - boost-latest
+      - ubuntu-toolchain-r-test
+    packages:
+      - libboost1.55-all-dev
+      - gcc-4.8
+      - g++-4.8
+      - gfortran-4.8
+      - liblapack-dev
+      - libgmp3-dev
+      - libsuitesparse-dev
+      - libeigen3-dev
+
+before_script:
+    - export CXX="g++-4.8" CC="gcc-4.8" FC="gfortran-4.8"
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -0,0 +1,29 @@
...
+      - liblapack-dev
+      - libgmp3-dev
+      - libsuitesparse-dev
+      - libeigen3-dev
+
+before_script:
+    - export CXX="g++-4.8" CC="gcc-4.8" FC="gfortran-4.8"
+    - cd ..
+    - git clone https://github.com/OPM/opm-common.git
+    - opm-common/travis/clone-opm.sh opm-material
```
joakim-hove: This should be: `opm-common/travis/clone-opm.sh ewoms`

andlaus: thanks for spotting this, i've pushed a fix (without a PR)

joakim-hove:
Observe that the general build script is `opm-common/travis/build-and-test.sh` - i.e. the `build-and-test-ewoms.sh` and `build-ewoms.sh` scripts will not be used[1].

[1]: Yes - there are some dangling files of that type in some of the other repositories, which should be removed.

andlaus:
> Observe that the general build script is opm-common/travis/build-and-test.sh - i.e. the build-and-test-ewoms.sh and build-ewoms.sh scripts will not be used[1].

I was wondering about that a bit while doing this. I'll remove these files, thanks for the hint!

PR #65, opened 2016-08-22
Simplify after ert reorg

akva2:
Depends on https://github.com/OPM/opm-common/pull/168

Note an unrelated fix for the 'check' target is included in the same PR.


akva2:
jenkins build this opm-common=168 please

andlaus:
merge at will...

PR #69, opened 2016-09-16
Make dune-localfunctions optional. This is only needed for the vcfv s…

dr-robertk:
This is only needed for the vcfv scheme.


andlaus:
damn! we duplicated our work. (I've spend the morning doing essentially the same: some things are better in your patch, some I like better in mine. I will try to bring them together.)

did you run the test suite?

dr-robertk:
I ran ebos on SPE9. I guess the cmake code for the vcfv tests needs to be adjusted.

andlaus:
ok. this should be superseeded by #70. closing it for now, thanks anyway!

PR #70, opened 2016-09-21
make dune-localfunctions optional

andlaus:
as it happens so often, this easy-in-principle change took quite a bit longer than anticipated and lead to me finding a bug in the constraints code (the constraints are only considered after the first iteration of a time step, so the Newton method must to at least two iterations if constraints are enabled) and the removal of the navier-stokes test problem (which only worked because of the aforementioned bug and even then oscillated like hell).

this PR is based on @dr-robertk's #69, but since it did not retain much of the original code, I did not want to blame @dr-robertk for any bugs so I took authorship. (@dr-robertk: if you want to be author of the last commit, let me know and I'll change it.)


andlaus:
alright: I hope I have added a patch to do the constraint DOFs properly now: constraints are now explicitly gathered at the beginning of each time step. Since this requires a separate loop over the while grid, it may be slower in some cases. on the other hand, this is done only once per time step, so the performance impact may be quite limited and it is only done if constraint DOFs are explicitly enabled, i.e. most simulations won't be affected at all...

I'll merge this (and the opm-common PR) myself once jenkins is happy...

andlaus:
jenkins build this please

andlaus:
the CI crew is happy...

PR #72, opened 2016-10-12
update redhat packaging

akva2:
- add new dependencies
- build using devtoolset-3 on redhat6


andlaus:
this only affects the package files, so merge at will.

PR #73, opened 2016-10-14
The great shared_ptr purge

jokva:
Downstream of  https://github.com/OPM/opm-parser/pull/939


andlaus:
seems good. you may merge once the upstream is in...

PR #74, opened 2016-10-17
ebos: add support for geology changes during the simulation

andlaus:
this adds support for geologic events during the simulation (e.g. pore volume changes of changing faults in the SCHEDULE section). The code currently has the same limitations as the one for flow in opm-simulators: these geologic events may only change the porosity of some cells or the values of the transmissibility, i.e., changes to the grid topology are not possible.


andlaus:
the results for model 2  do not seem to get worse by this. merging.

PR #76, opened 2016-10-25
fix build on Dune 2.3

### LINKED ISSUE #75: OPM build failed: ecfvstencil.hh:250:58: error: ‘ePtr’ was not declared
```
          ^
Linking CXX executable bin/art2dgf
[  0%] Built target art2dgf
Scanning dependencies of target co2injection_flash_ecfv
[  1%] Building CXX object CMakeFiles/co2injection_flash_ecfv.dir/tests/co2injection_flash_ecfv.cc.o
In file included from <https://ci.statoil.no/job/OPM7/ws/ewoms/ewoms/disc/ecfv/ecfvdiscretization.hh>:34:0,
                 from <https://ci.statoil.no/job/OPM7/ws/ewoms/tests/co2injection_flash_ecfv.cc>:37:
<https://ci.statoil.no/job/OPM7/ws/ewoms/ewoms/disc/ecfv/ecfvstencil.hh>: In member function ‘void Ewoms::EcfvStencil<Scalar, GridView>::updateTopology(const Element&)’:
<https://ci.statoil.no/job/OPM7/ws/ewoms/ewoms/disc/ecfv/ecfvstencil.hh>:250:58: error: ‘ePtr’ was not declared in this scope
         subControlVolumes_.emplace_back(SubControlVolume(ePtr));
                                                          ^
<https://ci.statoil.no/job/OPM7/ws/ewoms/ewoms/disc/ecfv/ecfvstencil.hh>: In member function ‘void Ewoms::EcfvStencil<Scalar, GridView>::updatePrimaryTopology(const Element&)’:
<https://ci.statoil.no/job/OPM7/ws/ewoms/ewoms/disc/ecfv/ecfvstencil.hh>:285:58: error: ‘ePtr’ was not declared in this scope
         subControlVolumes_.emplace_back(SubControlVolume(ePtr));
                                                          ^
make[2]: *** [CMakeFiles/co2injection_flash_ecfv.dir/tests/co2injection_flash_ecfv.cc.o] Error 1
make[1]: *** [CMakeFiles/co2injection_flash_ecfv.dir/all] Error 2
make: *** [all] Error 2
Build step 'Execute shell' marked build as failure
[WARNINGS] Skipping publisher since build result is FAILURE
```

andlaus:
that broke because of a silly mistake I made yesterday and I did not
test it with Dune 2.3. (maybe Dune 2.3 support should be dropped soon
anyway...)

this should fix #75. once the CI systems are green, I'll thus self-merge.

### DIFF
--- a/ewoms/disc/ecfv/ecfvstencil.hh
+++ b/ewoms/disc/ecfv/ecfvstencil.hh
@@ -241,14 +241,14 @@ public:
 #if DUNE_VERSION_NEWER(DUNE_COMMON, 2, 4)
         // add the "center" element of the stencil
         subControlVolumes_.clear();
-        subControlVolumes_.emplace_back(SubControlVolume(element));
+        subControlVolumes_.emplace_back(/*SubControlVolume(*/element/*)*/);
         elements_.clear();
         elements_.emplace_back(element);
 #else
         // add the "center" element of the stencil
         ElementPointer ePtr(element);
         subControlVolumes_.clear();
-        subControlVolumes_.emplace_back(SubControlVolume(ePtr));
+        subControlVolumes_.emplace_back(/*SubControlVolume(*/ePtr/*)*/);
         elements_.clear();
         elements_.emplace_back(ePtr);
 #endif
@@ -262,12 +262,12 @@ public:
             // degree of freedom and an internal face, else add a
             // boundary face
             if (intersection.neighbor()) {
-                elements_.emplace_back(intersection.outside());
-                subControlVolumes_.emplace_back(SubControlVolume(intersection.outside()));
-                interiorFaces_.emplace_back(SubControlVolumeFace(intersection, subControlVolumes_.size() - 1));
+                elements_.emplace_back(/*SubControlVolume(*/intersection.outside()/*)*/);
+                subControlVolumes_.emplace_back(/*SubControlVolume(*/intersection.outside()/*)*/);
+                interiorFaces_.emplace_back(/*SubControlVolumeFace(*/intersection, subControlVolumes_.size() - 1/*)*/);
             }
             else {
-                boundaryFaces_.emplace_back(SubControlVolumeFace(intersection, - 10000));
+                boundaryFaces_.emplace_back(/*SubControlVolumeFace(*/intersection, - 10000/*)*/);
             }
         }
     }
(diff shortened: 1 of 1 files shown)


andlaus:
jenkins build this please

atgeirr:
> maybe Dune 2.3 support should be dropped soon
> anyway

For the 2017.04 release we'll have to support 2.5, so it would be good to drop 2.3. However we may have users needing to stay on 2.3. I suggest we pose that question to the mailing list after the release is done.

pgdr:
Close the issue at will.

andlaus:
> I suggest we pose that question to the mailing list after the release is done.

+1. We should probably also query if somebody needs support for GCC < 4.9 (i.e., a GCC which does not support c++-2011 fully). That should allow us to drop `FindCXXFeatures.cmake` and the `boost::regex` dependency. opinions, @akva2?

andlaus:
jenkins build this please

andlaus:
jenkins build this please

andlaus:
jenkins build this please

atgeirr:
The Travis check should, I think, be sufficient to verify that this is ok and can be merged, otherwise what is checked by the jenkins run so that we must wait for it?

andlaus:
> The Travis check should, I think, be sufficient to verify that this is ok and can be merged, otherwise what is checked by the jenkins run so that we must wait for it?

true. jenkins compiled a few tests, so it should also be happy when it is finished. merging,

blattms:
The current Debian stable ships 2.3 and it is really handy being able
to use these packages. Of course people could use the backported 2.4
packages. I would suggest to leave it as is for the current release
and drop 2.3 afterwards. 

The next version  will most probably be using
DUNE 2.5, but that will not be released before spring 2017. BTW I am
currently testing with DUNE 2.5 and will provide patches if needed.

akva2:
if we are to drop gcc 4.8 support, we have to drop ubuntu trusty / use an non-standard toolchain as that is based on 4.8.5. we would also need to use an additional toolset on rhel7, as that is also on 4.8.5. rhel6 would be fine, as we already use gcc 4.9 there. so to me, it sounds a bit early.

andlaus:
ok. seems like the older versions will stick around for a while...

PR #77, opened 2016-10-25
ebos: precalculate the depth of the center of an element

andlaus:
it turns out that this is quite slow in Dune::CpGrid, so this patch caches the values. on my machine linearization for SPE9_CP went from taking 7.02 seconds to 6.46 seconds.

note that I think that this is stuff which should be merged after the 2016.10 release branches have been created. @atgeirr: if you want it in the release, you have permission to press green.


andlaus:
jenkins build this please

dr-robertk:
Wait. I have a patch for this

andlaus:
ok. superseeded by #78

PR #78, opened 2016-10-25
Use correct cellcenterdepth

dr-robertk:
This replaces the grid.cellCenterDepth method with the correct DUNE interface usage for the same computation.

### DIFF
--- a/applications/ebos/eclproblem.hh
+++ b/applications/ebos/eclproblem.hh
@@ -805,10 +805,25 @@ private:
     static bool enableEclOutput_()
     { return EWOMS_GET_PARAM(TypeTag, bool, EnableEclOutput); }
 
+    Scalar cellCenterDepth( const Element& element ) const
+    {
+        typedef typename Element :: Geometry Geometry;
+        static constexpr int zCoord = Geometry ::dimension - 1;
+        Scalar zz = 0.0;
+
+        const Geometry geometry = element.geometry();
+
+        const int corners = geometry.corners();
+        for (int i=0; i<corners; ++i)
+        {
+            zz += geometry.corner( i )[ zCoord ];
+        }
+        return zz/Scalar(corners);
+    }
+
     void updateElementDepths_()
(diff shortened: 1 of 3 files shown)


dr-robertk:
@andlaus: I have rebase on top of your PR. Making a PR for your local branch did not work.

dr-robertk:
jenkins build this please

andlaus:
good idea to do this generically. why did you close this?

dr-robertk:
Accident.

applications/ebos/eclproblem.hh
```diff
@@ -818,13 +833,14 @@ private:
         auto elemIt = gridView.template begin</*codim=*/0>();
         const auto& elemEndIt = gridView.template end</*codim=*/0>();
         for (; elemIt != elemEndIt; ++elemIt) {
+            const Element& element = *elemIt;
 #if DUNE_VERSION_NEWER(DUNE_COMMON, 2,4)
-            unsigned elemIdx = elemMapper.index(*elemIt);
+            const unsigned int elemIdx = elemMapper.index(element);
```
andlaus: I generally dislike declaring local variables and call-by-value method arguments as `const`, IMO it is just a distraction. Mind removing it? (I won't strictly require it.)

andlaus:
ah, travis ;)

dr-robertk:
@andlaus I don't mind

applications/ebos/eclproblem.hh
andlaus: well, on a personal "annoyingness scale" from 0 to 100 I would put it at about 8. (that's probably about the same score as curly brackets for each 'if' get, btw.) I just don't see any value in these qualifiers because they've never helped me to produce better code so far...

tl;dr: don't mind and do what you prefer!

andlaus:
since the release branch has been created and travis is happy, there's no reason to hold this back. thanks.

PR #81, opened 2016-10-30
[feature] Make mapper inside stencil a reference to avoid re-computation.

dr-robertk:
This PR makes the mapper inside the Stencil class a reference, to avoid re-computation of offsets etc.


andlaus:
ops, that was a bug. thanks for fixing it!

andlaus:
weirdly, I did not see any performance improvement by this patch -- if anything there was a tiny regression. (it is so small that it easily can be noise...)

PR #82, opened 2016-11-04
Fix mismatch between vector and allocator.

atgeirr:
With clang/libc++ this gives a static assertation failure.


andlaus:
I'm surprised that the current version compiles for me. thanks a lot!

PR #83, opened 2016-11-07
fix most pedantic compiler warnings in the basic infrastructure

andlaus:
i.e., using clang 3.8 to compile the test suite with the following
flags:

```
-Weverything
-Wno-documentation
-Wno-documentation-unknown-command
-Wno-c++98-compat
-Wno-c++98-compat-pedantic
-Wno-undef
-Wno-padded
-Wno-global-constructors
-Wno-exit-time-destructors
-Wno-weak-vtables
-Wno-float-equal
```

should not produce any warnings anymore. In my opinion the only flag
which would produce beneficial warnings is -Wdocumentation. This has
not been fixed in this patch because writing documentation is left for
another day (or, more likely, year).

note that this patch consists of a heavy dose of the OPM_UNUSED macro
and plenty of static_casts (to fix signedness issues). Fixing the
singedness issues were quite a nightmare and the fact that the Dune
API is quite inconsistent in that regard was not exactly helpful. :/

Finally this patch includes quite a few formatting changes (e.g., all
occurences of 'T &t' should be changed to `T& t`) and some fixes for
minor issues which I've found during the excercise.

I've made sure that all unit tests the test suite still pass
successfully and I've made sure that flow_ebos still works for Norne
and that it did not regress w.r.t. performance.

(Note that this patch does not fix compiler warnings triggered `ebos`
and `flow_ebos` but only those caused by the basic infrastructure or
the unit tests.)

### DIFF
--- a/applications/art2dgf.cc
+++ b/applications/art2dgf.cc
@@ -21,6 +21,7 @@
   copyright holders.
 */
 #include <opm/common/ErrorMacros.hpp>
+#include <opm/common/Exceptions.hpp>
 
 #include <dune/common/version.hh>
 #include <dune/common/fvector.hh>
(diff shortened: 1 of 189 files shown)


andlaus:
this is the compagnion PR to OPM/opm-material#177

atgeirr:
I still get plenty of warnings from the test suite, but this is probably a good start. I think most of my warnings come from untested !HAVE_DUNE_LOCALFUNCTIONS branches, for example:
```
/Users/atgeirr/opm-build-debug/ewoms/ewoms/disc/vcfv/p1fegradientcalculator.hh:92:40: warning: unused parameter 'elemCtx'
      [-Wunused-parameter]
    void prepare(const ElementContext& elemCtx,
```
If you can fix that file, and address the Travis failure (a typo in vcvfstencil.hh line 1148) I think this should be merged.

I'll not start any work on my own addressing warnings until this is merged.

andlaus:
> I still get plenty of warnings from the test suite, but this is probably a good start. I think most of my warnings come from untested !HAVE_DUNE_LOCALFUNCTIONS branches

right. I suppose that's a consequence of the combinatorial explosion of possibilities (I only tested with all modules installed). Anyway, it should be fixed now, thanks.

atgeirr:
There are still quite a lot of warnings, but I suspect they are caused by a limited number of code lines. I think this should be merged. (Assuming you have run tests and verified identical results with the master branch.)

andlaus:
I'm currently testing it without any optional modules, and I've found a few warnings issues if dune-alugrid is missing. (should be fixed now.) basically everything else were warnings in the ECL code.

andlaus:
> I think this should be merged.

ok, I'll just wait for travis.

andlaus:
travis has successfully compiled it. merging.

atgeirr:
After this, `make test-suite` gives me no warnings!

PR #84, opened 2016-11-10
fix a signedness issue when retrieving parameters

andlaus:
the issue only bites if the tests are compiled in debug mode, so it
has only been discovered now.


atgeirr:
Question: was this a benign warning situation, or was it a bug?

andlaus:
it was a bug: when using the parallel AMG solver with debug flags, the parameter system threw an exception because the "LinearSolverOverlap" parameter was registered as being of `int` type but queried as `unsigned`. maybe it would be better to catch such things at compile time, but I don't think that this is possible...

PR #85, opened 2016-11-11
ebos: also install the headers of the applications subdirectory

andlaus:
this should make flow_ebos work for system-installed eWoms modules.

@akva2: can you test whether this works and if yes press the Green Button?


akva2:
it would work but i don't think prefix/include/applications is very nice. i understand it has consequences but i really think prefix/include/ewoms/applications is more appropriate.

andlaus:
ok, do you think `applications` should be removed? (i.e. the headers would directly go to `ebos`)

akva2:
probably even better. my main concern is polluting prefix/include, though conflicts are unlikely it's still good to qualify the headers.

andlaus:
ok, I've removed the "applications" directory. Note that now a (pretty trivial) change is needed for flow_ebos, too.

@akva2: if you're happy with this, please press green.

andlaus:
> @akva2: if you're happy with this, please press green.

(in that case don't forget to also merge OPM/opm-simulators#908.)

akva2:
i'm happy. i guess there is no point waiting for travis.

PR #87, opened 2016-11-17
Minor fixes

andlaus:
this PR contains a few minor bugfixes concerning the explicit prefetching code, the determination of the upstream and downstream cells of fluxes and a smallish cleanup concerning the entity mapper.


atgeirr:
Changes to up/down-wind sounds quite significant, what was changed?

andlaus:
> what was changed?

this: https://github.com/OPM/ewoms/pull/87/files#diff-105eb94e1dc72082059baca5c9403c79R254

basically the upwind direction was undefined if the mobility on both sides of the intersection was zero. this sounds more scary than it actually was because in that case the flux was zero regardless of the upwind direction. (in principle the simulation could still segfault, hence this was slightly more than an annoyance.)

PR #90, opened 2016-11-22
ebos: export the ECL material law manager

andlaus:
this allows external code (i.e., flow_ebos) to reuse it.


andlaus:
since this is rather trivial and will not affect any existing code, I'll self-merge. (I consider it to be a small breach of encapsulation, though.)

PR #91, opened 2016-11-22
Fix some parallelization issues.

dr-robertk:
This PR fixes some parallelization issues with ebos. Especially the equilInit needs review.

### DIFF
--- a/ebos/collecttoiorank.hh
+++ b/ebos/collecttoiorank.hh
@@ -183,44 +183,32 @@ namespace Ewoms
                 ElementMapper;
 
             const CollectiveCommunication& comm = gridManager.grid().comm();
+            ElementMapper elemMapper(gridManager.gridView());
             {
                 std::set< int > send, recv;
                 // the I/O rank receives from all other ranks
                 if( isIORank() )
                 {
-                    const auto& eclGrid = gridManager.eclGrid();
+                    // the I/O rank needs a picture of the global grid, here we
+                    // use equilGrid which represents a view on the global grid
+                    const size_t globalSize = gridManager.equilGrid().leafGridView().size( 0 );
+                    // reserve memory
+                    globalCartesianIndex_.reserve( globalSize );
+                    globalCartesianIndex_.clear();
 
-                    // create the ACTNUM array based on the "real" grid which is used for
-                    // simulation. Note that this is still an approximation because the simulation
-                    // grid may also modify the geometry of cells (e.g. because of the PINCH
-                    // keyword), but at least the number of cells is correct, so all values are
-                    // hopefully displayed at approximately the correct location.
-                    std::vector<int> actnumData(eclGrid->getCartesianSize(), 0);
-                    ElementMapper elemMapper(gridManager.gridView());
-                    auto elemIt = gridManager.gridView().template begin<0>();
-                    const auto& elemEndIt = gridManager.gridView().template end<0>();
-                    for (; elemIt != elemEndIt; ++elemIt) {
+                    // loop over all elements (global grid) and store Cartesian index
+                    auto elemIt = gridManager.equilGrid().leafGridView().template begin<0>();
+                    const auto& elemEndIt = gridManager.equilGrid().leafGridView().template end<0>();
+                    int count = 0;
+                    for (; elemIt != elemEndIt; ++elemIt, ++count) {
 #if DUNE_VERSION_NEWER(DUNE_COMMON, 2,4)
                         int elemIdx = elemMapper.index(*elemIt );
 #else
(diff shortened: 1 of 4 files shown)


ebos/ecltransmissibility.hh
```diff
@@ -160,7 +160,9 @@ public:
                 const auto& intersection = *isIt;
 
                 // ignore boundary intersections for now (TODO?)
-                if (intersection.boundary())
+
+                // continue if no neighbor is present
+                if ( ! intersection.neighbor() )
```
andlaus: nice catch!

andlaus:
Thanks. this looks good. you have my blessings to merge this yourself once jenkins is green :)

dr-robertk:
jenkins build this please

ebos/collecttoiorank.hh
```diff
@@ -183,44 +183,32 @@ namespace Ewoms
...
-                    globalCartesianIndex_.reserve( cartesianSize );
-                    globalCartesianIndex_.clear();
-                    // get a global cartesian index for each active cell in the eclipse grid
-                    for( size_t cartIndex=0; cartIndex<cartesianSize; ++cartIndex )
-                    {
-                      if( actnumData[cartIndex] > 0 )
-                      {
-                        globalCartesianIndex_.push_back( cartIndex );
-                      }
+                        assert( count == elemIdx );
```
andlaus: this condition depends on the grid implementation, right? (i.e., the Dune grid interface does not mandate that the entities are iterated over in consecutive order of their indices, right? I suppose that nobody in their right mind would implement it differently, but still.)

ebos/eclequilinitializer.hh
```diff
@@ -107,8 +107,14 @@ public:
         // create a separate instance of the material law manager just because opm-core
         // only supports double as the type for scalars (but ebos may use float or quad)
         std::vector<int> compressedToCartesianEquilElemIdx(numEquilElems);
+        std::map<int,int> equilCartesianToCompressed;
```
andlaus: wouldn't using a 'std::vector<int>' be better here?

ebos/eclequilinitializer.hh
```diff
@@ -140,14 +145,24 @@ public:
...
+
         // copy the result into the array of initial fluid states
         initialFluidStates_.resize(numCartesianElems);
-        for (unsigned equilElemIdx = 0; equilElemIdx < numEquilElems; ++equilElemIdx) {
-            unsigned cartesianElemIdx = gridManager.equilCartesianIndex(equilElemIdx);
+        for (unsigned int elemIdx = 0; elemIdx < numElems; ++elemIdx) {
+            unsigned cartesianElemIdx = gridManager.cartesianIndex(elemIdx);
             auto& fluidState = initialFluidStates_[cartesianElemIdx];
 
+            const unsigned int equilElemIdx = localToEquilIndex[ elemIdx ];
```
andlaus: is this needed in that loop? I suppose the compiler will warn about an unused variable...

ebos/collecttoiorank.hh
dr-robertk: The assertion is simply there because this assumption is implicitly made here. So if for some reasons we catch this assertion then the code needs to be fixed.

ebos/eclequilinitializer.hh
dr-robertk: But then you would store a vector for the whole grid, the non-parallel grid. I usually try to avoid this.

blattms:
Seems like the collecttoiorank stuff should be very similar to the stuff in ParallelDebugOutput.hpp. Why another implementation here?

dr-robertk:
It is a copy and the copy was made because some things are different in both codes. I'll sort this out later. Right this only fixes certain failures in parallel runs.

ebos/eclequilinitializer.hh
andlaus: don't you to this anyway here? the mapping is from a non-parallel index to the distributed one, if I'm not mistaken. (and `std::map<int, int>` uses at least four times the memory as `std::vector<int>` on an per object  basis.)

andlaus: sorry, it's six times the amount of memory (2 64-bit pointers and 2 32-bit values instead of a single 32-bit value.)

dr-robertk:
I addressed the comments about std::map and rebased.

andlaus:
ok, I'll try to do the rest (if there's any). have a nice vacation!

dr-robertk:
Can we merge this?

andlaus:
yes. (I was just waiting for you.)

PR #94, opened 2016-11-23
Mini deck is reference, not pointer

jokva:
Downstream of https://github.com/OPM/opm-parser/pull/981


andlaus:
no objections. merge with OPM/opm-parser#981.

PR #96, opened 2016-11-29
fixed: only build ebos if opm-core was found

akva2:
condition must be explicit now, since HAVE_OPM_GRID no longer implies HAVE_OPM_CORE.


andlaus:
looks good, thanks.

PR #97, opened 2016-11-30
Use new include path.

atgeirr:
Apparently forgotten in the big move.


atgeirr:
I guess I'll close this, then...

andlaus:
sorry, I would have taken your PR instead of creating a new one. For some reason github only send me the notification email at ~14:00h, so I did not notice it in time...

atgeirr:
No problem, it's just that branches get so sad when they die unmerged :-)

PR #98, opened 2016-11-30
ebos: misc improvements

andlaus:
this contains a few minor improvements for ebos. most of them aim to make ebos more flexible if it is used in "framework mode" by external code (i.e., flow_ebos)


atgeirr:
Was it a good idea to include the missing-include-fix here? This PR requires a more and deeper testing than that simple fix...

andlaus:
ok, the include statement has been split out and merged separately.

andlaus:
thanks for looking at this @atgeirr.

I don't think there is much reason to hold this back any longer: travis is happy, and no existing functionality should have changed with the defaults. (I'm *very* sure about the latter.) merging.

PR #100, opened 2016-12-05
When using equilIdx equilMatrialLawManager has to be used.

dr-robertk:


dr-robertk:
Wait.

dr-robertk:
Ok, fixed the fix.

PR #101, opened 2016-12-05
CollectDataToIoRank: don't require the elements of a dune grid to be traversed in consecutive order

andlaus:
I'm sure that the results for sequential runs will be identical, but I'm not sure if the last three hunks are semantically correct.

### DIFF
--- a/ebos/collecttoiorank.hh
+++ b/ebos/collecttoiorank.hh
@@ -193,22 +193,19 @@ namespace Ewoms
                     // use equilGrid which represents a view on the global grid
                     const size_t globalSize = gridManager.equilGrid().leafGridView().size( 0 );
                     // reserve memory
-                    globalCartesianIndex_.reserve( globalSize );
-                    globalCartesianIndex_.clear();
+                    globalCartesianIndex_.resize(globalSize, -1);
 
                     // loop over all elements (global grid) and store Cartesian index
                     auto elemIt = gridManager.equilGrid().leafGridView().template begin<0>();
                     const auto& elemEndIt = gridManager.equilGrid().leafGridView().template end<0>();
-                    int count = 0;
-                    for (; elemIt != elemEndIt; ++elemIt, ++count) {
+                    for (; elemIt != elemEndIt; ++elemIt) {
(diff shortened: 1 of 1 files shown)


ebos/collecttoiorank.hh
```diff
@@ -193,22 +193,19 @@ namespace Ewoms
...
 #if DUNE_VERSION_NEWER(DUNE_COMMON, 2,4)
                         int elemIdx = elemMapper.index(*elemIt );
 #else
                         int elemIdx = elemMapper.map(*elemIt );
 #endif
-                        assert( count == elemIdx );
                         int cartElemIdx = gridManager.equilCartesianIndexMapper().cartesianIndex(elemIdx);
-                        globalCartesianIndex_.push_back( cartElemIdx );
+                        globalCartesianIndex_[elemIdx] = cartElemIdx;
                     }
```
dr-robertk: Looks ok.

dr-robertk:
Looks good. I'll merge.

ebos/collecttoiorank.hh
```diff
@@ -248,7 +244,7 @@ namespace Ewoms
                     // only store interior element for collection
                     if( element.partitionType() == Dune :: InteriorEntity )
                     {
-                        localIndexMap_.push_back( index );
+                        localIndexMap_[elemIdx] = elemIdx;
```
blattms: Did anybody test/run this code before merging? To the very least this looks suspicious (why would one need to store an identity map?) and seems to be a buggy change of semantics. Note that only interior entities were pushed previously (resulting in a map of entity index (by traversal) to interior entity index (by traversal). BTW Is the size of localIndexMap_ ever set?

andlaus: I ran it, albeit not parallel. I can't judge if that array is required, but I'm pretty sure that the result is the same as before if `Dune::CpGrid` is used for the grid.

blattms:
> I ran it, albeit not parallel. I can't judge if that array is required, but I'm pretty sure that the result is the same as before if `Dune::CpGrid` is used for the grid.

Look again, in a parallel run you will have non-interior entities.

andlaus:
that code is _really_ non-obvious IMO.

PR #102, opened 2016-12-05
ebos: replace the opm-parser related pointers/smart pointers by references

andlaus:
this seems to be the spirit of the season. note that now the deck object must be copied, but I suppose that copying it is pretty cheap.

this PR requires a companion for opm-simulators.


andlaus:
the companion PR is OPM/opm-simulators#967

andlaus:
succeeded by #103. closing.

PR #103, opened 2016-12-05
Transmissibility edge weights

andlaus:
this makes creating the grid a bit slower because the
transmissibilities need to be calculated twice: once for the
sequential grid and once for the distributed one. while corresponds to
the way `flow_legacy` does the load balancing and it should allow
better results, this does not seem to be the case for the Norne deck
if ZOLTAN is not available:

After loadbalancing process 3 has 4413 cells.
After loadbalancing process 2 has 12390 cells.
After loadbalancing process 0 has 13629 cells.
After loadbalancing process 1 has 21253 cells.

i.e., process 1 is responsible for about 4 times the number of cells
as process 3.

Note that this PR includes #102 (which should go in before).

### DIFF
--- a/ebos/eclbasegridmanager.hh
+++ b/ebos/eclbasegridmanager.hh
@@ -40,6 +40,7 @@
 #include <opm/parser/eclipse/EclipseState/Schedule/Schedule.hpp>
 
 #include <vector>
+#include <unordered_set>
 #include <array>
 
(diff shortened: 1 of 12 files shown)


andlaus:
the travis error is because it also tries to build the downstrean opm-simulators without applying OPM/opm-simulators#967

PR #106, opened 2016-12-06
make it possible to specify a separate floating-point type of the linear solver

andlaus:
The floating point type which ought to be used internally by the iterative linear solvers can now be specified via the "LinearSolverScalar" property. It defaults to "Scalar", so this patch does not change the default behaviour.

The most common case of this is probably to use single precision floating point scalars for the linear solver and to do the linearization using double precision. For the lens_immiscible_ecfv test problem with two grid refinements, the performance improvements seem to be considerable: On my machine the linear solver speedup is almost 50%. On the flip-side, there are precission problems, so the tolerance of the Newton-Raphson solver must be relaxed when using single-precision for the linear solver.

The exact command which I used for that measurement is:

```
./bin/lens_immiscible_ecfv --newton-raw-tolerance=1e-6 --grid-global-refinements=2
```

In the future, making the linear solver automatically switch between `float` and `double` might solve the precision problems, but I need to think a bit about how to implement this.

### DIFF
--- a/ewoms/linear/domesticoverlapfrombcrsmatrix.hh
+++ b/ewoms/linear/domesticoverlapfrombcrsmatrix.hh
@@ -49,19 +49,20 @@ namespace Linear {
  * The foreign overlap are all (row) indices which overlap with the
  * some of the current process's local indices.
  */
-template <class BCRSMatrix>
 class DomesticOverlapFromBCRSMatrix
 {
-    DomesticOverlapFromBCRSMatrix(const DomesticOverlapFromBCRSMatrix& ) = delete;
-
-    typedef Ewoms::Linear::ForeignOverlapFromBCRSMatrix<BCRSMatrix> ForeignOverlap;
+    typedef Ewoms::Linear::ForeignOverlapFromBCRSMatrix ForeignOverlap;
     typedef Ewoms::Linear::GlobalIndices<ForeignOverlap> GlobalIndices;
 
 public:
+    // overlaps should never be copied!
+    DomesticOverlapFromBCRSMatrix(const DomesticOverlapFromBCRSMatrix&) = delete;
+
     /*!
      * \brief Constructs the foreign overlap given a BCRS matrix and
      *        an initial list of border indices.
      */
+    template <class BCRSMatrix>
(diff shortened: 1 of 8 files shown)


blattms:
I have three comments:

1. Apart from #107 why do you need to account for partitions/ use mpi when converting matrices and vectors?
2. My stomach aches whenever I see people switching to float for performance. Like you noticed, there can be side effects in stability. Additionally some of the solvers are not even there for float (UMFPack). Using AMG with float will result in an iterative coarse solver (at least without DUNE master and SuperLU) being used which might reduce convergence considerably (for jumping coefficients). I do not think that using on example suffices.
3. In the same light, please comment on what iterative solver you have been using and how the number of steps of the iterative solver and setup time has changed.

andlaus:
> Apart from #107 why do you need to account for partitions/ use mpi when converting matrices and vectors?

?? (what do you mean? I could not find any partition dependent code in the Overlapping* classes.)

> My stomach aches whenever I see people switching to float for performance. 

I don't like this either, but its hard to argue against a 50% speedup, IMO. let's meet in the emergency room for stomach sicknesses ;)

> Like you noticed, there can be side effects in stability.

the linear solver actually never failed to converge, just the non-linear one. maybe something else is wrong...

> In the same light, please comment on what iterative solver you have been using and how the number of steps of the iterative solver and setup time has changed.

using the default settings (i.e., `BiCGSTAB<ILU0>`) and using 10 seconds as the length of the first time step, the number of linear iterations per non-linear iterations stayed basically the same for this step (47.7 for double and 42.8 for float), but using `float` required on additional Newton iteration (7 instead of 6). the full run of the simulation used about 5% more Newton iterations (440 instead of 418). I haven't measured the setup time of the linear solver, but I suppose the speedup for the ILU is in line.

andlaus:
> the linear solver actually never failed to converge, just the non-linear one. maybe something else is wrong...

it turned out to be "something else": the fixpoint tolerance passed to the linear solver was 100*epsilon which is way too large for single precision scalars. if this tolerance is set to 0.0, most tests pass. I added fixes to the PR...

andlaus:
Travis is happy! If there are no objections, I'll merge this in the afternoon.

PR #109, opened 2016-12-08
Linear solver tuning

andlaus:
this changes the default value for the maximum number of linear iterations of the non-AMG solvers from 250 to 1000 and relaxes the absolute tolerance for the linear solvers.

This improves the performance of some tests while the chances that it causes regressions elsewhere are pretty slim IMO.


andlaus:
travis is happy, merging.

PR #110, opened 2016-12-08
start(): reset the locale on startup

andlaus:
sometimes incorrect user defined locale settings caused trouble, so
let's reset them to the defaults before we do anything real. (also,
mixing locales in the output looks quite weird IMO:
e.g. German-formatted numbers and English text do not go too well
together.)


andlaus:
this is trivial. self-merging.

PR #111, opened 2016-12-08
ebos: update the PffGridVector for transmissibility after geologic events

andlaus:
this regression was found by @totto82. it was introduced when the explicit prefetching feature was added. Since it only affected decks with geology changes in the SCHEDULE section (a.k.a. Model 2), analysing this issue must have been a *huge* nightmare, so I owe @totto82 at least one beer.


totto82:
Thanks for the patch.

PR #117, opened 2016-12-27
do not access possibly non-existing elements and avoid zero reference pressure in the PVS model

andlaus:
this caused the water-air problem to segfault in parallel.

### DIFF
--- a/ewoms/disc/common/fvbasediscretization.hh
+++ b/ewoms/disc/common/fvbasediscretization.hh
@@ -1578,10 +1578,15 @@ public:
             ElementContext elemCtx(simulator_);
             ElementIterator elemIt = threadedElemIt.beginParallel();
             for (; !threadedElemIt.isFinished(elemIt); elemIt = threadedElemIt.increment()) {
+                const auto& elem = *elemIt;
+                if (elem.partitionType() != Dune::InteriorEntity)
+                    // ignore non-interior entities
+                    continue;
+
                 if (needFullContextUpdate)
-                    elemCtx.updateAll(*elemIt);
+                    elemCtx.updateAll(elem);
                 else {
-                    elemCtx.updatePrimaryStencil(*elemIt);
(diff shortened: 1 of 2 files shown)


ewoms/disc/common/fvbasediscretization.hh
```diff
@@ -1578,10 +1578,15 @@ public:
             ElementContext elemCtx(simulator_);
             ElementIterator elemIt = threadedElemIt.beginParallel();
             for (; !threadedElemIt.isFinished(elemIt); elemIt = threadedElemIt.increment()) {
+                const auto& elem = *elemIt;
+                if (elem.partitionType() != Dune::InteriorEntity)
+                    // ignore non-interior entities
+                    continue;
```
blattms: Does this mean what the comment suggests? Nothing gets assembled for non-interior entities? Are they not represented in the linear systems?

andlaus: it does what the comment suggests, but this method is about writing visualization output...

andlaus: > Are they not represented in the linear systems?

actually, non-interior entities are in fact not linearized on the local process. they are distributed -- and if they represent border entities -- summed up as a precompute step of the linear solve, though, so they are represented. (without being aware of them, the linear solvers would hardly converge?!)

blattms:
> actually, non-interior entities are in fact not linearized on the
> local process. they are distributed --

Not sure what you mean by "not linearized but are distributed".

> and if they represent border entities -- summed up as a precompute
> step of the linear solve, though, so they are represented.

What if you do not have dofs on border entities because they are
e.g. associated with elements only.

andlaus:
> Not sure what you mean by "not linearized but are distributed".

every process linearizes its interior and border DOFs and sends the results to the peer processes that need them.

> What if you do not have dofs on border entities because they are e.g. associated with elements only.

then the set of entities which are summed up is empty.

PR #118, opened 2016-12-29
Improve timer

andlaus:
because it seems to be the spirit of the season, here are some timer related improvements: With this, `Ewoms::Timer` uses basically the same semantics as `Dune::Timer`. In contrast to `Dune::Timer`, `Ewoms::Timer` objects can be summed up, measure CPU time and support determining the total CPU time of the whole MPI parallel simulation.

another neat (IMO) addition of this PR is `Ewoms::TimerGuard` which ensures that a timer is stopped once the `TimerGuard` object goes out of scope. This is particularly useful if a function is terminated by an exception thrown by a called function.

### DIFF
--- a/ewoms/common/simulator.hh
+++ b/ewoms/common/simulator.hh
@@ -33,6 +33,7 @@
 
 #include <ewoms/common/propertysystem.hh>
 #include <ewoms/common/timer.hh>
+#include <ewoms/common/timerguard.hh>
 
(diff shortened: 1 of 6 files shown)


andlaus:
jenkins build this please

andlaus:
since both jenkins and travis are happy and I verified locally that flow_ebos did not regress on norne in any way, I'll merge.

PR #119, opened 2016-12-29
do not pass a reference to the non-linear solver to Model::update()

andlaus:
the non-linear solver is (and always has been) an internal implementation detail of the model, so passing an object other than the internal Newton Raphson solver would not have worked anyway.


andlaus:
IMO this falls into the "trivial department". I made sure that the test suite still compiles and passes on my machine, so I'll merge.

PR #120, opened 2016-12-29
Implement a bicgstab linear solver

andlaus:
I.e., the `solvers.hh` file is removed. The main motivation for this is that it avoids having to distribute a file with a potentially incompatible license (i.e., GPLv2 + template exception vs GPLv2+), but the home-brewn bicgstab solver also seems to perform a tiny bit better (i.e., <= 3%) .

Also, the unmodified dune-istl linear solvers are made easily accessible. Of these, the restarted GMRES solver with the ILU-2 preconditioner is used for the water-air test problem, but I do not really recommend using these solvers because BiCGSTAB tends to be 20% to 30% slower than our home-brewn implementation (this is because the dune-istl solvers cannot use custom convergence criteria), but dune-istl offers more choices than just BiCGStab and this functionallity could be helpful when debugging issues related to solving the linear systems of equations.

Note that regardless of how pedantic the interpretation of DUNE's license is, there are no licensing issues anymore because we do no longer distribute any files derived from DUNE (for which the template exception might or might not apply).


andlaus:
jenkins build this please

andlaus:
lacking strong vetos, I'll merge this tomorrow...

joakim-hove:
> lacking strong vetos, I'll merge this tomorrow...

Well - I guess I am not the one to to put forward vetos here - but if I could I certainly would. Here is a brand new class for doing linear algebra; *without a single test* - I just don't get it.

andlaus:
> without a single test

that's not true: all the unit tests (you would probably call them "integration tests") except the one for the water-air problem use it. the results are exactly the same as before with the performance being basically the same with a tendency to be slightly better.

joakim-hove:
> all the unit tests (you would probably call them "integration tests") except the one for the water-air problem use it

Those are old tests - and it is certainly a good thing that they still pass. But this is new code, and then it would in my opinion be natural to also add a new test explicitly testing the new code.

andlaus:
IMO, the only thing which such a test does in practice is to make API changes harder. For classes that are used by a lot of downstream code, this is a good thing, but classes which should be considered internal implementation details with only a single implementation (such as `Ewoms::BiCGStabSolver`), it is not IMO.

Anyway, if somebody jumps the gun and provides such a test, I'm don't mind merging it.

pgdr:
I tend to agree with Joakim in this regard.  I strongly prefer _test-driven development_ and in that aspect, this work is only half done and shouldn't be merged.

andlaus:
IMO TDD is a good thing in principle, but if it gets applied in a too dogmatic fashion, it can become counterproductive. (the old rule "the dose makes the poison" applies.)

> this work is only half done and shouldn't be merged.

I think it is an improvement compared to what's in the repo now, so it should be merged. I freely admit that it's not perfect, but "perfect is the enemy of good"...

andlaus:
with regard to testing: IMO it is equally important that the code is easy to understand and that the API well documented.

atgeirr:
This PR resolves an unclear licensing issue which in my opinion is quite important. While I agree that we should strive towards test-driven development, we do not mandate it. The change does not affect the Flow family, and for the parts it *does* affect it is tested quite extensively. That it does not come with new tests should not be held against it, since it replaces existing code and satisfies all its tests. So from the testing point of view I think this is fine.

It would be good to have a second pair of eyes read through the algorithm though. I'll therefore take a brief look now.

andlaus:
> It would be good to have a second pair of eyes read through the algorithm though. I'll therefore take a brief look now.

that would be cool, thanks.

pgdr:
> I think it is an improvement compared to what's in the repo now, so it should be merged

The road to hell is paved with merged improvements.  And with that, I've raised my concerns and will stand down.

ewoms/linear/bicgstabsolver.hh
```diff
@@ -0,0 +1,357 @@
...
+                // p_i = r_(i-1) + beta*(p_(i-1) - omega_(i-1)*v_(i-1))
+                auto tmp = v[i];
+                tmp *= omega;
+                tmp -= p[i];
+                tmp *= -beta;
+                p[i] = r[i];
+                p[i] += tmp;
+
+                // y = p
+                y[i] = p[i];
```
atgeirr: Is this necessary, given that line 228 will overwrite it? I thought the apply() would not use the input `y` but perhaps I am wrong.

joakim-hove:
> While I agree that we should strive towards test-driven development, we do not mandate it

I agree it is not mandated; but although there seems to be an acceptance for enforcing style e.g. when it comes to braces and method names - testing seems to be deemed less important, and is very often met with the: "Would have been to have some tests - but whatever I'll merge".

>It would be good to have a second pair of eyes read through the algorithm though. I'll therefore take a brief look now.

Well - that kind of proves my point; the **absolutely best way** to get a pair of second eyes - which are *persisted in the code* - is by adding test; so here I really feel that you @atgeirr are arguing in circles - either this code is obviously correct - and will remain so far all eternity - in which case it should just be merged right away, otherwise there are things which might be incorrect/improved and so on - in which case adding tests make sense. 

OK - I have made my point; which is more about testing - and specifically the unwillingness in the OPM community to improve in that regard, more than this actual patch. I'll back off now.

ewoms/linear/bicgstabsolver.hh
```diff
@@ -0,0 +1,357 @@
...
+            }
+
+            // y = K^-1 * p_i
+            preconditioner_->apply(y, p);
+
+            // v_i = A*y
+            A_->apply(y, v);
```
atgeirr: I think this is correct, but I had to check the docs to verify that the apply() methods of the operators and preconditioners modify different arguments. Looks correct to me here in any case.

ewoms/linear/bicgstabsolver.hh
```diff
@@ -0,0 +1,357 @@
...
+                          "Breakdown of the BiCGStab solver (division by zero)");
+            alpha = rho_i/denom;
+
+            // h = x_(i-1) + alpha*y
+            // s = r_(i-1) - alpha*v_i
+            for (unsigned i = 0; i < n; ++i) {
+                auto tmp = y[i];
+                tmp *= alpha;
+                tmp += x[i];
+                h[i] = tmp;
```
atgeirr: Why not simply `x[i] - alpha*y[i]`? Is there any reason to expect better performance with an explicit temporary than an implicit one?

ewoms/linear/bicgstabsolver.hh
```diff
@@ -0,0 +1,357 @@
...
+
+            // z = K^-1*s
+            z = s;
+            preconditioner_->apply(z, s);
+
+            // t = Az
+            t = z;
+            A_->apply(z, t);
+
+            // omega_i = (t*s)/(t*t)
```
atgeirr: ... so only right preconditioning. Correct?

atgeirr:
@joakim-hove I think you are wrong about this code. It has relevant tests that were passed by the existing code, and it passes them too. It should therefore be considered a fairly good example of the freedom to refactor and change that having such tests grants you!

joakim-hove:
> I think you are wrong about this code. It has relevant tests that were passed by the existing code, and it passes them too. 

I see that point - and agree. But this is still a *brand new class* - there is no point in the lifetime of a class where it easier to add tests than during the conception; for this particular class that is about to become a missed opportunity.

andlaus:
> there seems to be an acceptance for enforcing style 

As far as I can see the policy w.r.t. is in the discretion of whoever merges the respective pull requests (i.e., the module maintainers). the same applies to the "everything must come with microtests" policy?!

> Well - that kind of proves my point; the absolutely best way to get a pair of second eyes

I honestly cannot see how, because as far as I can see the test which you want would *not* make sure that the algorithm is correct but only enforces the  API. (mainly because the test would use the results that are produced by the implementation, i.e. it would be "upside down".)

> and will remain so far all eternity

that boils down if the stabilized BiCG algorithm is "correct". I cannot judge that and neither can I test it. (the most I can test is that a linear system of equations gets solved, and this is extensively tested already.)

> OK - I have made my point; which is more about testing - and specifically the unwillingness in the OPM community to improve in that regard, more than this actual patch. I'll back off now.

IMO, thorough testing is a good thing in general. Personally, I think that easy readability is more important, though...

ewoms/linear/bicgstabsolver.hh
andlaus: maybe it's not required, i.e., it is possible that the dune-istl preconditioner interface does mandate that the destination object can be random. I'll check the Dune docu...

ewoms/linear/bicgstabsolver.hh
andlaus: yeah, that part of Dune is pretty weird, IMO: e.g. the arguments for the operations are reversed compared the standard C functions like `memcpy()` . (and don't get me started on the inconsistent naming. ;)

ewoms/linear/bicgstabsolver.hh
andlaus: the reason is again Dune: `Dune::FieldVector` does not provide `operator+()`, probably to avoid unnecessary temporary objects. In this instance `x[i] - alpha*y[i]` would probably the same w.r.t. performance...

ewoms/linear/bicgstabsolver.hh
andlaus: yep. that seems to be what the Dune preconditioner interface mandates.

ewoms/linear/bicgstabsolver.hh
andlaus: ok, it looks like this is not required because the Dune docu mandates y to be `K^-1*p`. (whether all preconditioners really adhere to this, I cannot say, though.) changed, thanks.

pgdr:
I'm sorry, but I cannot let this one go:

> IMO, the only thing which such a test does in practice is to make API changes harder

First and foremost:  Making API changes _is supposed to be hard!_

Second, "such a test in practice" makes sure you write _testable code_ which is much easier to read, and makes sure your code is run, at least once.

Third, it provides _"minimal working examples"_ for your code, i.e., it shows which assumptions you have and what you need in place for a function call to work as expected.

Four, a test allows you to verify that the code works also in _corner cases._  We cannot assume that the Norne or SPE cases test different corner cases of our APIs, but we _can assume_ that users will experience corner cases.  A regression test typically does not test corner cases; a unit test might.

Okay. _Now_ I'll stand down.

andlaus:
ok, seems like we disagree on some details ;) (Though I'd argue that the overlap of our development philosophy is > 80%, so let's not blow the differences up too much.)

anyway, I'll merge this now. follow-up improvements are always welcome!

PR #122, opened 2017-01-02
Pass enableSwatinit to the OPM initialization 

totto82:
Only apply swatinit in the initialization if swatinit is enabled.

### DIFF
--- a/ebos/eclequilinitializer.hh
+++ b/ebos/eclequilinitializer.hh
@@ -137,13 +137,17 @@ public:
                                             /*numFaces=*/0, // we don't care here
                                             opmPhaseUsage.num_phases);
 
+        // tell the initializers whether SWATINIT should be applied or not.
+        bool enableSwatInit = GET_PROP_VALUE(TypeTag, EnableSwatinit);
+
         // do the actual computation.
         Opm::initStateEquil(equilGrid,
                             opmBlackoilProps,
                             gridManager.deck(),
                             gridManager.eclState(),
                             simulator.problem().gravity()[dimWorld - 1],
-                            opmBlackoilState);
+                            opmBlackoilState,
+                            enableSwatInit);
 
         std::vector<int> localToEquilIndex( numElems, -1 );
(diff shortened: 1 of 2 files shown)


andlaus:
are you sure that this is correct? SWATINIT does not get applied to the material law manager here but only later in the problem (via `EclProblem::applySwatinit()`). That said I'm not very familiar with `Opm::initStateEquil()`...

andlaus:
as far as I can tell, the PR is good to go in from the code point of view. please merge it yourself once you convinced yourself that it is semantically correct. (thank for the work, BTW!)

totto82:
> are you sure that this is correct?

If enableSwatinit is set to false it will be ignored also in the computation of the threshold pressures. 
When calling flow_ebos the order of calculations is as follows
1) the initial state is calculated in ebos without SWATINIT (since enableSwatinit==false) 
2) default th-press is calculated
3) the initial state is calculated in flowEbosMain.hpp with SWATINIT. The call to Opm::initStateEquil() will also apply swatinit to the material law manager.

totto82:
I think the logic is "correct", at least in the sense that it reproduces Ecl.  I still think it is more sensible to include the SWATINIT effects in the default th-pressure calculations. (i.e. revert this PR). 

I also made some small changes to the thpressure calculations (excluding faces with negligible flow). I am still waiting for the full simulation to complete.

andlaus:
> I think the logic is "correct", at least in the sense that it reproduces Ecl.

ok, I see. go ahead and merge it. please make travis happy before, though.

atgeirr:
> I still think it is more sensible to include the SWATINIT effects in the default th-pressure calculations.

So should we wait for that solution to be implemented, or aim towards merging this?

atgeirr:
jenkins build this opm-core=1129 please

totto82:
> That solution

 is the current solution. So no need for waiting. But we should keep this issue in mind if (or when) we rewrite the initialization code.

PR #123, opened 2017-01-04
Refactor the linear solver backends and the convergence criteria

andlaus:
this gets rid of quite a bit of copy-and-pasted code, replaces the "equation weight vector" for the  combinedCriterion class by rescaling the linear system of equations to be solved, improves the API and class names of the convergence criteria with an eye on performance and puts the preconditioner- and solver wrappers into separate files.

note that with this the prestine linear solvers from dune-istl now also respect the weight of equations (due to the rescaling the linear systems before solving them). Interestingly, the prestine dune-istl solvers do not seem to profit from this performance-wise while the home-brewn `BiCGStabSolver` seems to get 7 to 8% faster, probably because the `CombinedCriterion` performs better. This brings the performance advantage of `Ewoms::BiCGStabSolver` close to 50% for my testcase when compared to `Dune::BiCGSTABSolver`.

### DIFF
--- a/ewoms/common/genericguard.hh
+++ b/ewoms/common/genericguard.hh
@@ -1,0 +1,60 @@
+// -*- mode: C++; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*-
+// vi: set et ts=4 sw=4 sts=4:
+/*
+  This file is part of the Open Porous Media project (OPM).
+
+  OPM is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 2 of the License, or
+  (at your option) any later version.
+
+  OPM is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with OPM.  If not, see <http://www.gnu.org/licenses/>.
+
+  Consult the COPYING file in the top-level source directory of this
+  module for the precise wording of the license and the list of
+  copyright holders.
+*/
+/*!
+ * \file
+ *
+ * \copydoc Ewoms::GenericGuard
+ */
+#ifndef EWOMS_GENERIC_GUARD_HH
+#define EWOMS_GENERIC_GUARD_HH
+
+namespace Ewoms {
+/*!
+ * \ingroup Common
+ *
+ * \brief A simple class which makes sure that a cleanup function is called once the
+ *        object is destroyed.
+ *
+ * This class is particularly useful in conjunction with lambdas for code that might
+ * throw exceptions.
+ */
+template <class Callback>
+class GenericGuard
+{
+public:
+    GenericGuard(Callback& callback)
+        : callback_(callback)
+    { }
+
+    ~GenericGuard()
+    {
+        callback_();
+    }
+
+private:
+    Callback& callback_;
+};
+
+} // namespace Ewoms
+
+#endif
--- a/ewoms/common/genericguard.hh
+++ b/ewoms/common/genericguard.hh
@@ -44,15 +44,33 @@ class GenericGuard
 public:
     GenericGuard(Callback& callback)
         : callback_(callback)
+        , isEnabled_(true)
     { }
 
     ~GenericGuard()
     {
-        callback_();
+        if (isEnabled_)
+            callback_();
     }
+
+    /*!
+     * \brief Specify whether the guard object is "on duty" or not.
+     *
+     * If the guard object is destroyed while it is "off-duty", the cleanup callback is
+     * not called. At construction, guards are on duty.
+     */
+    void setEnabled(bool value)
+    { isEnabled_ = value; }
+
+    /*!
+     * \brief Returns whether the guard object is "on duty" or not.
+     */
+    bool enabled() const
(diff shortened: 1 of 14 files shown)


andlaus:
jenkins build this please

andlaus:
jenkins build this please

andlaus:
With the CI systems being happy, I'll merge this tomorrow if no vetos are voiced...

blattms:
I assume that the rest of OPM still uses the upstream solvers whithout
scaling, right?

> note that with this the prestine linear solvers from dune-istl now
> also respect the weight of equations (due to the rescaling the
> linear systems before solving them). Interestingly, the prestine
> dune-istl solvers do not seem to profit from this performance-wise
> while the home-brewn `BiCGStabSolver` seems to get 7 to 8% faster,
> probably because the `CombinedCriterion` performs better. This
> brings the performance advantage of `Ewoms::BiCGStabSolver` close to
> 50% for my testcase when compared to `Dune::BiCGSTABSolver`.

Maybe I did not understand what the CombindCriterion does. I assumed
that it checks whether $\|w^T r_i\|/\|w^T r_0\| < \epsilon$ holds
where w^T is your weight/scale vector.

For a Krylov method you can write $r_i = P_i(A) r_0$ where P_i is a
polynomial of the matrix. Are sure that $w^T r_i$ can be the same as
$P_i(w^TA) w^T (b-Ax)$ besides some corner cases?

If not then you might be comparing apples with oranges. Does it make
sense to compare performance in this case?

andlaus:
> I assume that the rest of OPM still uses the upstream solvers whithout scaling, right?

nope, the equations are always scaled before the linear solver is started. the philosophy behind it is that "all kilograms are born equal".

That said, I just discovered some regression when using MPI and very fine grids (the lens problem with three global refinements does not work in parallel with this PR but it works fine with master). maybe it will take a bit longer to get this in...

andlaus:
> Maybe I did not understand what the CombindCriterion does. I assumed that it checks whether $\|w^T r_i\|/\|w^T r_0\| < \epsilon$ holds where w^T is your weight/scale vector.

correct. (BTW: w does not need to be transposed because it is diagonal.) observe that the w term can be folded into the original system of equations by left-multiplying both sides with w. this is the main idea of this PR.

> If not then you might be comparing apples with oranges.

maybe I do to some extend, but the actual linear system of equations which is solved by both solvers are exactly identical because they use the same code up until the solver is started. Even then, the convergence behaviour is probably identical for both solvers (haven't checked, you can by comparing the terminal output when passing `--linear-solver-verbosity=2`). as far as I can see, the only difference is the convergence criterion and some minor micro-optimizations of `Ewoms::BiCGStabSolver`.

blattms:
> With the CI systems being happy, I'll merge this tomorrow if no
> vetos are voiced...

Famous last word words...

> > I assume that the rest of OPM still uses the upstream solvers
> > whithout scaling, right?
> nope, ...
> I just discovered some regression ( on the lens problem)

Seems like we are changing code that affects flow without actually
testing it.

I am quite surprised how many implicite changes we got and are still
getting under the label "dense AD". It has become much harder to
identify changes that will affect flow than before the merge.

atgeirr:
> Seems like we are changing code that affects flow without actually
> testing it.

It was my understanding that none of these changes would or could affect flow-ebos. Is that wrong, @andlaus?

andlaus:
> It was my understanding that none of these changes would or could affect flow-ebos. Is that wrong, @andlaus?

no, that's 100% correct.

the regression I've mentioned earlier also seems to be somewhat esotheric: it only occurs for large grids (the lens test problem with three global grid refinements, i.e. 384x256), and only in parallel. (it would be quite unwise to include such things in automatic tests.) I'll investigate.

andlaus:
> I am quite surprised how many implicite changes we got and are still getting under the label "dense AD". It has become much harder to identify changes that will affect flow than before the merge.

the dense AD stuff is only relevant for the localized linearization, i.e. it has nothing to do with this code whatsoever.

andlaus:
ok, I've found and fixed the regression: sometimes it makes a difference whether the maximum or the minimum is used ;). if the CI systems are happy and nothing serious turns around the corner, I'll merge this tomorrow.

andlaus:
jenkins build this please

andlaus:
> We have untested in flow_ebos; that is the way it goes.

in what sense do you think it is untested? it is used by (almost) all eWoms tests, i.e. about 45 of them and it is also tested in parallel if MPI is installed. We can never test all cornercases!? (proving that the code is correct is called "model checking" in computer science and is basically impossible for non-toy code.)

joakim-hove:
> in what sense do you think it is untested? it is used by (almost) 

OK - sorry; I do not know for this particular code. I only commented based on the previous comments @blattms  from and mainly @atgeirr.

andlaus:
I rebased the branch on top of the current master to solve some merge conflicts.

andlaus:
jenkins build this please

blattms:
> the dense AD stuff is only relevant for the localized linearization, i.e. it has nothing to do with this code whatsoever.

Good (at least for now). I am sorry that I misinterpreted your
previous answer.

How ar people supposed to know what changes are relevant and what are
not? Could we maybe at least start tagging (e.g. "[flow]") to give
fellow developers a hint whether or not they are supposed to look at
an MR?

andlaus:
> Could we maybe at least start tagging (e.g. "[flow]") to give fellow developers a hint whether or not they are supposed to look at an MR?

I think that's a good idea. we could discuss the details at the dev-meeting?

PR #124, opened 2017-01-05
properly deal with exceptions thrown while switching the primary variables

andlaus:
this affects MPI-parallel runs of models that use primary variable switching, i.e. the PVS and the blackoil models. this was not noticed earlier because exceptions raised in this part of the code are quite rare. (but sometimes sh** hits the fan!)

### DIFF
--- a/ewoms/disc/common/fvbaselinearizer.hh
+++ b/ewoms/disc/common/fvbaselinearizer.hh
@@ -180,10 +180,26 @@ public:
             succeeded = 1;
             succeeded = gridView_().comm().min(succeeded);
         }
-        catch (Opm::NumericalProblem& e)
+        catch (const std::exception& e)
         {
             std::cout << "rank " << simulator_().gridView().comm().rank()
                       << " caught an exception while linearizing:" << e.what()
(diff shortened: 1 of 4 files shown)


andlaus:
jenkins build this please

ewoms/disc/common/fvbaselinearizer.hh
```diff
@@ -180,14 +180,30 @@ public:
...
+            succeeded = 0;
+            succeeded = gridView_().comm().min(succeeded);
+        }
+        catch (...)
+        {
+            std::cout << "rank " << simulator_().gridView().comm().rank()
+                      << " caught an exception while linearizing"
+                      << "\n"  << std::flush;
+            succeeded = 0;
+            succeeded = gridView_().comm().min(succeeded);
```
atgeirr: Not that I am an MPI expert, but should not the min() call be made on all processes, that is after the catch blocks?

andlaus:
jenkins is happy ( https://ci.opm-project.org/job/ewoms-PR-builder/29/console ) and my workstation is happy as well. since this clearly fixes a bug, I'll merge.

PR #125, opened 2017-01-05
newton: also call the endIteration_() callback if we do not need to solve the linearized system of equations

andlaus:
the purpose of this is to print the error for the last iteration of a timestep to the terminal.


andlaus:
since this is pretty trivial and my local testing succeeds, I'll merge without waiting for the CI systems.

PR #129, opened 2017-01-11
simulator: add a way to force the time step sizes externally

andlaus:
the forced time steps are read from a file which is supposed to feature one time step size in seconds per line. this kludge is intended to ease debugging/comparisons, so it does not even attempt to cover any corner cases (like e.g. non-convergent time steps). given that this feature is turned of by default, it should not hurt anyone to have it?!

### DIFF
--- a/ewoms/common/simulator.hh
+++ b/ewoms/common/simulator.hh
@@ -39,7 +39,10 @@
 #include <dune/common/parallel/mpihelper.hh>
 
 #include <iostream>
+#include <fstream>
 #include <iomanip>
+#include <vector>
+#include <string>
 #include <memory>
 
 namespace Ewoms {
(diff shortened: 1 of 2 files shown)


andlaus:
jenkins build this please

ewoms/common/simulator.hh
```diff
@@ -660,9 +673,17 @@ public:
                 problem_->endEpisode();
                 episodeBegins = true;
             }
-            else
-                // ask the problem to provide the next time step size
-                setTimeStepSize(problem_->nextTimeStepSize());
+            else {
+                Scalar dt;
+                if (predetTimeStepStream_.is_open() && !predetTimeStepStream_.eof())
```
atgeirr: I'd prefer if you would read the entire file at init into an `std::vector<double>` (or equivalent) instead, I think that is much less messy than keeping open a file. Also, iterating over a stream with is_open() and eof() etc. is much more prone to programmer mistakes than iterating over a vector.

Also: you really don't like braces...?

andlaus: > I'd prefer if you would read the entire file at init into an `std::vector<double>`

good idea.

> Also: you really don't like braces...?

Let's say I'm a braces agnostic ;) (at least if they are not doing too much to clarify how things are supposed to work.)

andlaus: ok, I've implemented "read all steps at init" and also added some braces...

PR #131, opened 2017-01-14
make the handbook compile on modern LaTeX distributions

andlaus:
TeXLive 2016 complains about \it and \bf.

Also, this patch adds a simple bash script to create the handbook from its LaTeX sources. Note that this script does *not* attempt to detect if all prerequisites (in terms of binaries and LaTeX packages) are properly available.

### DIFF
--- a/doc/handbook/models.tex
+++ b/doc/handbook/models.tex
@@ -16,7 +16,7 @@ systems is straightforward and can be found, e.\ g., in
 \subsection{Basic Definitions and Assumptions for the Compositional
   Model Concept}
 \textbf{Components:}
-The term {\it component} stands for constituents of the phases which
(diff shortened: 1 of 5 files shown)


doc/handbook/models.tex
```diff
@@ -16,7 +16,7 @@ \section{Physical and mathematical description}
 \subsection{Basic Definitions and Assumptions for the Compositional
   Model Concept}
 \textbf{Components:}
-The term {\it component} stands for constituents of the phases which
+The term \textit{component} stands for constituents of the phases which
```
pgdr: should probably be `\emph{·}`

andlaus: right. my LaTeX foo is not too advanced. I'll make a PR.

PR #133, opened 2017-01-16
handbook: use more sensible options for the hyperref package

andlaus:
this gets rid of the ugly red and green boxes around references. Note that for some reason, hyperlinks do not seem to work anyway and that using pdflatex directly does not work either :/

### DIFF
--- a/doc/handbook/build-handbook.sh
+++ b/doc/handbook/build-handbook.sh
@@ -1,11 +1,11 @@
 #! /bin/sh
 
-# this script build the eWoms handbook from its LaTeX sources. The
+# this script builds the eWoms handbook from its LaTeX sources. The
 # result file is called "ewoms-handbook.pdf"
 
-latex ewoms-handbook
-bibtex ewoms-handbook
-latex ewoms-handbook
-latex ewoms-handbook
-dvipdf ewoms-handbook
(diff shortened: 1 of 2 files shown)


doc/handbook/build-handbook.sh
```diff
@@ -1,11 +1,11 @@
...
-# this script build the eWoms handbook from its LaTeX sources. The
+# this script builds the eWoms handbook from its LaTeX sources. The
 # result file is called "ewoms-handbook.pdf"
 
-latex ewoms-handbook
-bibtex ewoms-handbook
-latex ewoms-handbook
-latex ewoms-handbook
-dvipdf ewoms-handbook
+latex ewoms-handbook || exit $?
```
pgdr: you could consider using `pdflatex` to generate the pdf immediately.

andlaus: I tried that, unfortunately it did not work. I guess this is because some pictures of the (Dumux inherited) tutorial use pstricks to place their labels. Possibly, I should do the work to get rid of them (IIRC the same is somehow possible with pdflatex using some inkscape plugin), but I'm a lazy bastard and the handbook is not very high on my list of priorities ;)

PR #135, opened 2017-01-17
Use copy instead of processing eclipse data for duplicating grid.

blattms:
Due to the nature of CpGrid (manages one shared pointer to the equil/global grid and after load balancing another one to the distributed grid) simply copying the grid before load balancing should be faster as it only adjusts one shared pointer.


andlaus:
thanks.

PR #137, opened 2017-01-18
Sets up a correct map from element to interior element index.

### LINKED ISSUE #105: PR #101 changed semantics and introduced memory bugs
As my comments in that PR were not honored. Here is another try:
[lines 245 ff](/andlaus/ewoms/blob/99f281484667c858e1a8137773c12347eff6723d/ebos/collecttoiorank.hh#L245-L248) seem to be a buggy change of semantics. Note that only interior entities were pushed previously (resulting in a map of entity index (by traversal) to interior entity index (by traversal). The size of localIndexMap_ is never set.

```
(gdb)
#0  Ewoms::CollectDataToIORank<Ewoms::EclCpGridManager<Ewoms::Properties::TTag::EclFlowProblem> >::CollectDataToIORank (this=0x1bad038, gridManager=...)
    at /home/mblatt/src/dune/opm/ewoms/ebos/collecttoiorank.hh:247
247                             localIndexMap_[elemIdx] = elemIdx;
(gdb) p elemIdx
$5 = 1
(gdb) p localIndexMap_
$6 = std::vector of length 0, capacity 165
(gdb) 
```

blattms:
I assume that is what the code should do. Not 100% sure though.
It also closes #105 by actually resizes the arrays we store values using random access in.

### DIFF
--- a/ebos/collecttoiorank.hh
+++ b/ebos/collecttoiorank.hh
@@ -120,6 +120,7 @@ namespace Ewoms
                     indexMap.resize( localSize );
                     for( size_t i=0; i<localSize; ++i )
                     {
+                        assert( localIndexMap_[ i ] != -1 );
                         const int id = distributedGlobalIndex_[ localIndexMap_[ i ] ];
                         indexMap[ i ] = globalPosition_[ id ] ;
 #ifndef NDEBUG
@@ -227,6 +228,7 @@ namespace Ewoms
 
                 // store the local Cartesian index
                 IndexMapType distributedCartesianIndex;
+                int interiorElementIdx = 0;
                 distributedCartesianIndex.resize(gridSize, -1);
 
                 auto localView = gridManager.grid().leafGridView();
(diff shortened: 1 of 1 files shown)


blattms:
It turns out that my changes raise an assertion in a sequential run. Why is this code even executed in this case?

andlaus:
thanks, will merge.

> Why is this code even executed in this case?

I can't tell you for sure, but I guess it is easier to just create this object unconditionally.

blattms:
I said

> It turns out that my changes raise an assertion in a sequential run

Your merge might have been a bit premature...

andlaus:
yeah, but I understood that the PR has been updated to fix it...

andlaus:
anyway, stand-alone ebos and flow_ebos seem to work with optimizations in sequential mode, so this is not a big deal IMO...

blattms:
I would not be so sure that everybody uses `-DNDEBUG` when compiling OPM. I only set this for benchmarking.

PR #138, opened 2017-01-18
Switch of collecting data in non-parallel run.

blattms:
Might save some time, space, and suprises.

Completes #137. @andlaus merges faster than I commit.

### DIFF
--- a/ebos/collecttoiorank.hh
+++ b/ebos/collecttoiorank.hh
@@ -177,8 +177,14 @@ namespace Ewoms
 
         CollectDataToIORank( const GridManager& gridManager )
             : toIORankComm_( ),
-              isIORank_( gridManager.grid().comm().rank() == ioRank )
+              isIORank_( gridManager.grid().comm().rank() == ioRank ),
+              isParallel_( gridManager.grid().comm().size() > 1 )
         {
+            if ( !isParallel_ )
+            {
+                // no need to collect anything.
+                return;
+            }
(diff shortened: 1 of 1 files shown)


blattms:
and gets rid of the assertion mentioned in #137

andlaus:
push green if you deem it to be ready.

blattms:
I will not self merge. Otherwise I cannot complain about that in the future anymore.

atgeirr:
Now *that* is a deadlock...

andlaus:
> I will not self merge. Otherwise I cannot complain about that in the future anymore.

ok, I'll merge, but I refuse to be responsible for any breakage.

atgeirr:
> ok, I'll merge, but I refuse to be responsible for any breakage.

I think a maintainer will always at least share the responsibility for handling a situation where things break, unless bullied by the PR author to merge :-) (I do not think that is the case here.)

More to the point: did you check this @andlaus, by reading or running the code, or both?

andlaus:
Well, I think that the author of a patch has a certain responsibility to test and self-review the stuff before opening a PR. (at least myself has been doing this for my PRs during the last few years as you've doubtlessly noticed. ;))

I looked at the code, and it nothing seemed to be obviously wrong, but I usually only compile stuff in optimized mode and I do not run flow_ebos in parallel because the latter does not work yet anyway. (the first point was also true for #137, BTW.)

Issue #139, opened 2017-01-18
CollectToIoRank broken in parallel


blattms:
While it works sequentially (either when compiled with `-DNDEBUG' or with PR #138, it still get an assertion in a parallel run:
```
flow_ebos: /home/mblatt/src/dune/opm/ewoms/ebos/collecttoiorank.hh:127: Ewoms::CollectDataToIORank<GridManager>::DistributeIndexMapping::DistributeIndexMapping(const std::vector<int>&, const std::vector<int>&, Ewoms::CollectDataToIORank<GridManager>::IndexMapType&, Ewoms::CollectDataToIORank<GridManager>::IndexMapStorageType&) [with GridManager = Ewoms::EclCpGridManager<Ewoms::Properties::TTag::EclFlowProblem>; Ewoms::CollectDataToIORank<GridManager>::IndexMapType = std::vector<int>; Ewoms::CollectDataToIORank<GridManager>::IndexMapStorageType = std::vector<std::vector<int> >]: Assertion `checkPosition_.find( id ) == checkPosition_.end()' failed.

Program received signal SIGABRT, Aborted.
0x00007ffff20e1067 in __GI_raise (sig=sig@entry=6)
    at ../nptl/sysdeps/unix/sysv/linux/raise.c:56
56      ../nptl/sysdeps/unix/sysv/linux/raise.c: Datei oder Verzeichnis nicht gefunden.
(gdb) up
....
(gdb) 
#4  0x0000000000d71727 in Ewoms::CollectDataToIORank<Ewoms::EclCpGridManager<Ewoms::Properties::TTag::EclFlowProblem> >::DistributeIndexMapping::DistributeIndexMapping (this=0x7fffffffd280, 
    globalIndex=std::vector of length 300, capacity 300 = {...}, 
    distributedGlobalIndex=std::vector of length 348, capacity 348 = {...}, 
    localIndexMap=std::vector of length 174, capacity 174 = {...}, 
    indexMaps=std::vector of length 2, capacity 2 = {...})
    at /home/mblatt/src/dune/opm/ewoms/ebos/collecttoiorank.hh:127
127                             assert( checkPosition_.find( id ) == checkPosition_.end() );
(gdb) p checkPosition_ 
$1 = std::set with 1 elements = {[0] = -1}
(gdb) p id
$2 = -1
(gdb) p distributedGlobalIndex
$3 = std::vector of length 348, capacity 348 = {-1, -1, -1, -1, -1, -1, -1, 
[...]

PR #140, opened 2017-01-18
Fixes pushing to resized array distributedCartesianIndex.

blattms:
Previously distributedCartesianIndex was resized and afterwards all
entries were added with push_back. Therefore the array was twice as
big as expected and contained wrong values in the front.

With this commit we insert the values using random access. Thus the
size is as expeceted and the index of the entries do not depend on the
order of the grid traversal.

This is yet untested.

### DIFF
--- a/ebos/collecttoiorank.hh
+++ b/ebos/collecttoiorank.hh
@@ -247,7 +247,7 @@ namespace Ewoms
 #else
                     int elemIdx = elemMapper.map( element );
 #endif
-                    distributedCartesianIndex.push_back( gridManager.cartesianIndex( elemIdx ) );
+                    distributedCartesianIndex[elemIdx] = gridManager.cartesianIndex( elemIdx );
 
                     // only store interior element for collection
                     if( element.partitionType() == Dune :: InteriorEntity )


andlaus:
jenkins build this please

blattms:
Please hold off merging until I have tested this in parallel in my debugger. That might take a while..

blattms:
It fixes the problem that it advertises. Unfortunately, it reveals the next one (when running parallel)

I declare this safe to merge.

blattms:
When comparing with ParallelDebugOutput and how the arrays are set up there, I think I am on the wrong track and my bugfixes are bugs themselves. Closing

blattms:
This is ineed needed as it turns out.

blattms:
I accidentally deleted my branch (bug in shell script) and this was closed automatically. Restored and reopened.

blattms:
Is there something wrong with my fix?

dr-robertk:
No, looks good.

andlaus:
feel free to press green.

blattms:
BTW this should close #139

blattms:
Self merging after testing it again.

andlaus:
okay, I'll merge this now because it seems to be rather simple and the were buttlers once happy with it. (I haven't tested it on the latest master, but it would be strange if it was not working)

andlaus:
seems like we had the same idea at the same time ;)

PR #141, opened 2017-01-18
Fix extra warnings

andlaus:
this PR fixes a few warnings that are produced by GCC if it is passed `-Wextra` and removes the "evaluation point" concept from the intensive quantities. the latter potentially makes things a tiny bit faster because eight bytes are shoved of the intensive quantities object.


andlaus:
jenkins build this please

PR #142, opened 2017-01-18
ebos: fix the interactions between SWATINIT and threshold pressures

andlaus:
hopefully this makes standalone `ebos` arrive at the same initial condition as `flow_ebos` if both, SWATINIT and threshold pressures are enabled. (we need to calculate the initial condition if threshold pressures as well as SWATINIT are enabled. `ebos` and `flow_ebos` diverged after OPM/opm-core#1129.)

the proposed patch is a kludge IMO, but in the light that in my opinion, SWATINIT and threshold pressures are both physically not justified and given the fact that SWATINIT must not be considered for
the threshold pressures should be considered to be a bug of the reference simulator, I think the patch is okay.

I ran "model 2" on `flow_ebos` with this and the results and performance are identical.

### DIFF
--- a/ebos/eclequilinitializer.hh
+++ b/ebos/eclequilinitializer.hh
@@ -31,6 +31,7 @@
 #include <ewoms/common/propertysystem.hh>
 
 #include <opm/material/fluidstates/CompositionalFluidState.hpp>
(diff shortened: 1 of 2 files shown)


andlaus:
jenkins build this please

PR #144, opened 2017-01-19
Start using face geometry computed the Ecl way

totto82:
Face centers are computed using the cell corners. With this
implementation the face center seen from a cell may be different from the
face center seen from its neighbor.

Face normals with area lengths are calculated using the face corners
directly not using a triangulation point in the center of the face.

This gives transmissibility almost equal to eclipse.

Depends on OPM/opm-grid#261


andlaus:
no objections. whoever merges OPM/opm-grid#261 has the button.

andlaus:
jenins build this opm-grid=261 please

andlaus:
ok, this seems to be all ready. thanks for the great work Mr Holmes, err @totto82.

PR #147, opened 2017-01-26
EclCpGridManager: export the transmissibilities of the sequential grid

andlaus:
This only makes these transmissibilities available for parallel computations. The reason is that in the  sequential case, they do not need to be computed during grid creation and they are are also accessible via the problem object.


andlaus:
jenkins build this with downstream please

andlaus:
with the CI systems happy and this patch being unlikely to break existing code, I'll merge.

PR #150, opened 2017-02-07
Avoid ambiguous call of std::abs.

dr-robertk:


andlaus:
thanks. in which context did this fall on your feet?

PR #143, opened 2017-01-18
localIndexMaps in CollectToIORank need to be lists of interior element indices.

blattms:
Correct version of #137 that gets reverted with this PR,

When comparing the code with `ParallelDebugOutput.hh` (the source for copy and paste) I realized that #137 must be wrong as these arrays get passed to `P2PCommunicator` and therefore should look the same in both cases.

I conclude that reserving memory instead of resizing the vector was correct, but we have to use `push_back` to populate the vector to get a list of interior element indices. That is the input P2PCommunicator seems to need.

@dr-robertk should probably have a look at this, too.

Patch is tested together with #140 that I will open again as it is needed, too.

PR #152, opened 2017-02-10
When EnableEclOutput is false the EclWriter should not be created.

dr-robertk:
This fixes a problem with collecttoiorank which needs to be investigated
separately. The corresponding PR is #143 which does not seem to work yet.


andlaus:
the patch looks good and because travis is happy, there's not much reason to hold it back. thanks.

PR #153, opened 2017-02-13
ebos: no longer fail if the EQLDIMS keyword is not explicitly mentioned in the deck

andlaus:
the deck object currently does not provide a way to auto-generate an all-defaulted keyword if it has not been mentioned in the deck, so only access the keyword object if it is present.

possibly, this information is also exposed by EclipseState, but I have not found anything suitable.

@atgeirr: please merge if this fixes/improves the issues you've recently had.

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -75,6 +75,7 @@
 #include <opm/common/Valgrind.hpp>
 #include <opm/parser/eclipse/Deck/Deck.hpp>
 #include <opm/parser/eclipse/EclipseState/EclipseState.hpp>
+#include <opm/parser/eclipse/EclipseState/Tables/Eqldims.hpp>
 #include <opm/parser/eclipse/EclipseState/Schedule/Schedule.hpp>
 #include <opm/common/ErrorMacros.hpp>
 #include <opm/common/Exceptions.hpp>
@@ -827,8 +828,7 @@ public:
         // SWATINIT.
         const auto& deck = this->simulator().gridManager().deck();
         const auto& eclState = this->simulator().gridManager().eclState();
-        int numEquilRegions =
-            deck.getKeyword("EQLDIMS").getRecord(0).getItem("NTEQUL").template get<int>(0);
+        int numEquilRegions = eclState.getTableManager().getEqldims().getNumEquilRegions();
         bool useThpres = deck.hasKeyword("THPRES") && numEquilRegions > 1;
         bool useSwatinit =
             GET_PROP_VALUE(TypeTag, EnableSwatinit) &&
@@ -1101,9 +1101,7 @@ private:
     {
         const auto& deck = this->simulator().gridManager().deck();
         const auto& eclState = this->simulator().gridManager().eclState();
-        int numEquilRegions =
-            deck.getKeyword("EQLDIMS").getRecord(0).getItem("NTEQUL").template get<int>(0);
-
+        int numEquilRegions = eclState.getTableManager().getEqldims().getNumEquilRegions();
         bool useThpres = deck.hasKeyword("THPRES") && numEquilRegions > 1;
         bool useSwatinit =
             GET_PROP_VALUE(TypeTag, EnableSwatinit) &&
--- a/ebos/eclthresholdpressure.hh
+++ b/ebos/eclthresholdpressure.hh
@@ -36,6 +36,7 @@
 #include <opm/parser/eclipse/Deck/Deck.hpp>
 #include <opm/parser/eclipse/EclipseState/EclipseState.hpp>
 #include <opm/parser/eclipse/EclipseState/Grid/GridProperty.hpp>
+#include <opm/parser/eclipse/EclipseState/Tables/Eqldims.hpp>
 #include <opm/parser/eclipse/EclipseState/SimulationConfig/SimulationConfig.hpp>
 #include <opm/parser/eclipse/EclipseState/SimulationConfig/ThresholdPressure.hpp>
 #include <opm/common/ErrorMacros.hpp>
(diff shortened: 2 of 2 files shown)


joakim-hove:
This project would benefit *immensely* if people could check what was available already, and either just use that - or improve it first, instead of just rolling a one off hack. The `EclipseState` already has functionality in the `Eqldims` class - with proper default handling and all; admittedly the exposure to downstream classes could be improved - but that would be an investment.

andlaus:
> The EclipseState already has functionality in the Eqldims class - with proper default handling and all; admittedly the exposure to downstream classes could be improved - but that would be an investment.

as far as I can see, the EclipseState code works the same as proposed in this PR, but it does not export the result:

https://github.com/OPM/opm-parser/blob/master/opm/parser/eclipse/EclipseState/Tables/TableManager.cpp#L115

if you want to want to propose a better solution, I'm certainly open to merge it, but until this "better solution" is available, this patch is the best thing we have!?

joakim-hove:
> as far as I can see, the EclipseState code works the same as proposed in this PR, 

The `EclipseState` solution is way better than this - because it actually internalizes the default value of all the items in the `EQLDIMS` keyword.

> but it does not export the result:

No - that is right - I encourage you make a PR to expose it from the `TableManager` - it would take you exactly 2 minutes.

> this patch is the best thing we have!?

I must admit this patch *really* annoys me. We have said many times that we want downstream to use `EclipseState` instead of direct access to the `Deck` object as much as possible. Every time people like you @andlaus agree that that is a good idea; but when push comes to shove you always choose the path of minimum resistance and implement non-reusable hack based on direct `Deck`access. **I just don't get it.**

andlaus:
> The EclipseState solution is way better than this - because it actually internalizes the default value of all the items in the EQLDIMS keyword.

I don't really know (and care) if it is better or not. the point is that, as far as I can see, *it does not exist* at the moment! if it does, please tell me what the sanctioned way of retrieving NTEQUIL is and I'll use that.

> No - that is right - I encourage you make a PR to expose it from the TableManager - it would take you exactly 2 minutes.

it would take you the same amount of time -- and for you it would include getting it merged. ;)

> We have said many times that we want downstream to use EclipseState instead of direct access to the Deck object as much as possible.

That's the point where I guess that we start to disagree: for simple keywords, it is just IMO not worthwhile to be dogmatic about including them into the EclipseState API. (and I somehow doubt that all 365 keywords for which opm-parser currently has JSON definitions even *should* be exposed by EclipseState.) from my point of view you're welcome to have a different opinion, but you must be willing to take most of the pain, IMO.

anyway: @atgeirr: does this patch fix your issue. (I can't test because I probably don't have the incriminating deck.)

atgeirr:
I need to reopen this to get the command line arguments I need to check if it works, so please leave it open for now.

I think it is unfair to blame @andlaus for not knowing this behaviour was internalized and exposed from the EclipseState (in fact I am not sure: is it really exposed now or is it just *possible* to do it?), it is of course ok to request that is used instead of direct Deck access.

joakim-hove:
> I think it is unfair to blame @andlaus for not knowing this behaviour 

Well - I disagree; at the very least he could have asked.

andlaus:
> Well - I disagree; at the very least he could have asked.

I tend to ask via patches, I've been told that this is the most precise way ;)

anyway, I'll repeat: please tell me what the sanctioned way of retrieving NTEQUIL is and I'll use that.

joakim-hove:
> I tend to ask via patches

Well - if your patches are meant questions you should just accept the answer, which was a screaming **NO** - and don't argue it.

> I'll repeat: please tell me what the sanctioned way of retrieving NTEQUIL is and I'll use that.

As I said you will need a minor patch to expose the `Eqldims` object from the `TableManager` - when that is done you can get the item you want [here](https://github.com/OPM/opm-parser/blob/master/opm/parser/eclipse/EclipseState/Tables/Eqldims.hpp#L51)

andlaus:
> As I said you will need a minor patch to expose the Eqldims object from the TableManager - when that is done you can get the item you want here

as I said: you're calling the shots in opm-parser, so it is less effort for you to do it than it is for me.

Note that my main reason for this hesitance is also not the work involved here, but  that for most new features which we will need to fix up in the future, this would imply that a significant part of the work would become getting EclipseState changes accepted before even being in a position to start the "real" fix in the simulator. I'd prefer to avoid that situation...

anyway, this is what you asked for? OPM/opm-parser#1041

joakim-hove:
> [...] but that for most new features which we will need to fix up in the future, this would imply that a significant part of the work would become getting EclipseState changes [....]

It is hard work making a good simulator code base - we *constantly* have to fight entropy. If we are to have any chance of succeeding we must all strive to the Right &trade; thing; at all times, and in this case the Right &trade; thing was freaking obvious.

> anyway, this is what you asked for? OPM/opm-parser#1041

Looks like just what the doctor ordered!

andlaus:
> It is hard work making a good simulator code base - we constantly have to fight entropy. If we are to have any chance of succeeding we must all strive to the Right ™ thing; at all times, and in this case the Right ™ thing was freaking obvious.

We agree about having to to the "Right ™ thing", but I suppose we disagree on the "Freaking Obvious™" part ;)

andlaus:
with OPM/opm-parser#1041 merged, I've changed this PR to use the internalized Eqldims object from the table manager. now everyone should be happy?

note that currently the header for Eqldims needs to be explicitly included which I find a bit inconvenient and unintuitive (but for me this is acceptable).

joakim-hove:
> now everyone should be happy?

No objections from me at least!

andlaus:
> No objections from me at least!

good; none from me either ;)

@atgeirr: if this fixes your issue, press green.

PR #155, opened 2017-03-01
fix the two-point gradient calculator

andlaus:
for the vertex-centered finite volume discretization, the relevant position for calculating the gradient is *not* the position of the center of the *sub*-control volume, but the center of the whole control volume.

Since this patch changes the gradients of all VCFV simulations that use twopoint gradients, quite a few new reference solutions are required by this PR. I've looked at all of them manually and made sure
that they look reasonable. (The PR also includes a new reference solution for the test for the adaptive finger problem -- which uses ECFV -- but that's because the grid refinement seems to be quite sensitive to the solution proceedure for that problem, not because there's any difference in the solutions that can be noticed visually.)

note that all simulations that use element centered finite volume discretization because there sub-control volumes are identical to full control volumes. In particular, ebos -- and by extension `flow_ebos` -- is unaffected because it does not even use that code.


andlaus:

andlaus:
with jenkins and travis happy and flow_ebos being unaffected, there's no point in waiting to merge this, IMO.

PR #156, opened 2017-03-10
Added function to set SOMAX (for restarting)

babrodtk:
This adds the missing function required by OPM/opm-simulators#1087

### DIFF
--- a/ewoms/models/blackoil/blackoilmodel.hh
+++ b/ewoms/models/blackoil/blackoilmodel.hh
@@ -432,6 +432,13 @@ public:
     { return maxOilSaturation_[globalDofIdx]; }
 
     /*!
+     * \brief Sets an elements maximum oil phase saturation observed during the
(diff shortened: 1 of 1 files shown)


ewoms/models/blackoil/blackoilmodel.hh
```diff
@@ -432,6 +432,13 @@ public:
     { return maxOilSaturation_[globalDofIdx]; }
 
     /*!
+     * \brief Sets an elements maximum oil phase saturation observed during the
+     *        simulation (used for restarting from UNRST-files).
+     */
+    void setMaxOilSaturation(const Scalar& value, unsigned globalDofIdx)
```
andlaus: pass scalar arguments that are not modified by the function by value please. (whether you prefer to declare those `const` or not, I don't care too much.)

andlaus:
after removing the `&`, you can merge yourself (i.e., you don't need to wait for me)

PR #160, opened 2017-03-15
Policy change: warn when missing well/group.

joakim-hove:
See: https://github.com/OPM/opm-simulators/issues/1099


andlaus:
no objections -- merge when you consider this to be ready.

atgeirr:
I'll merge it now to keep consistency, since I just merged the legacy Flow version of this.

PR #162, opened 2017-03-16
make AD work for the vertex-centered finite volume discretization

andlaus:
this allows to use automatic differentiation in conjunction with the vertex-centered finite volume discretization; in fact, the approach can be used for basically any discretization.

By far the worst offender was the Stokes model which does not even work correctly yet because non-trivial boundary conditions cannot really be specified. Note that this is an issue that's unrelated to the AD-enablement and was up to now most likely burried by other bugs. I'll investigate this in due time. (note that "due time" may mean "never"!)

a distant second on the podium of nastiness was the Forchheimer velocity model. For this, this PR adds a test that uses AD and compares the result to the same reference solution as the test that uses finite differences, i.e., Forchheimer should be working properly.

I've also tested tested this PR with `flow_ebos` on Norne, and if anything the performance is slightly better (i.e., about 1 to 2 %), while the results seem to stay identical. the number of iterations was also different in the 1 to 2% range, though.


andlaus:

atgeirr:
This is a huge PR, and it is not clear to me which parts touch on the flow_ebos chain?

andlaus:
> which parts touch on the flow_ebos chain?

more or less the stuff that is in `ewoms/disc/{common,ecfv}` and `ewoms/models/blackoil`. I tested quite extensively that `flow_ebos` is not negatively affected by it though.

andlaus:
*grr* Dune 2.3 does not support ADL:

andlaus:

andlaus:
*grr* GCC 4.8:

andlaus:

andlaus:
alright: I've rebased this on top of the current master, made sure that the eWoms and opm-simulators unit tests continue to work (the Stokes model was hit by a bus during that process), and that the performance of flow_ebos does not regress. Since this is IMO a worthwhile enhancement, I'll merge this PR next week if there no convincing technical reasons not to do so are raised and if jenkins succeeds.

andlaus:

andlaus:

andlaus:

andlaus:
I take that silence means approval. merging.

PR #166, opened 2017-03-26
Adapt to change in ert ecl_type.

joakim-hove:
See: https://github.com/OPM/opm-output/pull/176

But - I must admit I was dumbfounded when I stumbled across this code; as it says in a comment: "This code has shamelessly been copy pasted from opm-core". Since then the code in question has been moved to opm-output and refactored massively. I would assume the copy here in ewoms code can and should be retired?

### DIFF
--- a/ebos/ertwrappers.hh
+++ b/ebos/ertwrappers.hh
@@ -38,9 +38,9 @@
 #include <ert/ecl/ecl_endian_flip.h>
 #include <ert/ecl/ecl_grid.h>
 #include <ert/ecl/ecl_kw_magic.h>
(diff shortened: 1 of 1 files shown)


andlaus:
no objections, merge at will.

> I would assume the copy here in ewoms code can and should be retired?

yes, I don't mind replacing it with stuff that uses opm-output. the reason this has been added in the first place is that the output code opm-core code was not up to speed at that time and the reason it has not been replaced yet is that this requires some work on stuff which is currently not on my list of priorities. (note that once that is done, the opm-simulators code for ecl output can probably be retired as well.)

PR #168, opened 2017-04-07
Add include of opm/common/Unused.hpp

blattms:
It is used here but not included. The problem surfaced when I
had to change inclusion order in opm-simulators for debugging.


andlaus:
thanks.

PR #170, opened 2017-04-15
ebos: hysteresis needs to be updated for all elements

andlaus:
in the parallel case, this avoids the participating processes from desyncronizing if the solution is consistent on all processes.


andlaus:
if all tests pass, and there are no other objections, I'll cherry pick this one into the release branch.

blattms:
Closes OPM/opm-simulators#1143

PR #171, opened 2017-04-18
make explicit initialization for twophase cases work

andlaus:
if the initial solution is explicitly given by the deck using the PRESSURE, SWAT, etc. keywords, the specified state can be thermodynamically impossible. To avoid inconsistencies, we use a flash calculation to find a state that is in thermodynamic equilibrium and exhibits the same masses as the explicitly specified solution. Since the flash solver needs to compute quantities for all fluid phases, but two-phase blackoil simulations usually do not specify the properties of one phase, the flash solver crashed. This patch works around that
issue by simply not using the flash solver in the twophase case, i.e., the explicit initial condition must be thermodynamically consistent in order to produce the stable results for a different choice of primary variables.

@akva2: If Jenkins is happy and you don't object, I'll also cherry-pick this into the release branch.

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -1134,8 +1134,9 @@ private:
         const auto& eclState = gridManager.eclState();
 
         // since the values specified in the deck do not need to be consistent, we use an
-        // initial condition that conserves the total mass specified by these values.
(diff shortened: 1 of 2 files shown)


andlaus:

PR #172, opened 2017-04-20
EclCpGridManager: do not delete unintialized pointers

andlaus:
this was found using GCC-7's address sanitizer. I suspect that this did not surface earlier (i.e., with valgrind), because newly allocated memory gets initialized to zero by the operating system, so the value
of the pointer was zero and the delete operator did what was right by coincidence. the new asan seems to initialize memory randomly, though.

i also tried to fix this using `std::unique_ptr`, but this seems to suffer from the same problem. maybe it is a bug in GCC-7's unique_ptr implementation.

lacking objections I'll cherry-pick this patch into the release branch.

### DIFF
--- a/ebos/eclcpgridmanager.hh
+++ b/ebos/eclcpgridmanager.hh
@@ -248,6 +248,8 @@ protected:
         // equilGrid_being a shallow copy only the global view.
         equilGrid_ = new Dune::CpGrid(*grid_);
         equilCartesianIndexMapper_ = new CartesianIndexMapper(*equilGrid_);
+
+        globalTrans_ = nullptr;
     }
 
     Grid* grid_;


atgeirr:
I think the class and fix is still dangerous: it contains new members but inherits constructors from the base class. I believe this leaves the pointers uninitialized. A better fix would be to add initializers to all the pointers at their point of definition:

```
     Grid* grid_ = nullptr;
...
```

andlaus:
the best fix would be to use `std::unique_ptr` and avoid explicit memory management entirely IMO. as mentioned, that leads to the same ASAN complaint, and explicitly initializing all unique pointers with `nullptr` strikes me as equally in-elegant as the current solution :(

atgeirr:
> the best fix would be to use std::unique_ptr

Sure, I assumed that it was not possible to use here since otherwise, well, it would have been used...?

andlaus:
Once I've got some time to spare, I'll look for a better solution. For now, let's take the minimal solution, i.e. this patch?

andlaus:
ah, I forgot:

andlaus:

blattms:
> because newly allocated memory gets initialized to zero by the
> operating system,

This will not happen (at least when compiled with opmtimization). You
might just get lucky if it had zeros in it before allocation.

andlaus:
yeah, I know it was undefined behaviour. the strange thing is that something like

```c++
std::unique_ptr<int> foo;
foo.reset(new int(123));
```

produces exactly the same complaint.

PR #173, opened 2017-04-20
fix some mild memory leaks found by ASAN

andlaus:
these leaks are not very important because these objects are de-facto singletons, i.e., they usually get allocated only once per invocation of the program.

### DIFF
--- a/ewoms/common/propertysystem.hh
+++ b/ewoms/common/propertysystem.hh
@@ -611,7 +611,7 @@ public:
         { return PropertyInfo<typename GetProperty<TypeTag, PropTag>::template GetEffectiveTypeTag_<TypeTag>::type, PropTag>::propertyName; }
     };
 
-    typedef std::list<SpliceRegistryEntryBase*> SpliceList;
+    typedef std::list<std::unique_ptr<SpliceRegistryEntryBase> > SpliceList;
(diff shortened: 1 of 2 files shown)


andlaus:

ewoms/common/propertysystem.hh
```diff
@@ -646,9 +646,9 @@ public:
     {
         std::string typeTagName = Dune::className<TypeTag>();
 
-        SpliceRegistryEntry<TypeTag, Splice1> *tmp = new SpliceRegistryEntry<TypeTag, Splice1>;
+        std::unique_ptr<SpliceRegistryEntry<TypeTag, Splice1> > tmp(new SpliceRegistryEntry<TypeTag, Splice1>);
 
-        splices_[typeTagName].push_front(tmp);
+        splices_[typeTagName].push_front(std::move(tmp));
```
blattms: Was it not you that said that moving a unique pointer is ugly?
Why not in one line without move?
```
splices_[typeTagName].emplace_front(new SpliceRegistryEntry<TypeTag, Splice1>);
``

andlaus: yeah, correct. I've changed the patch, thanks.

PR #175, opened 2017-04-27
VtkMultiphaseModule: fix incorrect loop variables

andlaus:
this is supposed to be a loop over all spatial axis, not over the fluid phases. the latest GCC from SVN rightfully complains about this. the error has not been discovered earlier because VTK velocity output is disabled by default and also, it worked "by coincidence" if the number of fluid phases was equal to the number of spatial dimensions.


andlaus:
jenkins build this please

PR #177, opened 2017-05-04
Ecl mobility shortcut fix

andlaus:
this PR contains patches for minor issues which i've stumbled over during hacking on the blackoil solvent extension. all except the last one are quite definitely harmless.

### DIFF
--- a/ebos/eclfluxmodule.hh
+++ b/ebos/eclfluxmodule.hh
@@ -119,12 +119,6 @@ class EclTransExtensiveQuantities
 
 public:
     /*!
-     * \brief Returns transmissibility for a given sub-control volume face.
-     */
-    Scalar transmissibility() const
-    { return trans_; }
-
-    /*!
      * \brief Return the intrinsic permeability tensor at a face [m^2]
      */
     const DimMatrix& intrinsicPermeability() const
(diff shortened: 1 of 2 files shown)


andlaus:

atgeirr:
The last commit requires this to be tested a little more carefully, as it changes runs. What cases have you verified (including checking that the results are equivalent)?

ebos/eclfluxmodule.hh
```diff
@@ -26,7 +26,7 @@
  * \brief This file contains the flux module which is used for ECL problems
  *
  * This approach to fluxes is very specific to two-point flux approximation and applies
- * what the Eclipse Technical Description calls the "NEWTRAN" tramsmissibilty approach.
+ * what the Eclipse Technical Description calls the "NEWTRAN" transmissibilty approach.
```
blattms: Still one i missing in transmissibility, though.

andlaus: right, thanks. that's a whack-a-mole word for typos ;)

fix pushed.

andlaus:
I've tested that the results for norne stay identical with virtually unchanged performance, and I've tested that model 2 still runs through. Lacking objections, I'll merge this tomorrow or on Monday. (whichever comes first. ;)

PR #178, opened 2017-05-05
blackoil: implement the immiscible solvent extension

andlaus:
This adds an "ECLIPSE-style" extension of the black oil model for an additional "solvent" component. Conceptually this is IMO pretty questionable, since it adds a second "gas phase" that does not mix with "ordinary" gas. I suppose the reason why this extension was conceived by E100 is that if all you have is hammer, everything looks like a nail...

Functionality-wise, this patch is still not fully complete because miscibility of the solvent "phase" is not yet implemented. As far as I can see, the API changes required by miscibility are quite limited, though.

this PR depends on OPM/opm-material#229

### DIFF
--- a/ebos/eclfluxmodule.hh
+++ b/ebos/eclfluxmodule.hh
@@ -32,6 +32,7 @@
 #define EWOMS_ECL_FLUX_MODULE_HH
 
 #include <ewoms/disc/common/fvbaseproperties.hh>
+#include <ewoms/models/blackoil/blackoilproperties.hh>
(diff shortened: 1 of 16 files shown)


andlaus:
jenkins build this with downstreams opm-material=229 please

andlaus:
since jenkins is happy and I've locally tested this with model 2 and norne (no difference in results and performance for both), I'll merge this. note that for this to have any effect some out-of-tree patches by @totto82 for opm-simulators and opm-core are required.

PR #180, opened 2017-05-09
Fixes to make the solvent module compile and run

totto82:
Tested together with solvent branch in opm-simulator for flow_ebos

### DIFF
--- a/ewoms/models/blackoil/blackoilsolventmodules.hh
+++ b/ewoms/models/blackoil/blackoilsolventmodules.hh
@@ -133,7 +133,7 @@ public:
     static void setNumSatRegions(unsigned numRegions)
     {
         ssfnKrg_.resize(numRegions);
-        ssfnKrg_.resize(numRegions);
+        ssfnKrs_.resize(numRegions);
     }
 
     /*!
@@ -164,7 +164,7 @@ public:
             // solvents have disabled at compile time
             return;
 
-        VtkBlackOilSolventModule<TypeTag>::registerParameters();
+        Ewoms::VtkBlackOilSolventModule<TypeTag>::registerParameters();
     }
 
     /*!
@@ -461,8 +461,8 @@ public:
             Evaluation Fsolgas = solventSaturation_/Stot;
 
             Evaluation& krg = asImp_().mobility_[gasPhaseIdx];
-            solventMobility_ = krg * ssfnKrs.eval(Fsolgas);
-            krg *= ssfnKrg.eval(Fhydgas);
+            solventMobility_ = krg * ssfnKrs.eval(Fsolgas, /*extrapolate=*/true);
+            krg *= ssfnKrg.eval(Fhydgas, /*extrapolate=*/true);
         }
     }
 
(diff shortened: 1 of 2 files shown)


ewoms/models/blackoil/blackoilsolventmodules.hh
```diff
@@ -164,7 +164,7 @@ public:
             // solvents have disabled at compile time
             return;
 
-        VtkBlackOilSolventModule<TypeTag>::registerParameters();
+        //VtkBlackOilSolventModule<TypeTag>::registerParameters();
     }
```
totto82: I get an error message when I don't disable the VTK output. Is this working for you?

andlaus: yes, for me it works. what is the error you get?

ewoms/models/blackoil/blackoilsolventmodules.hh
```diff
@@ -133,7 +133,7 @@ public:
     static void setNumSatRegions(unsigned numRegions)
     {
         ssfnKrg_.resize(numRegions);
-        ssfnKrg_.resize(numRegions);
+        ssfnKrs_.resize(numRegions);
```
andlaus: nice catch!

(when initializing from an ECL deck, this method is not used, though?)

ewoms/models/blackoil/blackoilsolventmodules.hh
totto82: Program threw an exception: [/home/thsa/workspace/opm/ewoms/ewoms/common/parametersystem.hh:729] Parameters can only retieved after _all_ of them have been registered.

andlaus: okay, that means that a parameter gets used but not registered. I would rather think that this error occurs if `registerParameters()` is *not* called. could you please find out which parameter it is and where it happens? (catch exceptions in gdb and go up the stack a few frames.)

also: does this also happen for plain ebos or only for flow_ebos?

totto82: `registerParameters() `tries to EWOMS_GET_PARAM(TypeTag, bool, EnableVtkOutput) and fails

I don't find 

```
if (!EWOMS_GET_PARAM(TypeTag, bool, EnableVtkOutput))
            return;
```

in the other modules `registerParameters()`

Can I remove it? (Removing it solves the problem)

ewoms/models/blackoil/blackoilsolventmodules.hh
```diff
@@ -461,8 +461,8 @@ public:
             Evaluation Fsolgas = solventSaturation_/Stot;
 
             Evaluation& krg = asImp_().mobility_[gasPhaseIdx];
-            solventMobility_ = krg * ssfnKrs.eval(Fsolgas);
-            krg *= ssfnKrg.eval(Fhydgas);
+            solventMobility_ = krg * ssfnKrs.eval(Fsolgas, /*extrapolate=*/true);
```
andlaus: while this change is correct, isn't there a problem with the deck if the properties specified therein must be extrapolated?

andlaus:
okay, this PR looks good enough for me now. merging, thanks!

ewoms/models/blackoil/blackoilsolventmodules.hh
totto82: You may get slightly negative solvent and gas saturations. Like -1e-20. SSFN is strictly [0,1]

totto82: I have experienced a bit with truncating saturations but it seems Newton doesn't like it. i.e. the number of non-linear iterations may increase.

andlaus: ah, okay. no problem then.

PR #181, opened 2017-05-12
Use volume weighted arithmetic average of NTG for cells merged by minpv.

totto82:
Make the TRAN[X,Y] like flow_legacy and a bit more like Eclipse. 

With this PR TRAN[XYZ] matches whats computed in geoProps within acceptable values on Norne.

### DIFF
--- a/ebos/ecltransmissibility.hh
+++ b/ebos/ecltransmissibility.hh
@@ -108,6 +108,10 @@ public:
         const std::vector<double>& ntg =
             eclState.get3DProperties().getDoubleGridProperty("NTG").getData();
 
+        // modify the ntg values for the cells merged with minpv
+        std::vector<double> ntg_mod (ntg);
+        minPvFillNtg_(ntg_mod);
+
         unsigned numElements = elemMapper.size();
 
         extractPermeability_();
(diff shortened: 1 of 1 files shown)


andlaus:

andlaus:
thanks. the patch looks a bit too complex for me, but I cannot immediately see any better solution. I'll merge if jenkins is happy. (travis fails, but I have trouble to understand why just by looking at the log.)

akva2:
there is generic libecl breakage in travis.

andlaus:
> there is generic libecl breakage in travis.

good to know. I suspected that the error was not related to this PR...

andlaus:
jenkins is happy, yay! merging!

ebos/ecltransmissibility.hh
```diff
@@ -398,6 +402,44 @@ private:
...
+        const auto& eclGrid = eclState.getInputGrid();
+        const auto& porv = eclState.get3DProperties().getDoubleGridProperty("PORV").getData();
+        const auto& actnum = eclState.get3DProperties().getIntGridProperty("ACTNUM").getData();
+        const auto& cartMapper = gridManager_.cartesianIndexMapper();
+        const auto& cartDims = cartMapper.cartesianDimensions();
+        assert(dimWorld > 1);
+        const size_t nxny = cartDims[0] * cartDims[1];
+
+        for (size_t cartesianCellIdx = 0; cartesianCellIdx < ntg.size(); ++cartesianCellIdx) {
+            const double cellVolume = eclGrid.getCellVolume(cartesianCellIdx);
```
blattms: I know this is merged already, but still:

Would it change anything to exclude non-active cells by adding:
```
if( ! actnum[cartesianCellIdx) continue;
```

Let us say I have 3 cells connect along the z-direction with a too small pore volume. When we reach the middle cell, well change its ntg value by adding `ntg[cartesianCellIdxAbove]*cellAboveVolume`. But will we not add the volume scaled ntg value of the uppermost cell twice for the lower cell when that is reached in the for loop? Am I missing something or is that simply no problem?

If it is a problem we should change `while` to `if`.

totto82: I think you are right. It should be` if`. And yes non-active cells should be excluded to avoid unnecessary calculations. The `while` to `if `change should also be fixed in flow_legacy. Thanks for noticing.

totto82: Fix in #182

blattms:
Cool, thanks!

PR #182, opened 2017-05-15
Fix minpvProps calculations

totto82:
Use the unmodified ntg values to get the "correct" values when more than 2 cells are merged together using minPv. This makes the code equivalent with the code in Geoprops.

### DIFF
--- a/ebos/ecltransmissibility.hh
+++ b/ebos/ecltransmissibility.hh
@@ -105,12 +105,9 @@ public:
         auto& transMult = eclState.getTransMult();
         ElementMapper elemMapper(gridView);
 
-        const std::vector<double>& ntg =
-            eclState.get3DProperties().getDoubleGridProperty("NTG").getData();
-
-        // modify the ntg values for the cells merged with minpv
-        std::vector<double> ntg_mod (ntg);
-        minPvFillNtg_(ntg_mod);
+        // get the ntg values, the ntg values are modified for the cells merged with minpv
+        std::vector<double> ntg;
+        minPvFillNtg_(ntg);
 
         unsigned numElements = elemMapper.size();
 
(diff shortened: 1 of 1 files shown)


ebos/ecltransmissibility.hh
```diff
@@ -402,37 +402,49 @@ private:
...
-            ntg[cartesianCellIdx] *= cellVolume;
-            double totalCellVolume = cellVolume;
+
+            // don't do anything for inactive cells
+            if( ! actnum[cartesianCellIdx])
+                continue;
+
+            // don't do anything if the cell above has porv > minPVvalue.
+            int cartesianCellIdxAbove = cartesianCellIdx - nxny;
+            if (porv[cartesianCellIdxAbove] > eclGrid.getMinpvValue() )
```
andlaus: skipping the cells which exhibit a PV less than the minimum work should be the same and is IMO more streamlined. use

```c++
// don't do anything if the cell has a pore volume smaller than the minimum one
if (porv[cartesianCellIdx] < eclGrid.getMinpvValue())
  continue;
```

andlaus: also, explicitly initializing `ntg_avg` (or `averageNtg`?) to 1.0 even for deactivated cells might be a good idea.

totto82:
```
> // don't do anything if the cell has a pore volume smaller than the minimum one
> if (porv[cartesianCellIdx] < eclGrid.getMinpvValue())
>   continue;
```
This is not the same as `if (porv[cartesianCellIdxAbove] > eclGrid.getMinpvValue()` 

We could add your suggestion as well to avoid some extra calculations, but I guess the number of cells with porv < minpv should be very small.

andlaus: > We could add your suggestion as well to avoid some extra calculations, but I guess the number of cells with porv < minpv should be very small.

the point was not so much the extra calculations, but that looking at the cell above here makes things harder to understand? (if you disagree, let's leave it as-is.)

ebos/ecltransmissibility.hh
```diff
@@ -402,37 +399,55 @@ private:
...
         const auto& actnum = eclState.get3DProperties().getIntGridProperty("ACTNUM").getData();
+        const std::vector<double>& ntg =
+            eclState.get3DProperties().getDoubleGridProperty("NTG").getData();
+
         const auto& cartMapper = gridManager_.cartesianIndexMapper();
         const auto& cartDims = cartMapper.cartesianDimensions();
         assert(dimWorld > 1);
         const size_t nxny = cartDims[0] * cartDims[1];
 
+        averageNtg.resize(ntg.size());
```
andlaus: I'll sleep better if you add `std::fll(averageNtg.begin(), averageNtg.end(), 1.0);` here.

PR #184, opened 2017-05-23
Avoid negative indicies in minPvFillNtg_

totto82:


andlaus:
thanks. merging!

andlaus:
on a closer look, the NTG value for the cells in the topmost layer is always `1.0` with this patch?

I guess that this is best fixed by changing

```c++
averageNtg.clear();
averageNtg.resize(ntg.size(),1.0);
```

to

```c++
averageNtg = ntg;
```

andlaus:
forget what I just said, I didn't consider line 425.

andlaus:
correction of correction: line 425 only applies for inactive cells. so the fix proposed above would also allow to get rid of that if.

PR #185, opened 2017-05-23
Add support for miscible solvent runs

totto82:
- initialization only via Ecl Deck
- adds modification of relperms and cap-pressures for miscible runs
- adds TL mixing model for effective viscosity and density calculations

Tested on SPE1CASE2_SOLVENT_MISC_TL, SPE9_CP_CO2_MISC_TL, SPE5CASE1-3
and SPE3CASE1_SOLVENT_MISC

Issues:
- some discrepancy of the results for the SPE3 case compared to
flow_solvent.
- some extra convergence issues on the SPE5s and SPE9.

Issues most probably related to the well handling.

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -779,6 +779,26 @@ public:
     }
 
     /*!
+     * \brief Returns the index of the relevant region for thermodynmic properties
+     */
+    template <class Context>
+    unsigned miscnumRegionIndex(const Context& context, unsigned spaceIdx, unsigned timeIdx) const
+    { return miscnumRegionIndex(context.globalSpaceIndex(spaceIdx, timeIdx)); }
+
+    /*!
+     * \brief Returns the index the relevant MISC region given a cell index
+     */
+    unsigned miscnumRegionIndex(unsigned elemIdx) const
+    {
+        if (miscnum_.empty())
+            return 0;
+
+        return miscnum_[elemIdx];
+    }
+
+
+
+    /*!
      * \copydoc FvBaseProblem::name
      */
     std::string name() const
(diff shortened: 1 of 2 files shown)


andlaus:

andlaus:
nice! I've skimmed over the patch and nothing obviously wrong hit me. are you aware of any issues which need to be fixed before merging or do you think it is ready?

totto82:
I have a few more minor fixes I want to include before it is ready. 
Any idea why Jenkins is unhappy?

totto82:
Now all the issues reported in the PR is fixed. If Jenkins is happy this can be merged.

andlaus:
> Any idea why Jenkins is unhappy?

the `compareECLInitFiles_flow_ebos+NORNE_ATW2013` test from opm-simulators fails. probably not because this PR, though?

andlaus:
hm, that issue also occurs locally on my machine:

```
32: Program threw an exception: [/home/and/src/opm-output/opm/test_util/EclFilesComparator.cpp:370] Deviations exceed tolerances.
32: The absolute deviation is 0.0301113, and the tolerance limit is 0.02.
32: The relative deviation is 0.00106973, and the tolerance limit is 1e-05.
32: Comparing '/home/and/src/opm-simulators/tests/results/init/flow_ebos+norne/NORNE_ATW2013' to '/home/and/src/opm-simulators/../opm-data/norne/opm-simulation-reference/flow_ebos/NORNE_ATW2013'.
```

could #184 be the culprit?

andlaus:
ok, seems like #184 *was* the culprit. fixed by #186.

totto82:
The #184 may change TRANX. I noticed Jenkins was never asked for #184. Maybe reverting #184, making a new PR and then ask Jenkins?

totto82:
I guess you already did the reverting part in #186. We need the fix in #184. Should I make a new PR?

andlaus:
I hope that the fix for the negative cartesian indices is still present in #186 because cartesian indices should not be used outside of the cell volume accumulation while loop anymore.

andlaus:

andlaus:
jenkins is happy at last. merging!

PR #187, opened 2017-06-06
Write extrusion factor

andlaus:
this PR contains a minor cleanup of the FV base intensive quantities to reduce of manual fiddling if changing the attributes and -- more importantly -- it enables to write the extrusion factor of DOFs to the VTK files so that they can be inspected visually.


andlaus:
jenkins build this please

andlaus:
since jenkins is happy and travis hit the recent opm-parser buildsystem snack, I'll merge this now.

PR #190, opened 2017-06-14
Update cmake version when running travis.

joakim-hove:


andlaus:
looks a bit like the second patch slipped in here by accident. else, no objections from my side, i.e., merge at will.

joakim-hove:
> looks a bit like the second patch slipped in here by accident.

Ehhh yes  - that is right. The second commit is this PR: #198

PR #192, opened 2017-06-14
ebos: fix the compilation

### LINKED ISSUE #191: PR #189 broke flow_ebos.
See my comment in #189.

andlaus:
one must use eclProps.getDoubleGridProperty() while the has() method is called eclProps.hasDeckDoubleGridProperty() (mind the Deck!).

this fixes #191.


andlaus:
I've tested that this works locally and jenkins will take too long until it reaches opm-simulator, so I'll merge this immediately.

PR #195, opened 2017-06-15
Add polymer module in ebos

totto82:
Adds a conservation equation for polymer.
Polymer concentration in the water phase is used as primary variable
The polymer influences the viscosity of the water, and leaves gas and oil
unaffected. 

Shear effects can be modeled using the PLYSHLOG and SHRATE keywords.

### DIFF
--- a/ewoms/models/blackoil/blackoilintensivequantities.hh
+++ b/ewoms/models/blackoil/blackoilintensivequantities.hh
@@ -31,6 +31,8 @@
 #include "blackoilproperties.hh"
 #include "blackoilfluidstate.hh"
 #include "blackoilsolventmodules.hh"
+#include "blackoilpolymermodules.hh"
+
 
 #include <opm/material/fluidstates/CompositionalFluidState.hpp>
 #include <opm/common/Valgrind.hpp>
@@ -53,6 +55,7 @@ class BlackOilIntensiveQuantities
     : public GET_PROP_TYPE(TypeTag, DiscIntensiveQuantities)
     , public GET_PROP_TYPE(TypeTag, FluxModule)::FluxIntensiveQuantities
     , public BlackOilSolventIntensiveQuantities<TypeTag>
+    , public BlackOilPolymerIntensiveQuantities<TypeTag>
 {
     typedef typename GET_PROP_TYPE(TypeTag, DiscIntensiveQuantities) ParentType;
     typedef typename GET_PROP_TYPE(TypeTag, IntensiveQuantities) Implementation;
@@ -68,6 +71,7 @@ class BlackOilIntensiveQuantities
     typedef typename GET_PROP_TYPE(TypeTag, FluxModule) FluxModule;
 
     enum { enableSolvent = GET_PROP_VALUE(TypeTag, EnableSolvent) };
+    enum { enablePolymer = GET_PROP_VALUE(TypeTag, EnablePolymer) };
     enum { numPhases = GET_PROP_VALUE(TypeTag, NumPhases) };
     enum { numComponents = GET_PROP_VALUE(TypeTag, NumComponents) };
     enum { waterCompIdx = FluidSystem::waterCompIdx };
@@ -148,6 +152,8 @@ public:
         fluidState_.setSaturation(oilPhaseIdx, So);
 
         asImp_().solventPreSatFuncUpdate_(elemCtx, dofIdx, timeIdx);
+        asImp_().polymerPreSatFuncUpdate_(elemCtx, dofIdx, timeIdx);
+
 
         // now we compute all phase pressures
         Evaluation pC[numPhases];
@@ -173,6 +179,8 @@ public:
         Opm::Valgrind::CheckDefined(mobility_);
 
         asImp_().solventPostSatFuncUpdate_(elemCtx, dofIdx, timeIdx);
+        asImp_().polymerPostSatFuncUpdate_(elemCtx, dofIdx, timeIdx);
+
 
         Scalar SoMax = elemCtx.model().maxOilSaturation(globalSpaceIdx);
 
@@ -308,6 +316,8 @@ public:
         }
 
         asImp_().solventPvtUpdate_(elemCtx, dofIdx, timeIdx);
+        asImp_().polymerPropertiesUpdate_(elemCtx, dofIdx, timeIdx);
+
 
         // update the quantities which are required by the chosen
         // velocity model
@@ -376,6 +386,8 @@ public:
 
 private:
     friend BlackOilSolventIntensiveQuantities<TypeTag>;
+    friend BlackOilPolymerIntensiveQuantities<TypeTag>;
+
 
     Implementation& asImp_()
     { return *static_cast<Implementation*>(this); }
--- a/ewoms/models/blackoil/blackoilintensivequantities.hh
+++ b/ewoms/models/blackoil/blackoilintensivequantities.hh
@@ -152,8 +152,6 @@ public:
         fluidState_.setSaturation(oilPhaseIdx, So);
 
         asImp_().solventPreSatFuncUpdate_(elemCtx, dofIdx, timeIdx);
-        asImp_().polymerPreSatFuncUpdate_(elemCtx, dofIdx, timeIdx);
-
 
         // now we compute all phase pressures
         Evaluation pC[numPhases];
@@ -178,8 +176,9 @@ public:
         MaterialLaw::relativePermeabilities(mobility_, materialParams, fluidState_);
         Opm::Valgrind::CheckDefined(mobility_);
 
+        // update the Saturation functions for the blackoil extensions.
         asImp_().solventPostSatFuncUpdate_(elemCtx, dofIdx, timeIdx);
-        asImp_().polymerPostSatFuncUpdate_(elemCtx, dofIdx, timeIdx);
+        asImp_().polymerSatFuncUpdate_(elemCtx, dofIdx, timeIdx);
 
 
         Scalar SoMax = elemCtx.model().maxOilSaturation(globalSpaceIdx);
--- a/ewoms/models/blackoil/blackoilintensivequantities.hh
+++ b/ewoms/models/blackoil/blackoilintensivequantities.hh
@@ -176,10 +176,8 @@ public:
         MaterialLaw::relativePermeabilities(mobility_, materialParams, fluidState_);
         Opm::Valgrind::CheckDefined(mobility_);
 
-        // update the Saturation functions for the blackoil extensions.
+        // update the Saturation functions for the blackoil solvent module.
         asImp_().solventPostSatFuncUpdate_(elemCtx, dofIdx, timeIdx);
-        asImp_().polymerSatFuncUpdate_(elemCtx, dofIdx, timeIdx);
-
 
         Scalar SoMax = elemCtx.model().maxOilSaturation(globalSpaceIdx);
 
--- a/ewoms/models/blackoil/blackoilpolymermodules.hh
+++ b/ewoms/models/blackoil/blackoilpolymermodules.hh
@@ -29,7 +29,7 @@
 #define EWOMS_BLACK_OIL_POLYMER_MODULE_HH
 
 #include "blackoilproperties.hh"
- // #include <ewoms/io/vtkblackoilpolymermodule.hh>
+#include <ewoms/io/vtkblackoilpolymermodule.hh>
 #include <ewoms/models/common/quantitycallbacks.hh>
 
 #include <opm/material/common/Tabulated1DFunction.hpp>
@@ -365,10 +365,10 @@ public:
     static void registerParameters()
     {
         if (!enablePolymer)
-            // polymers have disabled at compile time
+            // polymers have been disabled at compile time
             return;
 
-        //Ewoms::VtkBlackOilPolymerModule<TypeTag>::registerParameters();
+        Ewoms::VtkBlackOilPolymerModule<TypeTag>::registerParameters();
     }
 
     /*!
@@ -378,16 +378,16 @@ public:
                                       Simulator& simulator)
     {
         if (!enablePolymer)
-            // polymers have disabled at compile time
+            // polymers have been disabled at compile time
             return;
 
-        //model.addOutputModule(new Ewoms::VtkBlackOilPolymerModule<TypeTag>(simulator));
+        model.addOutputModule(new Ewoms::VtkBlackOilPolymerModule<TypeTag>(simulator));
     }
 
     static bool primaryVarApplies(unsigned pvIdx)
     {
         if (!enablePolymer)
-            // polymers have disabled at compile time
+            // polymers have been disabled at compile time
             return false;
 
         return pvIdx == polymerConcentrationIdx;
--- a/ewoms/models/blackoil/blackoilpolymermodules.hh
+++ b/ewoms/models/blackoil/blackoilpolymermodules.hh
@@ -1,0 +1,1148 @@
+// -*- mode: C++; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*-
+// vi: set et ts=4 sw=4 sts=4:
+/*
+  This file is part of the Open Porous Media project (OPM).
+
+  OPM is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 2 of the License, or
+  (at your option) any later version.
+
+  OPM is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with OPM.  If not, see <http://www.gnu.org/licenses/>.
+
+  Consult the COPYING file in the top-level source directory of this
+  module for the precise wording of the license and the list of
+  copyright holders.
+*/
+/*!
+ * \file
+ *
+ * \brief Contains the classes required to extend the black-oil model by polymer.
+ */
+#ifndef EWOMS_BLACK_OIL_POLYMER_MODULE_HH
+#define EWOMS_BLACK_OIL_POLYMER_MODULE_HH
+
+#include "blackoilproperties.hh"
+ // #include <ewoms/io/vtkblackoilpolymermodule.hh>
+#include <ewoms/models/common/quantitycallbacks.hh>
+
+#include <opm/material/common/Tabulated1DFunction.hpp>
+
+#if HAVE_OPM_PARSER
+#include <opm/parser/eclipse/Deck/Deck.hpp>
+#include <opm/parser/eclipse/EclipseState/EclipseState.hpp>
+#include <opm/parser/eclipse/EclipseState/Tables/PlyadsTable.hpp>
+#include <opm/parser/eclipse/EclipseState/Tables/PlymaxTable.hpp>
+#include <opm/parser/eclipse/EclipseState/Tables/PlyrockTable.hpp>
+#include <opm/parser/eclipse/EclipseState/Tables/PlyshlogTable.hpp>
+#include <opm/parser/eclipse/EclipseState/Tables/PlyviscTable.hpp>
+#include <opm/parser/eclipse/EclipseState/Tables/PlyshlogTable.hpp>
+#endif
+
+#include <opm/common/Valgrind.hpp>
+#include <opm/common/Unused.hpp>
+#include <opm/common/ErrorMacros.hpp>
+#include <opm/common/Exceptions.hpp>
+
+#include <dune/common/fvector.hh>
+
+#include <string>
+
+namespace Ewoms {
+/*!
+ * \ingroup BlackOil
+ * \brief Contains the high level supplements required to extend the black oil
+ *        model by polymer.
+ */
+template <class TypeTag, bool enablePolymerV = GET_PROP_VALUE(TypeTag, EnablePolymer)>
+class BlackOilPolymerModule
+{
+    typedef typename GET_PROP_TYPE(TypeTag, Scalar) Scalar;
+    typedef typename GET_PROP_TYPE(TypeTag, Evaluation) Evaluation;
+    typedef typename GET_PROP_TYPE(TypeTag, PrimaryVariables) PrimaryVariables;
+    typedef typename GET_PROP_TYPE(TypeTag, IntensiveQuantities) IntensiveQuantities;
+    typedef typename GET_PROP_TYPE(TypeTag, ExtensiveQuantities) ExtensiveQuantities;
+    typedef typename GET_PROP_TYPE(TypeTag, ElementContext) ElementContext;
+    typedef typename GET_PROP_TYPE(TypeTag, FluidSystem) FluidSystem;
+    typedef typename GET_PROP_TYPE(TypeTag, Model) Model;
+    typedef typename GET_PROP_TYPE(TypeTag, Simulator) Simulator;
+    typedef typename GET_PROP_TYPE(TypeTag, EqVector) EqVector;
+    typedef typename GET_PROP_TYPE(TypeTag, RateVector) RateVector;
+    typedef typename GET_PROP_TYPE(TypeTag, Indices) Indices;
+
+    typedef Opm::MathToolbox<Evaluation> Toolbox;
+
+    typedef typename Opm::Tabulated1DFunction<Scalar> TabulatedFunction;
+
+    static constexpr unsigned polymerConcentrationIdx = Indices::polymerConcentrationIdx;
+    static constexpr unsigned contiPolymerEqIdx = Indices::contiPolymerEqIdx;
+    static constexpr unsigned contiWaterEqIdx = Indices::conti0EqIdx + FluidSystem::waterCompIdx ;
+    static constexpr unsigned waterPhaseIdx = FluidSystem::waterPhaseIdx;
+
+
+    static constexpr unsigned enablePolymer = enablePolymerV;
+    static constexpr unsigned numEq = GET_PROP_VALUE(TypeTag, NumEq);
+    static constexpr unsigned numPhases = FluidSystem::numPhases;
+
+public:
+    enum AdsorptionBehaviour { Desorption = 1, NoDesorption = 2 };
+
+#if HAVE_OPM_PARSER
+    /*!
+     * \brief Initialize all internal data structures needed by the polymer module
+     */
+    static void initFromDeck(const Opm::Deck& deck, const Opm::EclipseState& eclState)
+    {
+        // some sanity checks: if polymers are enabled, the POLYMER keyword must be
+        // present, if polymers are disabled the keyword must not be present.
+        if (enablePolymer && !deck.hasKeyword("POLYMER")) {
+            OPM_THROW(std::runtime_error,
+                      "Non-trivial polymer treatment requested at compile time, but "
+                      "the deck does not contain the POLYMER keyword");
+        }
+        else if (!enablePolymer && deck.hasKeyword("POLYMER")) {
+            OPM_THROW(std::runtime_error,
+                      "Polymer treatment disabled at compile time, but the deck "
+                      "contains the POLYMER keyword");
+        }
+
+        if (!deck.hasKeyword("POLYMER"))
+            return; // polymer treatment is supposed to be disabled
+
+        const auto& tableManager = eclState.getTableManager();
+
+        unsigned numSatRegions = tableManager.getTabdims().getNumSatTables();
+        setNumSatRegions(numSatRegions);
+
+        // initialize the objects which deal with the PLYROCK keyword
+        const auto& plyrockTables = tableManager.getPlyrockTables();
+        if (!plyrockTables.empty()) {
+            assert(numSatRegions == plyrockTables.size());
+            for (unsigned satRegionIdx = 0; satRegionIdx < numSatRegions; ++ satRegionIdx) {
+                const auto& plyrockTable = plyrockTables.template getTable<Opm::PlyrockTable>(satRegionIdx);
+                setPlyrock(satRegionIdx,
+                           plyrockTable.getDeadPoreVolumeColumn()[satRegionIdx],
+                           plyrockTable.getResidualResistanceFactorColumn()[satRegionIdx],
+                           plyrockTable.getRockDensityFactorColumn()[satRegionIdx],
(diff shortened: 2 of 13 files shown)


totto82:
Update: Solve the shear multiplier equation in the log space directly to avoid usage of expensive log and exp. evaluations. Speed up: 15%

GitPaean:
> Update: Solve the shear multiplier equation in the log space directly to avoid usage of expensive log and exp. evaluations. Speed up: 15%

What is this about? One specific commit or something?

totto82:
I just amended the previous commit.

ewoms/models/blackoil/blackoilpolymermodules.hh
```diff
@@ -0,0 +1,1148 @@
...
+        } else {
+            OPM_THROW(std::runtime_error, "PLMIXPAR must be specified in POLYMER runs\n");
+        }
+
+        hasPlyshlog_ = deck.hasKeyword("PLYSHLOG");
+        hasShrate_ = deck.hasKeyword("SHRATE");
+
+        if (hasPlyshlog_) {
+            const auto& plyshlogTables = tableManager.getPlyshlogTables();
+            assert(numPvtRegions == plyshlogTables.size());
```
andlaus: maybe this should be numMixRegions, but I'm not sure...

ewoms/models/blackoil/blackoilpolymermodules.hh
```diff
@@ -0,0 +1,1148 @@
...
+            for (unsigned pvtRegionIdx = 0; pvtRegionIdx < numPvtRegions; ++ pvtRegionIdx) {
+                const auto& plyshlogTable = plyshlogTables.template getTable<Opm::PlyshlogTable>(pvtRegionIdx);
+
+                Scalar plyshlogRefPolymerConcentration = plyshlogTable.getRefPolymerConcentration();
+                std::vector<double> waterVelocity = plyshlogTable.getWaterVelocityColumn().vectorCopy();
+                std::vector<double> shearMultiplier = plyshlogTable.getShearMultiplierColumn().vectorCopy();
+
+                // do the unit version here for the waterVelocity
+                Opm::UnitSystem unitSystem = deck.getActiveUnitSystem();
+                double siFactor = hasShrate_? unitSystem.parse("1/Time").getSIScaling() : unitSystem.parse("Length/Time").getSIScaling();
```
andlaus: please use `Scalar` instead of `double` here.

ewoms/models/blackoil/blackoilpolymermodules.hh
```diff
@@ -0,0 +1,1148 @@
...
+                for (size_t i = 0; i < waterVelocity.size(); ++i ) {
+                    waterVelocity[i] *= siFactor;
+                    waterVelocity[i] = std::log(waterVelocity[i]);
```
andlaus: the `std::log()` seems a bit strange. some comment why this is done here would be good.

ewoms/models/blackoil/blackoilpolymermodules.hh
```diff
@@ -0,0 +1,1148 @@
...
+                }
+            }
+        }
+
+        if (hasShrate_) {
+            if(!hasPlyshlog_) {
+                OPM_THROW(std::runtime_error, "PLYSHLOG must be specified if SHRATE is used in POLYMER runs\n");
+            }
+            const auto& shrateKeyword = deck.getKeyword("SHRATE");
+            std::vector<double> shrateFromDeck = shrateKeyword.getSIDoubleData();
```
andlaus: use `const std::vector<double>&` here. the performance impact will be pretty tiny in this instance, but it usually pays to have good habits...

ewoms/models/blackoil/blackoilpolymermodules.hh
```diff
@@ -0,0 +1,1148 @@
...
+            shrate_.resize(numPvtRegions);
+            for (unsigned pvtRegionIdx = 0; pvtRegionIdx < numPvtRegions; ++ pvtRegionIdx) {
+                if (shrateFromDeck.size() == 0) {
```
andlaus: alternatively, you can use `shrateFromDeck.empty()` for the condition.

ewoms/models/blackoil/blackoilpolymermodules.hh
```diff
@@ -0,0 +1,1148 @@
...
+
+    /*!
+     * \brief Specify the number of mix regions.
+     *
+     * This must be called before setting the PLYMAC and PLMIXPAR of any region.
+     */
+    static void setNumMixRegions(unsigned numRegions)
+    {
+        plymaxCmax_.resize(numRegions);
+        plymixparTL_.resize(numRegions);
```
andlaus: please rename these two to `plymaxMaxConcentration_` and `plymixparToddLongstaff_`. abbreviations may be obvious to you, but to someone new to the code they tend to be very confusing.

ewoms/models/blackoil/blackoilpolymermodules.hh
```diff
@@ -0,0 +1,1148 @@
...
+            return false;
+
+        return pvIdx == polymerConcentrationIdx;
+    }
+
+    static std::string primaryVarName(unsigned pvIdx OPM_OPTIM_UNUSED)
+    {
+        assert(primaryVarApplies(pvIdx));
+
+        return "saturation_polymer";
```
andlaus: that should be `concentration_polymer` or maybe even `polymer_waterconcentration`!?

ewoms/models/blackoil/blackoilpolymermodules.hh
```diff
@@ -0,0 +1,1148 @@
...
+                    Opm::decay<Scalar>(extQuants.waterShearFactor());
+        }
+
+    }
+
+    /*!
+     * \brief Assign the polymer specific primary variables to a PrimaryVariables object
+     */
+    static void assignPrimaryVars(PrimaryVariables& priVars,
+                                  Scalar polymerSaturation)
```
andlaus: polymerConcentration

ewoms/models/blackoil/blackoilpolymermodules.hh
```diff
@@ -0,0 +1,1148 @@
...
+
+    static const Scalar& plyrockDeadPoreVolume(const ElementContext& elemCtx,
+                                            unsigned scvIdx,
+                                            unsigned timeIdx)
+    {
+        unsigned satnumRegionIdx = elemCtx.problem().satnumRegionIndex(elemCtx, scvIdx, timeIdx);
+        return plyrockDeadPoreVolume_[satnumRegionIdx];
+    }
+
+    static const Scalar& plyrockResidualResistanceFactor(const ElementContext& elemCtx,
```
andlaus: please return primitive floating point types (`Scalar`s) by value, not as const references. (this also happens below and above.)

ewoms/models/blackoil/blackoilpolymermodules.hh
```diff
@@ -0,0 +1,1148 @@
...
+     * Input is polymer concentration and either the water velocity or the shrate if has_shrate_ is true.
+     * The pvtnumRegionIdx is needed to make sure the right table is used.
+     */
+    template <class Evaluation>
+    static Evaluation computeShearFactor(const Evaluation& polymerConcentration,
+                                   unsigned pvtnumRegionIdx,
+                                   const Evaluation& v0) {
+
+        const auto& viscosityMultiplierTable = plyviscViscosityMultiplier_[pvtnumRegionIdx];
+        Scalar viscosityMultiplier =  Opm::scalarValue( viscosityMultiplierTable.eval(polymerConcentration, /*extrapolate=*/true) );
```
andlaus: change this to `viscosityMultiplierTable.eval(Opm::scalarValue(polymerConcentration), /*extrapolate=*/true)` instead of evaluating everything with the derivatives and immediately throwing the derivatives of the result away.

ewoms/models/blackoil/blackoilpolymermodules.hh
```diff
@@ -0,0 +1,1148 @@
...
+
+        const Scalar eps = 1e-14;
+        // return 1.0 if the polymer has no effect on the water.
+        if ( std::abs( (viscosityMultiplier - 1.0) ) < eps){
+            return Evaluation(1.0);
```
andlaus: `return 1.0` should be equivalent here.

ewoms/models/blackoil/blackoilpolymermodules.hh
```diff
@@ -0,0 +1,1148 @@
...
+        }
+
+        const std::vector<Scalar>& shearEffectRefLogVelocity = plyshlogShearEffectRefLogVelocity_[pvtnumRegionIdx];
+        auto v0AbsLog = Opm::log(Opm::abs(v0));
+        // return 1.0 if the velocity /sharte is smaller than the first velocity entry.
+        if (v0AbsLog.value() < shearEffectRefLogVelocity[0])
```
andlaus: `v0AbsLog < shearEffectRefLogVelocity[0]` is identical and also works if v0AbsLog is a Scalar...

ewoms/models/blackoil/blackoilpolymermodules.hh
```diff
@@ -0,0 +1,1148 @@
...
+        const std::vector<Scalar>& shearEffectRefMultiplier = plyshlogShearEffectRefMultiplier_[pvtnumRegionIdx];
+        int numTableEntries = shearEffectRefLogVelocity.size();
+        assert (shearEffectRefMultiplier.size() == numTableEntries);
+
+        std::vector<Scalar> shearEffectMultiplier(numTableEntries, 1.0);
+        for (int i = 0; i < numTableEntries; ++i) {
+            shearEffectMultiplier[i] = (1.0 + (viscosityMultiplier - 1.0)*shearEffectRefMultiplier[i]) / viscosityMultiplier;
+            shearEffectMultiplier[i] = Opm::log(shearEffectMultiplier[i]);
+        }
+        // store the log velocity and log multipliers in a table for easy look up and
```
andlaus: full sentences, please: "store the logarithm of the..."

ewoms/models/blackoil/blackoilpolymermodules.hh
```diff
@@ -0,0 +1,1148 @@
...
+        // linear interpolation in the log space.
+        TabulatedFunction logShearEffectMultiplier = TabulatedFunction(numTableEntries, shearEffectRefLogVelocity, shearEffectMultiplier );
```
andlaus: this seems quite slow: it creates the `logShearEffectMultiplier` object for every call of the method. it should probably be made an attribute of the class and only be initialized once during construction.

ewoms/models/blackoil/blackoilpolymermodules.hh
```diff
@@ -0,0 +1,1148 @@
...
+
+        // Find sheared velocity (v) that satisfies
+        // F = log(v) + log (Z) - log(v0) = 0;
+
+        // Set up the function
+        // u = log(v)
```
andlaus: isn't this `f(u) = u + log(M(u)) - abs(log(v0))`?

ewoms/models/blackoil/blackoilpolymermodules.hh
```diff
@@ -0,0 +1,1148 @@
...
+    void polymerPostSatFuncUpdate_(const ElementContext& elemCtx,
+                                   unsigned dofIdx,
+                                   unsigned timeIdx)
+    {
+        const PrimaryVariables& priVars = elemCtx.primaryVars(dofIdx, timeIdx);
+        polymerConcentration_ = priVars.makeEvaluation(polymerConcentrationIdx, timeIdx);
+        const Scalar cmax = PolymerModule::plymaxCmax(elemCtx, dofIdx, timeIdx);
+
+        // permeability reduction due to polymer
+        const Scalar& maxAdsorbtion = PolymerModule::plyrockMaxAdsorbtion(elemCtx, dofIdx, timeIdx);
```
andlaus: no reference here, please!

ewoms/models/blackoil/blackoilpolymermodules.hh
```diff
@@ -0,0 +1,1148 @@
...
+        const auto& plyadsAdsorbedPolymer = PolymerModule::plyadsAdsorbedPolymer(elemCtx, dofIdx, timeIdx);
+        polymerAdsorption_ = plyadsAdsorbedPolymer.eval(polymerConcentration_, /*extrapolate=*/true);
+        if (PolymerModule::plyrockAdsorbtionIndex(elemCtx, dofIdx, timeIdx) == PolymerModule::NoDesorption ) {
+                const Scalar& maxPolymerAdsorption = elemCtx.problem().maxPolymerAdsorption(elemCtx, dofIdx, timeIdx);
+                polymerAdsorption_ = std::max(Evaluation(maxPolymerAdsorption) , polymerAdsorption_);
```
andlaus: I guess this should be `Opm::max(maxPolymerAdsorption, polymerAdsorption_)`

ewoms/models/blackoil/blackoilpolymermodules.hh
```diff
@@ -0,0 +1,1148 @@
...
+        }
+
+        // compute resitanceFactor
+        const Scalar& residualResistanceFactor = PolymerModule::plyrockResidualResistanceFactor(elemCtx, dofIdx, timeIdx);
```
andlaus: scalar reference again.

ewoms/models/blackoil/blackoilpolymermodules.hh
```diff
@@ -0,0 +1,1148 @@
...
+        const Evaluation resistanceFactor = 1.0 + (residualResistanceFactor - 1.0) * polymerAdsorption_ / maxAdsorbtion;
+
+        // compute effective viscosities
+        auto& fs = asImp_().fluidState_;
+        const Evaluation& muWater = fs.viscosity(waterPhaseIdx);
+        const auto& viscosityMultiplier = PolymerModule::plyviscViscosityMultiplier(elemCtx, dofIdx, timeIdx);
```
andlaus: probably `plyviscViscosityMultiplier()` should be renamed to `plyviscViscosityMultiplierTable()` (or something) because to me, the former makes the impression that it returns a Scalar or an Evaluation.

ewoms/models/blackoil/blackoilpolymermodules.hh
```diff
@@ -0,0 +1,1148 @@
...
+        Evaluation viscosityPolymerEffective = pow(viscosityMixture, plymixparTL) * pow(viscosityPolymer, 1.0 - plymixparTL);
+        Evaluation viscosityWaterEffective = pow(viscosityMixture, plymixparTL) * pow(muWater, 1.0 - plymixparTL);
+
+        Evaluation cbar = polymerConcentration_ / cmax;
+        // waterViscosity / effectiveWaterViscosity
+        waterViscosityCorrection_ = muWater * ( (1.0 - cbar) / viscosityWaterEffective + cbar / viscosityPolymerEffective );
+
+        // adjust water mobility
+        Evaluation& mobWater = asImp_().mobility_[waterPhaseIdx];
+        mobWater *= waterViscosityCorrection_ / resistanceFactor;
```
andlaus: I think that it is easier to read if these two lines are merged, i.e.

```
asImp_().mobility_[waterPhaseIdx] *= waterViscosityCorrection_ / resistanceFactor;
```

ewoms/models/blackoil/blackoilpolymermodules.hh
```diff
@@ -0,0 +1,1148 @@
...
+    {
+        // update rock properties
+        polymerAlivePoreSpace_ = 1.0 - PolymerModule::plyrockDeadPoreVolume(elemCtx, scvIdx, timeIdx);
+        polymerRockDensity_ = PolymerModule::plyrockRockDensityFactor(elemCtx, scvIdx, timeIdx);
+    }
+
+    const Evaluation& polymerConcentration() const
+    { return polymerConcentration_; }
+
+    const Scalar& polymerAlivePoreSpace() const
```
andlaus: if it is called `PolymerModule::plyrockDeadPoreVolume()` in the polymer module, it should be `polymerAlivePoreVolume()` here, shouldn't it? also, it does not look like this returns a pore volume but  the *fraction* of the pore volume.

ewoms/models/blackoil/blackoilpolymermodules.hh
```diff
@@ -0,0 +1,1148 @@
...
+    {
+        unsigned pvtnumRegionIdx = elemCtx.problem().pvtRegionIndex(elemCtx, scvIdx, timeIdx);
+        return plyviscViscosityMultiplier_[pvtnumRegionIdx];
+    }
+
+    static const Scalar& plymaxCmax(const ElementContext& elemCtx,
+                                    unsigned scvIdx,
+                                    unsigned timeIdx)
+    {
+        unsigned polymerMixRegionIdx = 0; //elemCtx.problem().polymerMixRegionIndex(elemCtx, scvIdx, timeIdx);
```
andlaus: this and the method below don't seem to be right. if they are, the parameters should be annotated by OPM_UNUSED.

ewoms/models/blackoil/blackoilpolymermodules.hh
```diff
@@ -0,0 +1,1148 @@
...
+                plyviscViscosityMultiplier_[pvtRegionIdx].setXYContainers(c, visc);
+            }
+        } else {
+            OPM_THROW(std::runtime_error, "PLYVISC must be specified in POLYMER runs\n");
+        }
+
+        // initialize the objects which deal with the PLYMAX keyword
+        const auto& plymaxTables = tableManager.getPlymaxTables();
+        unsigned numMixRegions = plymaxTables.size();
+        assert(numMixRegions == 1);// TODO: supports PLMIXNUM
```
totto82: It is missing in the parser.

ewoms/models/blackoil/blackoilpolymermodules.hh
totto82: This is correct

ewoms/models/blackoil/blackoilpolymermodules.hh
totto82: It should be log(Z)

ewoms/models/blackoil/blackoilpolymermodules.hh
```diff
@@ -0,0 +1,1158 @@
...
+            return u + logShearEffectMultiplier.eval(u, true) - v0AbsLog;
+        };
+        // and its derivative
+        auto dF = [&logShearEffectMultiplier](const auto& u) {
+            return 1 + logShearEffectMultiplier.evalDerivative(u, true);
+        };
+
+        // Solve F = 0 using Newton
+        // Use log(v0) as inital value for u
+        auto u = v0AbsLog;
```
GitPaean: Any reason to use the Newton's method to replace the existing line-intersection method for the PLYSHLOG? I thought the line-intersection method is probably more robust (considering a weird table contains both shear-thinning and shear-thickening) since no iteration is required and probably more efficient (while never done any comparison). 
I guess the reason to solve the shear effect in log space is to use the line-intersection method, otherwise it can be done with the same way with PLYSHEAR.

totto82: You need to do in in the log space to get the same interpolation. I haven't done any comparison on robustness and performance for this part of the code. But flow_ebos_polymer is approx twice as fast as flow_polymer for the test cases in opm-data. And I don't see any issues with convergence for any of these cases. Please report it if you have issues with other non-open tests.

GitPaean: Detailed investigation probably needs some efforts and I was wondering there are some reasons.  PLYSHEAR solves the shear effects iteratively and PLYSHLOG maps the multiplication relation to a addition relation to make the problem a line intersection problem.  Solving a line intersection problem with Newton's method is a little wasteful in my opinion.  It will be okay as long as it works.

ewoms/models/blackoil/blackoilpolymermodules.hh
andlaus: I see. hopefully it is not missing too much longer: OPM/opm-parser#1093

andlaus: Actually, it probably is already supported. the corresponding grid property is MISCNUM if I see this correctly.

totto82: MISCNUM is for the solvent model. For polymer PLMIXNUM should be used. 
OK. Lets get PLMIXNUM into the parser and remove this assert.

ewoms/models/blackoil/blackoilpolymermodules.hh
GitPaean: BTW: which case you used to get the twice speedup of flow_ebos_polymer, which is much bigger than the speedup than the other blackoil cases. although the addition of the polymer equation can make the speedup factor different. 

I just make a test with the `3D_plyshlog/case-shear-thickening`. The running time is as follows,
flow_ebos_polymer
```
Total time (seconds):         48.0362
Solver time (seconds):        46.7685
 Assembly time (seconds):     35.0288 (Failed: 0; 0%)
 Linear solve time (seconds): 10.6237 (Failed: 0; 0%)
 Update time (seconds):       0.918453 (Failed: 0; 0%)
 Output write time (seconds): 0.847553
Overall Well Iterations:      153 (Failed: 0; 0%)
Overall Linearizations:       575 (Failed: 0; 0%)
Overall Newton Iterations:    390 (Failed: 0; 0%)
Overall Linear Iterations:    5703 (Failed: 0; 0%)
```
flow_polymer
```
Total time (seconds):         60.1866
Solver time (seconds):        59.5996
 Assembly time (seconds):     37.3703
 Linear solve time (seconds): 20.8016
 Update time (seconds):       1.0359
 Output write time (seconds): 0.0654342
Overall Well Iterations:      0
Overall Linearizations:       705
Overall Newton Iterations:    520
Overall Linear Iterations:    6796
```

totto82: same case on my machine 
flow_ebos_polymer
================    End of simulation     ===============

Total time (seconds):         18.2887
Solver time (seconds):        17.6745
 Assembly time (seconds):     12.9285 (Failed: 0; 0%)
 Linear solve time (seconds): 4.23329 (Failed: 0; 0%)
 Update time (seconds):       0.400164 (Failed: 0; 0%)
 Output write time (seconds): 0.393959
Overall Well Iterations:      153 (Failed: 0; 0%)
Overall Linearizations:       575 (Failed: 0; 0%)
Overall Newton Iterations:    390 (Failed: 0; 0%)
Overall Linear Iterations:    5691 (Failed: 0; 0%)

flow_polymer

================    End of simulation     ===============

Total time (seconds):         29.3902
Solver time (seconds):        29.063
 Assembly time (seconds):     19.8528
 Linear solve time (seconds): 8.43879
 Update time (seconds):       0.553843
 Output write time (seconds): 0.0354438
Overall Well Iterations:      0
Overall Linearizations:       704
Overall Newton Iterations:    519
Overall Linear Iterations:    6754

ewoms/models/blackoil/blackoilintensivequantities.hh
```diff
@@ -172,7 +176,10 @@ public:
         MaterialLaw::relativePermeabilities(mobility_, materialParams, fluidState_);
         Opm::Valgrind::CheckDefined(mobility_);
 
+        // update the Saturation functions for the blackoil extensions.
         asImp_().solventPostSatFuncUpdate_(elemCtx, dofIdx, timeIdx);
+        asImp_().polymerSatFuncUpdate_(elemCtx, dofIdx, timeIdx);
```
andlaus: almost: I meant to remove the *satFuncUpdate methods entirely and move their contents to `polymerPropertiesUpdate_()`

ewoms/models/blackoil/blackoilpolymermodules.hh
```diff
@@ -365,10 +367,10 @@ public:
     static void registerParameters()
     {
         if (!enablePolymer)
-            // polymers have been disabled at compile time
```
andlaus: this change and the equivalent one in the next method seems to be a minor rebase screwup

ewoms/models/blackoil/blackoilpolymermodules.hh
```diff
@@ -713,8 +715,8 @@ public:
             shearEffectMultiplier[i] = (1.0 + (viscosityMultiplier - 1.0)*shearEffectRefMultiplier[i]) / viscosityMultiplier;
             shearEffectMultiplier[i] = Opm::log(shearEffectMultiplier[i]);
         }
-        // store the log velocity and log multipliers in a table for easy look up and
-        // linear interpolation in the log space.
+        // store the logarithm velocity and logarithm multipliers in a table for easy look up and
```
andlaus: s/logarithm velocity/logarithm of the velocity/

ewoms/models/blackoil/blackoilpolymermodules.hh
```diff
@@ -713,8 +715,8 @@ public:
...
+        // linear interpolation in the logarithm space.
```
andlaus: s/logarithm space/logarithmic space/

andlaus:
except for the tiny nits which I've commented about, I'm happy with this PR now. That said I don't consider myself to be an expert of the polymer extension:

@GitPaean If you're also happy with this, please press green.

ewoms/models/blackoil/blackoilpolymermodules.hh
totto82: For the 2D case: 
flow_ebos_polymer
```
Total time (seconds):         0.40405
Solver time (seconds):        0.329839
 Assembly time (seconds):     0.206329 (Failed: 0.0149605; 7.25079%)
 Linear solve time (seconds): 0.071355 (Failed: 0.00975039; 13.6646%)
 Update time (seconds):       0.0301974 (Failed: 0.00249778; 8.27151%)
 Output write time (seconds): 0.0396279
Overall Well Iterations:      372 (Failed: 0; 0%)
Overall Linearizations:       858 (Failed: 66; 7.69231%)
Overall Newton Iterations:    620 (Failed: 66; 10.6452%)
Overall Linear Iterations:    3502 (Failed: 576; 16.4477%)
```

flow_polymer:

```


Total time (seconds):         2.40668
Solver time (seconds):        2.34383
 Assembly time (seconds):     1.9059
 Linear solve time (seconds): 0.335955
 Update time (seconds):       0.0707482
 Output write time (seconds): 0.0172451
Overall Well Iterations:      0
Overall Linearizations:       1135
Overall Newton Iterations:    903
Overall Linear Iterations:    3838


```

andlaus:
jenkins build this please

ewoms/models/blackoil/blackoilpolymermodules.hh
GitPaean: Thanks.  I have been almost using debug mode for development. And also, it looks some compiler or hardware related favor flow_ebos more than the flow_legacy.  The test result from the 3D case probably represent better the the efficiency improvement for practical use. And the size of SPE9 will be even better in evaluating the performance.

totto82: @GitPaean Performance hasn't been the main concern in this PR. I believe I could speed up the shear factor calculations a bit, but I also don't think that is very critical at the moment. For the 3D case without shear factor the speed up factor is 2. for the shear factor tests it is closer to 1.5. Creating a new table for each call to computeShearFactor() is the main reason for the slow down.

GitPaean: > For the 3D case without shear factor the speed up factor is 2. for the shear factor tests it is closer to 1.5.
I understand the performance of development related to polymer functionality has not been a main concern. In my honest opinion, replacing old functionality needs justifications or specific reasons.  If the comparison of performance of the simulation with/without shear factors show some possible performance degeneration due to the shear calculation, it might not be a good idea to replace.  And also the results also show some degeneration of agreement to ECL. Possibly because solving the line intersection problem in an iterative way results in less accurate  result compared with solving the line intersection problem directly.
The following is some comparison of the case compared with the ECL reference.
flow_polymer
https://www.dropbox.com/s/w1hsl7y698zrfx4/flow_polymer.pdf?dl=0
flow_ebos_polymer
https://www.dropbox.com/s/jx2vldx830rce53/flow_ebos_polymer.pdf?dl=0 

But it is not easy to tell, it can be because the blackoil part cause the comparison degeneration.

totto82: I profiled the code and speeded up a bit. Now it is almost factor 2 (1.8) also for the `case-shear-thickening`

GitPaean:
Please refer to the PR OPM/opm-simulators#972 for the non-one mixing parameters.

totto82:
@GitPaean Thanks for the tip. This is now implemented the same way in ebos with good match also when omega != 1

andlaus:
jenkins build this please

GitPaean:
Thanks for the bigger efforts in improving the results. It will be merged after OPM/opm-core#1169 gets in.

atgeirr:
Thanks for the code, testing and overall effort!

ewoms/models/blackoil/blackoilpolymermodules.hh
```diff
@@ -0,0 +1,1146 @@
...
+                    *up.fluidState().invB(waterPhaseIdx)
+                    *up.polymerViscosityCorrection()
+                    /extQuants.polymerShearFactor()
+                    *up.polymerConcentration();
+
+            // modify water
+            flux[contiWaterEqIdx] /=
+                    extQuants.waterShearFactor();
+        } else {
+            flux[contiPolymerEqIdx] =
```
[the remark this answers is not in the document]
GitPaean: I think it was a silly question since no derivatives should be dropped for sure.   I guess the derivatives might be handled when handling the neighbor cells themselves. I do not understand the details though.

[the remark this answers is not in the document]
andlaus: I'm not sure if this is still relevant for you, but this part is the standard trick when using dense AD: all derivatives are always in respect to the primary variables of the *interior* element (NB, that in newer versions of the code, this has been generalized to *focused* entity). The derivatives for all quantities that do *not* depend on the primary variables of the interior elements must thus be thrown away, IOW if the upstream entity is not the interior one, we need to use scalars instead of evaluations.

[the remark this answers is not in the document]
GitPaean: > dense AD: all derivatives are always in respect to the primary variables of the interior element (NB, that in newer versions of the code, this has been generalized to focused entity). 

I understand this part. This concept is well used in the well modeling code to assemble the well matrix. In the well code, we did not throw away the derivatives from the neighbors, what we did is to put the derivatives in the corresponding location in the matrix.  I think I am still lack of the general understanding of the details of the assembling part of ewoms/ebos. Anyway, I will figure out more with time.

PR #189, opened 2017-06-13
ebos: reduce direct deck usage

andlaus:
this reduces the number of direct deck accesses in `EclProblem` in favour of `EclipseState`-grid properties. as far as i can see, directly using the deck object cannot fully be avoided because `EclipseState` does not seem to provide internalized versions of all these keywords yet.

PR #196, opened 2017-06-16
ebos: do not mistake the pressure for the gas saturation anymore

andlaus:
this is a copy-and-pasto that was introduced with #189.


andlaus:
since this is an obvious bug, there's no need to annoy jenkins with this IMO. merging.

atgeirr:
> since this is an obvious bug, there's no need to annoy jenkins with this IMO. merging.

I disagree, since the fix might be flawed for silly reasons. I agree the bug looks obvious...

PR #198, opened 2017-06-19
Using std::time_t from TimeMap

joakim-hove:
Downstream of: https://github.com/OPM/opm-parser/pull/1092


andlaus:
looks good. merge at will.

PR #199, opened 2017-06-19
fix some field names of the VTK output module for the solvent blackoil model extension

andlaus:
this is kind of trivial, but if you want paraview not to be confused about the files produced with this it is critical.


andlaus:

PR #204, opened 2017-06-27
Make BlackoilModel work for two-phases with hardcoded numPhases.

dr-robertk:
This PR fixes several issues with the BlackoilModel when the number of phases is hardcoded to 2 for efficiency of two-phase cases.

PR #205, opened 2017-06-29
Make it possible to formulate the blackoil eq in terms of surfacevolumes

totto82:
A Tag that enables surface volume scaling is added to the property
system.
NEW_PROP_TAG(EnableSurfaceVolumeScaling);
If enabled the blackoil equations will be formulated in terms of surface
volumes instead of mass/volume.

### DIFF
--- a/ewoms/models/blackoil/blackoillocalresidual.hh
+++ b/ewoms/models/blackoil/blackoillocalresidual.hh
@@ -69,6 +69,8 @@ class BlackOilLocalResidual : public GET_PROP_TYPE(TypeTag, DiscLocalResidual)
     // if compositionSwitchIdx is negative then this feature is disabled in Indices
     static const bool compositionSwitchEnabled = (compositionSwitchIdx >= 0 );
 
+    static constexpr bool blackoilConserveSurfaceVolume = GET_PROP_VALUE(TypeTag, BlackoilConserveSurfaceVolume);
+
     typedef Opm::MathToolbox<Evaluation> Toolbox;
     typedef BlackOilSolventModule<TypeTag> SolventModule;
     typedef BlackOilPolymerModule<TypeTag> PolymerModule;
@@ -118,15 +120,17 @@ public:
         }
 
         // convert surface volumes to component masses
-        unsigned pvtRegionIdx = intQuants.pvtRegionIndex();
-        storage[conti0EqIdx + waterCompIdx] *=
-            FluidSystem::referenceDensity(waterPhaseIdx, pvtRegionIdx);
-        if (compositionSwitchEnabled)
-            storage[conti0EqIdx + gasCompIdx] *=
-                FluidSystem::referenceDensity(gasPhaseIdx, pvtRegionIdx);
-        storage[conti0EqIdx + oilCompIdx] *=
-            FluidSystem::referenceDensity(oilPhaseIdx, pvtRegionIdx);
+        if (!blackoilConserveSurfaceVolume) {
+            unsigned pvtRegionIdx = intQuants.pvtRegionIndex();
+            storage[conti0EqIdx + waterCompIdx] *=
+                    FluidSystem::referenceDensity(waterPhaseIdx, pvtRegionIdx);
+            if (compositionSwitchEnabled)
+                storage[conti0EqIdx + gasCompIdx] *=
+                        FluidSystem::referenceDensity(gasPhaseIdx, pvtRegionIdx);
 
+            storage[conti0EqIdx + oilCompIdx] *=
+                    FluidSystem::referenceDensity(oilPhaseIdx, pvtRegionIdx);
+        }
         // deal with solvents (if present)
         SolventModule::addStorage(storage, intQuants);
 
@@ -185,6 +189,17 @@ public:
                 evalPhaseFluxes_<Scalar>(flux, phaseIdx, extQuants, up);
         }
 
+        if (!blackoilConserveSurfaceVolume) {
+           for (unsigned phaseIdx = 0; phaseIdx < numPhases; ++ phaseIdx) {
+               unsigned compIdx = FluidSystem::solventComponentIndex(phaseIdx);
+               unsigned upIdx = static_cast<unsigned>(extQuants.upstreamIndex(phaseIdx));
+               const IntensiveQuantities& up = elemCtx.intensiveQuantities(upIdx, timeIdx);
+               unsigned pvtRegionIdx = up.pvtRegionIndex();
+               Scalar refDensity = FluidSystem::referenceDensity(phaseIdx, pvtRegionIdx);
+               flux[conti0EqIdx + compIdx] *= refDensity;
+           }
+        }
+
         // deal with solvents (if present)
         SolventModule::computeFlux(flux, elemCtx, scvfIdx, timeIdx);
 
@@ -212,31 +227,29 @@ protected:
                           const IntensiveQuantities& up) const
     {
         unsigned compIdx = FluidSystem::solventComponentIndex(phaseIdx);
-        unsigned pvtRegionIdx = up.pvtRegionIndex();
         const auto& fs = up.fluidState();
 
         Evaluation surfaceVolumeFlux =
-            Toolbox::template decay<UpEval>(fs.invB(phaseIdx))
-            * extQuants.volumeFlux(phaseIdx);
+                Toolbox::template decay<UpEval>(fs.invB(phaseIdx))
+                * extQuants.volumeFlux(phaseIdx);
 
         flux[conti0EqIdx + compIdx] +=
-            surfaceVolumeFlux *
-            FluidSystem::referenceDensity(phaseIdx, pvtRegionIdx);
+                surfaceVolumeFlux;
 
         // dissolved gas (in the oil phase).
         if (phaseIdx == oilPhaseIdx && FluidSystem::enableDissolvedGas()) {
             flux[conti0EqIdx + gasCompIdx] +=
-                FluidSystem::referenceDensity(gasPhaseIdx, pvtRegionIdx)
-                * Toolbox::template decay<UpEval>(fs.Rs())
-                * surfaceVolumeFlux;
+                    Toolbox::template decay<UpEval>(fs.Rs())
+                    * surfaceVolumeFlux;
+
         }
 
         // vaporized oil (in the gas phase).
         if (phaseIdx == gasPhaseIdx && FluidSystem::enableVaporizedOil()) {
             flux[conti0EqIdx + oilCompIdx] +=
-                FluidSystem::referenceDensity(oilPhaseIdx, pvtRegionIdx)
-                * Toolbox::template decay<UpEval>(fs.Rv())
-                * surfaceVolumeFlux;
+                    Toolbox::template decay<UpEval>(fs.Rv())
+                    * surfaceVolumeFlux;
+
         }
     }
 
--- a/ewoms/disc/common/fvbasediscretization.hh
+++ b/ewoms/disc/common/fvbasediscretization.hh
@@ -250,6 +250,10 @@ SET_INT_PROP(FvBaseDiscretization, TimeDiscHistorySize, 2);
 //! Most models use extensive quantities for their storage term (so far, only the Stokes
 //! model does), so we disable this by default.
 SET_BOOL_PROP(FvBaseDiscretization, ExtensiveStorageTerm, false);
+
+// use volumetric residuals is default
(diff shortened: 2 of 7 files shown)


totto82:
The purpose of this PR is to remove convertResults() in BlackoilModelEbos.hpp. I have not tested it in combination with ebos. We may need to scale the convergence criteria if surface volumes is used instead of mass / volume.

andlaus:
in principle a way to chose whether to formulate the residual in terms of specific quantities or extensive ones is a good thing.

as far as I can see, there should be no fundamental problems with conserving the total mass when it comes to the solvers: the linear solvers only look at the *reduction* of the residual anyway and -- depending on element size -- the tolerance of the non-linear solver sometimes larger and sometimes smaller than for volumetric residuals. I have not spend much thought on which possibility is the better one, though. 

the implementation of this PR needs to be slightly improved, though:

- the proposed patch seems to be black-oil specific, yet it modifies the *very* generic part of the finite volume local residuals. the property should thus be split into two: one property that determines whether the local residual is divided by the cell volume or not (`UseVolumetricResidual`?) and a property that determines whether mass or surface volume is conserved by the blackoil model (`BlackOilConserveMass`?).
- problems should not need to set default values for these properties, i.e. a value for them should be set in `fvbasediscretization.hh` and `blackoilmodel.hh`.
- on this branch, does `make test-suite` work on your machine?

totto82:
Thanks for the comments. I did what you suggested. The tests now compiles and doesn't fail.

totto82:
I also tested a patch very I re-scaled the well equations and the residual in getConvergence() in flow_ebos instead. But the convergence was slightly worse and the code more messy.

ewoms/models/blackoil/blackoillocalresidual.hh
```diff
@@ -65,6 +65,8 @@ class BlackOilLocalResidual : public GET_PROP_TYPE(TypeTag, DiscLocalResidual)
     enum { oilCompIdx = FluidSystem::oilCompIdx };
     enum { waterCompIdx = FluidSystem::waterCompIdx };
 
+    static constexpr bool enableSurfaceVolumeScaling = GET_PROP_VALUE(TypeTag, EnableSurfaceVolumeScaling);
```
andlaus: I would prefer this property to be prefixed by `BlackOil`. If you think that this is a bad idea, I don't have a big problem to accept it as-is.

ewoms/models/blackoil/blackoillocalresidual.hh
```diff
@@ -214,24 +218,41 @@ protected:
...
+                    FluidSystem::referenceDensity(phaseIdx, pvtRegionIdx);
+        }
 
         // dissolved gas (in the oil phase).
         if (phaseIdx == oilPhaseIdx && FluidSystem::enableDissolvedGas()) {
-            flux[conti0EqIdx + gasCompIdx] +=
-                FluidSystem::referenceDensity(gasPhaseIdx, pvtRegionIdx)
-                * Toolbox::template decay<UpEval>(fs.Rs())
-                * surfaceVolumeFlux;
+            if (enableSurfaceVolumeScaling) {
```
andlaus: it's probably better to introduce a variable for this, i.e.

```c++
Scalar factor = 1.0;
if (enableSurfaceVolumeScaling)
  factor = FluidSystem::referenceDensity(gasPhaseIdx, pvtRegionIdx);

flux[conti0EqIdx + gasCompIdx] += factor*Opm::decay<UpEval>(fs.Rs())*surfaceVolumeFlux;
```

or -- even better -- to compute everything in terms of surface volume do the conversion to mass at the end of the `computeFlux()` function. this would also make it more consistent with `computeStorage()`.

totto82: Clean up on its way

ewoms/models/blackoil/blackoillocalresidual.hh
totto82: I am not sure if I understand what you suggest. It is probably a good idea. Maybe you could fix this in another PR.

andlaus: simple: I propose to rename the `EnableSurfaceVolumeScaling` property to `BlackOilConserveSurfaceVolume` (or similar). IMO, that would make it clear that this property is specific to the black oil model and it would also be more explicit on what it does.

totto82: Give me a second or two and I will update the PR

totto82: Now its renamed as you suggested. I agree that its makes it more clear.

andlaus:
great! as far as I'm concerned we're done now. let's ask Jenkins about its opinion, just to be sure:

andlaus:

andlaus:
huh? jenkins fails for SPE1...

andlaus:
Okay, I've found the reason for the deviations that cause the SPE1 test to fail: it is because after this patch the conversion from surface volume to mass is done in one go at the end of `computeFlux()` while this is currently done individually for the terms of sum. this is what I would call numeric noise and is IMO another indication that the tolerances for the unit tests are currently too tight.

alfbr:
Assuming that your analysis is correct, I agree. Let's have look at it when people start returning from vacation.

andlaus:
> Assuming that your analysis is correct, I agree. Let's have look at it when people start returning from vacation.

see #208.

andlaus:
okay, I had a closer look what exactly could cause this test failure:

- on the latest master -- i.e. even without this PR -- I was not able to make all tests pass on my machine regardless of the compiler flags used: SPE3CASE1 failed for both flow_ebos and flow_legacy
- nevertheless, the flow_ebos test for SPE1CASE2 failed with this PR but succeeded on master, so I used that one to drill deeper. be aware that this analysis is specific to my machine, so on the jenkins rig the situation may be different
- as expected, the initial FIP numbers are exactly identical for both branches
- the residuals for the reservoir equations differ slightly, but noticeably, even for the iterations of the first report step. as a consequence the FIP numbers -- and thus probably also the total production numbers -- for both branches drift apart over time
- for report step 43, i.e. after 1307 days, the version of the master branch takes one iteration more than the version of this PR. I suppose that this is the main reason for the differences in the final results.
- in the end, the FIP numbers indicate a difference of 129 sm^3 for oil and 1128 sm^3 for gas. (maybe the units are SBBL and MSCF, but this does not really matter IMO.) the final volume of water in place is identical for both runs

for me, the take away of this excercise is that even tiny differences in the linearization proceedure can cause relatively large changes of the final solution, probably because the linear and non-linear solvers tend to amplify them, i.e., this is usually called the "butterfly effect".

@akva2, @totto82, @alfbr: opinions?

alfbr:
Thanks for the analysis, deeply appreciated!

> on the latest master -- i.e. even without this PR -- I was not able to make all tests pass on my machine regardless of the compiler flags used: SPE3CASE1 failed for both flow_ebos and flow_legacy

This sounds like a strong indicator for increasing tolerances for SPE3CASE1. Unfortunately you are not sharing any information as to what the deviations were. i.e., what would the tolerances need to be for the tests to pass?

> for me, the take away of this excercise is that even tiny differences in the linearization proceedure can cause relatively large changes of the final solution, probably because the linear and non-linear solvers tend to amplify them, i.e., this is usually called the "butterfly effect".

Unfortunately, that is not the whole story. As we have discovered over the last months, we have had instabilities in the numerics that causes fluctuations. Actually, bugs. The latest one being merged just now: https://github.com/OPM/opm-simulators/pull/1240 Another that comes to mind is the month old discovery of saturations well outside the [0,1] interval for model 2 realizations. It turned out there were bugs in Dune 2.3.1 which were already fixed in 2.4.1. Actually, for all I know, your discoveries were nullified by today's merge, so please rerun Jenkins to see what the state is now.

Generally speaking though, I do believe the time has come to revisit the tolerances. We have now painstakingly sorted out a number of issues causing numerical instabilities, so fluctuations in the current code should probably be within the tolerances set. Let's allow @akva2 to return from vacation first though.

andlaus:
> This sounds like a strong indicator for increasing tolerances for SPE3CASE1.

these problems seem to be quite spurious. after OPM/opm-data#234 has been merged, they seem to be gone on my machine again.

> Unfortunately you are not sharing any information as to what the deviations were. i.e., what would the tolerances need to be for the tests to pass?

I'm not sure what tolerances should be used, and honestly, I'm also a bit indifferent. I just think that the current tolerances are too tight to be of much use besides the jenkins rig, and that's a bit unfortunate because it prevents people from doing the basic testing before opening their PRs.

I should add that for the current eWoms unit tests, the tolerances are very relaxed -- e.g. the tolerance for saturations is 10% -- because I intended these tests to only detect *qualitative* differences, but not *quantitative* ones. so far this approach seems to have worked well enough, albeit I vaguely remember that some minor bugs slipped through in the past.

> Actually, for all I know, your discoveries were nullified by today's merge, so please rerun Jenkins to see what the state is now.

For this analysis I did not use jenkins at all, so it should still be valid: I did it on my workstation using the latest master versions of all modules from of today ~11:00am. (there was a minor conflict due to #204 when merging this PR which I fixed.)

andlaus:
> This sounds like a strong indicator for increasing tolerances for SPE3CASE1.

these problems seem to be quite spurious. after OPM/opm-data#234 has been merged, they seem to be gone on my machine again.

> Unfortunately you are not sharing any information as to what the deviations were. i.e., what would the tolerances need to be for the tests to pass?

I'm not sure what tolerances should be used, and honestly, I'm also a bit indifferent. I just think that the current tolerances are too tight to be of much use besides the jenkins rig, and that's a bit unfortunate because it prevents people from doing the basic testing before opening their PRs.

I should add that for the current eWoms unit tests, the tolerances are very relaxed -- e.g. the tolerance for saturations is 10% -- because I intended these tests to only detect *qualitative* differences, but not *quantitative* ones. so far this approach seems to have worked well enough, albeit I vaguely remember that some minor bugs slipped through in the past.

> Actually, for all I know, your discoveries were nullified by today's merge, so please rerun Jenkins to see what the state is now.

For this analysis I did not use jenkins at all, so it should still be valid: I did it on my workstation using the latest master versions of all modules from of today ~11:00am. (there was a minor conflict due to #204 when merging this PR which I fixed.)

but I did as you requested: with the current master, the difference in the final FIP numbers w.r.t. this PR is even larger (398 sm^3), but the difference in iteration numbers gets much smaller: the number or non-linear iterations is now identical, the master version needs one iteration more.

alfbr:
> these problems seem to be quite spurious. after OPM/opm-data#234 has been merged, they seem to be gone on my machine again.

Right, so that one was actually caused by a critical bug. Possibly, you were on Dune 2.5 where the bug is already fixed.

> I should add that for the current eWoms unit tests, the tolerances are very relaxed -- e.g. the tolerance for saturations is 10%

I am afraid we both need to find another way of income if 10% deviations in saturations are ignored.

andlaus:
> Possibly, you were on Dune 2.5 where the bug is already fixed.

nope, that should not be the culprit: when I attempt to get the opm-simulators tests to work, I use the Ubuntu 16.04 dune packages (2.4.1) and very mild optimizations (-O2).

alfbr:
> nope, that should not be the culprit: when I attempt to get the opm-simulators tests to work, I use the Ubuntu 16.04 dune packages (2.4.1) and very mild optimizations (-O2).

Thanks for the info, I am afraid Jenkins is still on 2.3.1. Care to rerun this one on Jenkins now?

andlaus:
> I am afraid we both need to find another way of income if 10% deviations in saturations are ignored.

Then I'm pretty sure that we should think about changing our jobs already: For example, try to globally refine all cells of Norne once or twice and look at the results. I bet that differences in saturations larger than 10% compared to the unrefined results are the norm, not the exception. ;)

BTW: that's probably a good task for a summer intern or a bachelor thesis.

> Thanks for the info, I am afraid Jenkins is still on 2.3.1. Care to rerun this one on Jenkins now?

the conflicts need to be resolved first. Since this PR is on @totto82's repository, I cannot do this without closing it and opening a new one. If you consider this issue to be pressing enough, I'll do. (locally, SPE3CASE1 and SPE1CASE2 still fail for flow_ebos with this but not with master, so I don't expect Jenkins to disgress.)

alfbr:
> the conflicts need to be resolved first. Since this PR is on @totto82's repository, I cannot do this without closing it and opening a new one. If you consider this issue to be pressing enough, I'll do.

Then I suggest we just wait for a few days. Nothing pressing on my side for this one.

totto82:

PR #207, opened 2017-06-30
black oil: fix some invalid memory accesses in the native two-phase case

andlaus:
valgrind did not catch these bugs, but they led to very weird test failures. the libstdc++ debug mode came to rescue here.

the second patch fixes some very minor warnings which are produced by the newish blackoil polymer module in some cases.


andlaus:

andlaus:
no objections from either buttler. in it goes!

PR #208, opened 2017-07-01
Blackoil conserve surfvol poc

andlaus:
this PR is identical to #205 except for the fact that the multiplication of the volume fluxes with the reference densities of the phases is done for the individual terms again. this PR is *not* intended to be merged, but ought to show why the SPE1 test fails for #205.


andlaus:

andlaus:
closing this. the only point of this PR was to proof that the jenkins failures of #205 are caused by a different ordering of the arithmetic operations.

PR #209, opened 2017-07-13
remove support for dune < 2.4

andlaus:
this allows quite a substantial amount of code decluttering, mainly because Dune 2.4 renamed mapper.map() to mapper.index().

### DIFF
--- a/dune.module
+++ b/dune.module
@@ -9,5 +9,5 @@ Description: The eWoms simulation framework for flow  and transport in porous me
 Url: http://opm-project.org
 MaintainerName: The Open Porous Media Project
 Maintainer: opm@opm-project.org
-Depends: dune-grid (>= 2.3) dune-istl  (>= 2.3) opm-common opm-material
-Suggests: dune-localfunctions (>= 2.3) dune-fem  (>= 2.4) dune-alugrid (>= 2.4) opm-parser opm-grid opm-core
+Depends: dune-grid (>= 2.4) dune-istl  (>= 2.4) opm-common opm-material
+Suggests: dune-localfunctions (>= 2.4) dune-fem  (>= 2.4) dune-alugrid (>= 2.4) opm-parser opm-grid opm-core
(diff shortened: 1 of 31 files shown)


andlaus:

alfbr:
This is a rather invasive change. We are still not able to set up a stable build chain without 2.3.1 on Debian stable. Our own Jenkins server runs the old stable and just shifted from 2.3.1 this week. We are working hard on making that happen for linuxbenchmarking.com too, but unsuccessful so far. Hence, this PR is a bit too early, and we need at least some more weeks before we kill off 2.3.1 support. Preferably we do it after vacation time.

andlaus:
>  Our own Jenkins server runs the old stable and just shifted from 2.3.1 this week.

@akva2 recently announced on the OPM mailing that he successfully upgraded Jenkins to Dune 2.4.1. Unfortunately I cannot provide a link because the archiving infrastructure of the mailing list seems to be broken at the moment.

akva2:
... guess on who's instructions ;)

andlaus:
and, btw, for me as the eWoms maintainer, it is a considerable pain in the b*** to continue maintaining Dune 2.3 compatibility in the code, and it is an even more considerable pain to keep the unit tests passing. that does not even touch on the fact that the performance on dune 2.3 is suboptimal -- at least I've been told so.

tl;dr: we should not encourage people to shoot themselves into their feet.

andlaus:
Okay, I've also adapted the remaining modules; at least the ones which were part of the 2017.04 release:

- OPM/opm-material#238
- OPM/opm-grid#277
- OPM/opm-core#1170
- OPM/opm-simulators#1241
- OPM/opm-upscaling#220

opm-common, opm-parser and opm-output do not use dune and thus don't need to be adapted. travis still needs to be dealt with one way or another. (probably we should think about removing it for everything except opm-parser.)

alfbr:
> for me as the eWoms maintainer, it is a considerable pain in the b*** 

Yes, and that is why we are working very hard to make the transition fast. All I am saying is that we are not there just yet, and as long as eWoms is in the build chain of flow, this is probably something you need to deal with, and accept the need for coordination.

andlaus:
thanks for getting this in!

alfbr:
You are welcome :-)

PR #210, opened 2017-07-19
require dune >= 2.4

akva2:
Required for correct results in opm-simulators.


andlaus:
while I've personally got no objections, others may disagree and they might be on vacation so let's wait a while before we merge.

just to verify that I understand this correctly: To me it looks like this PR is identical to #209 except that the code obsoleted by this change is not immediately removed?

akva2:
yes, sorry, i was in auto mode. we'll use #209

PR #212, opened 2017-07-25
ConditionalStorage: Switch it to pointer semantics

andlaus:
this is IMO quite a bit more intuitive because this class is a layer of indirection in front of some object, i.e., it can be regarded as a smart pointer.


andlaus:

PR #213, opened 2017-07-27
fix up the intensive quantities caching

andlaus:
if the intensive quantities cache was disabled, this caused segmentation faults when an exception was thrown during the simulation.


andlaus:

PR #214, opened 2017-07-28
remove the specialized copy constructors and copy operators from all IQ classes

andlaus:
after OPM/opm-material#242, using memcpy() for this does not yield any performance benefit anymore but it may cause trouble in the future if some models do not use trivially copyable intensive quantities. (this
should probably be avoided for performance reasons, but the proper way to do this is to do this explicitly, i.e., using std::is_trivially_copyable)


andlaus:

PR #215, opened 2017-08-04
Improve the adaptPrimaryVariables code

totto82:
Changes
- water only case is treated separately
- an epsilon can be added to restrict the switching. Useful to avoid
oscillating between primary variable meaning
- call the method recursively since a variable switch may occur more than
once at a time

Testing 
Pass all tests in ewoms 
Runs SPE9 fine with ebos. 
Tested in combination with flow_ebos on SPE9, Norne, Model2-0 and Model2-5
Tested with flow_ebos_solvent

### DIFF
--- a/ewoms/models/blackoil/blackoilintensivequantities.hh
+++ b/ewoms/models/blackoil/blackoilintensivequantities.hh
@@ -123,14 +123,14 @@ public:
         Evaluation Sw = priVars.makeEvaluation(Indices::waterSaturationIdx, timeIdx);
 
         Evaluation Sg = 0.0;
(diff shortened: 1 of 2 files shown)


andlaus:
Nice! Is my mental model that this PR is a prerequisite of OPM/opm-core#1172 and OPM/opm-simulators#1247 but that merging it on its own will not affect `flow_ebos*` in any way correct?

Regarding the code itself: I've briefly looked at the patch and I did not notice any obvious shenanigans. That said, I currently cannot test it in any way, so anyone who's confident enough to press Green is encouraged to do so.

PR #216, opened 2017-08-23
Avoid negative effective saturations in solvent module

totto82:
Motivation: Run field case model with CO2 injection.


totto82:

ewoms/models/blackoil/blackoilsolventmodules.hh
```diff
@@ -1140,14 +1143,14 @@ private:
         const Evaluation muSolventPow = pow(muSolvent, 0.25);
 
         Evaluation muMixOilSolvent = muOil;
-        if ( std::abs(oilSolventEffSat.value()) > cutOff)
```
andlaus: should this not be `oilSolventEffSat > 0.0` or `oilSolventEffSat > cutoff? if the absolute value is considered, `oilSolventEffSat` and thus the base of the exponent can be negative as far as I can see.

This is unrelated, but the code here also won't work if `oilSolventEffSat` is a primitive floating point value. to fix this, change `oilSolventEffSat.value()`  to `Opm::scalarValue(oilSolventEffSat)`.

andlaus:

ewoms/models/blackoil/blackoilsolventmodules.hh
totto82: Isn't that what I did? This whole PR is about making sure the base is non-negative. I will remove the .value().

totto82: I notices that the code is filled with std::abs( foo.value() ). I can make a separate PR fixing this later.

andlaus: oops, yes. seems like I missed the right column. stupid me ;)

andlaus: > I notices that the code is filled with std::abs( foo.value() ). I can make a separate PR fixing this later.

no worries -- currently this code is not used with finite differences anyway.

totto82:
Any reason why this is not merged?

andlaus:
dementia. thanks for the ping.

PR #217, opened 2017-08-23
Allow parameters to be passed to Simulator.

dr-robertk:
This PR enables the ability to create the Opm::Deck and Opm::EclipseState outside of the EbosSimulator. The default behavior is unchanged.


dr-robertk:
If there are no objections then I'll merge this.

PR #219, opened 2017-08-25
Avoid negative oil saturations for water only case

totto82:
set the gas saturation to zero and water saturation to 1.0 for water only case.


totto82:

totto82:
fix 2p case

totto82:

atgeirr:
This has been tested with Norne and model 2 as well, so in it goes!

Issue #218, opened 2017-08-24
Property system in library source files.


dr-robertk:
I notices that the properties cannot be compiled into the library if there are more then two files doing this. The result is linker errors. Bug or feature?

andlaus:
> Bug or feature?

I consider it to be a bug, but you're welcome to see it as a feature. Anyway, #220 should fix it; it would be nice if you could test-drive that PR.

andlaus:
should be fixed in master. if not, please reopen.

PR #220, opened 2017-09-03
make the property system work with multiple compile units

andlaus:
so far, the linker bailed out due to duplicate definitions of variables if multiple compile units used the same type tag. This is problematic if the sources are split into separate compile units and that use the same type tag; in particular, this applies for traditional libraries.

Due to various C++ peculiarities, this patch complicates the internal implementation of the property system quite a bit, but given that the usage of it (as well as the compile time) stay unchanged, I do not
consider this to be a big problem. Note that the introspection code is particularly problematic because it needs static initializers that do not cause the linker to choke in the case of multiple compile units.

Finally, to prevent future regressions, this patch adds a unit test for the lens problem which uses multiple compile units. (This test is called lens_immiscible_ecfv_ad_mcu and basically identical to the existing lens_immiscible_ecfv_ad test and I thus think that it is pretty unimaginative -- improvement proposals are welcome.)

This PR should fix the issue raised in #218. if jenkins is happy I indend to self-merge this next week. (if nobody presses green first, permission to do so hereby granted.)

### DIFF
--- a/ewoms/common/propertysystem.hh
+++ b/ewoms/common/propertysystem.hh
@@ -84,8 +84,10 @@ namespace Properties {
         }                                                               \
         static std::string propertyName() { return #PropTagName; }      \
     };                                                                  \
-    static const int fooPropInfo_ ## EffTypeTagName ## _ ## PropTagName  = \
-        PropertyInfo<TTAG(EffTypeTagName), PTAG(PropTagName)>::init();
+    namespace fooPropInfo_ ## EffTypeTagName {                          \
+    static const int foo_ ## PropTagName  =                             \
+        PropertyInfo<TTAG(EffTypeTagName), PTAG(PropTagName)>::init();  \
+    }
 
 
 //! Internal macro which is only required if the property introspection is enabled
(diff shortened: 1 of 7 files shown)


andlaus:

ewoms/common/propertysystem.hh
```diff
@@ -72,51 +72,47 @@ namespace Properties {
...
             PropertyRegistry::addKey(key);                              \
             return 0;                                                   \
         }                                                               \
-        static std::string propertyName;                                \
-        static int foo;                                                 \
+        static std::string propertyName() { return #PropTagName; }      \
     };                                                                  \
-    std::string PropertyInfo<TTAG(EffTypeTagName), PTAG(PropTagName)>::propertyName; \
-    int PropertyInfo<TTAG(EffTypeTagName), PTAG(PropTagName)>::foo =    \
+    static const int fooPropInfo_ ## EffTypeTagName ## _ ## PropTagName  = \
```
andlaus: note that this won't work if `EffTypeTagName ## _ ## PropTagName` is not unique: e.g., a property `bar_baz` on the type tag `foo` clashes with the property `baz` defined for the type tag `foo_bar`.  this is a quite pathetic corner case, though.

andlaus: ok, I've found a solution for this case and pushed a patch.

andlaus:

andlaus:
seems like there are no objections. Going Green!

PR #221, opened 2017-09-04
Fix to make 2p gasoil and wateroil compile and run

totto82:
- method for mapping from canonicalToActiveComponentIndex() is added
- guards are added to avoid access to non present components.

### DIFF
--- a/ewoms/models/blackoil/blackoillocalresidual.hh
+++ b/ewoms/models/blackoil/blackoillocalresidual.hh
@@ -66,8 +66,8 @@ class BlackOilLocalResidual : public GET_PROP_TYPE(TypeTag, DiscLocalResidual)
     enum { waterCompIdx = FluidSystem::waterCompIdx };
     enum { compositionSwitchIdx = Indices::compositionSwitchIdx };
 
-    // if compositionSwitchIdx is negative then this feature is disabled in Indices
-    static const bool compositionSwitchEnabled = (compositionSwitchIdx >= 0 );
+    static const bool compositionSwitchEnabled = Indices::gasEnabled;
+    static const bool waterEnabled = Indices::waterEnabled;
 
     static constexpr bool blackoilConserveSurfaceVolume = GET_PROP_VALUE(TypeTag, BlackoilConserveSurfaceVolume);
 
@@ -96,24 +96,26 @@ public:
             if (!FluidSystem::phaseIsActive(phaseIdx))
                 continue;
 
-            unsigned compIdx = FluidSystem::solventComponentIndex(phaseIdx);
+            unsigned activeCompIdx = Indices::canonicalToActiveComponentIndex(FluidSystem::solventComponentIndex(phaseIdx));
             LhsEval surfaceVolume =
                 Toolbox::template decay<LhsEval>(fs.saturation(phaseIdx))
                 * Toolbox::template decay<LhsEval>(fs.invB(phaseIdx))
                 * Toolbox::template decay<LhsEval>(intQuants.porosity());
 
-            storage[conti0EqIdx + compIdx] += surfaceVolume;
+            storage[conti0EqIdx + activeCompIdx] += surfaceVolume;
 
             // account for dissolved gas
             if (phaseIdx == oilPhaseIdx && FluidSystem::enableDissolvedGas()) {
-                storage[conti0EqIdx + gasCompIdx] +=
+                unsigned activeGasCompIdx = Indices::canonicalToActiveComponentIndex(gasCompIdx);
+                storage[conti0EqIdx + activeGasCompIdx] +=
                     Toolbox::template decay<LhsEval>(intQuants.fluidState().Rs())
                     * surfaceVolume;
             }
 
             // account for vaporized oil
             if (phaseIdx == gasPhaseIdx && FluidSystem::enableVaporizedOil()) {
-                storage[conti0EqIdx + oilCompIdx] +=
+                unsigned activeOilCompIdx = Indices::canonicalToActiveComponentIndex(oilCompIdx);
+                storage[conti0EqIdx + activeOilCompIdx] +=
                     Toolbox::template decay<LhsEval>(intQuants.fluidState().Rv())
                     * surfaceVolume;
             }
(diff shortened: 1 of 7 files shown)


ewoms/models/blackoil/blackoillocalresidual.hh
```diff
@@ -96,7 +97,7 @@ public:
             if (!FluidSystem::phaseIsActive(phaseIdx))
                 continue;
 
-            unsigned compIdx = FluidSystem::solventComponentIndex(phaseIdx);
+            unsigned compIdx = Indices::canonicalToActiveComponentIndex(FluidSystem::solventComponentIndex(phaseIdx));
```
andlaus: better call this variable `activeCompIdx` (or similar)

ewoms/models/blackoil/blackoillocalresidual.hh
```diff
@@ -106,14 +107,14 @@ public:
 
             // account for dissolved gas
             if (phaseIdx == oilPhaseIdx && FluidSystem::enableDissolvedGas()) {
-                storage[conti0EqIdx + gasCompIdx] +=
+                storage[conti0EqIdx + Indices::canonicalToActiveComponentIndex(gasCompIdx)] +=
```
andlaus: and use the (renamed) `compIdx` variable here. maybe I missed something.

ewoms/models/blackoil/blackoillocalresidual.hh
```diff
@@ -106,14 +107,14 @@ public:
...
                     Toolbox::template decay<LhsEval>(intQuants.fluidState().Rs())
                     * surfaceVolume;
             }
 
             // account for vaporized oil
             if (phaseIdx == gasPhaseIdx && FluidSystem::enableVaporizedOil()) {
-                storage[conti0EqIdx + oilCompIdx] +=
+                storage[conti0EqIdx + Indices::canonicalToActiveComponentIndex(oilCompIdx)] +=
```
andlaus: I'd prefer if some temporary variable for the active component index was used here and below instead of having this very long statement within the bracket operator.

ewoms/models/blackoil/blackoillocalresidual.hh
```diff
@@ -122,13 +123,14 @@ public:
         // convert surface volumes to component masses
         if (!blackoilConserveSurfaceVolume) {
             unsigned pvtRegionIdx = intQuants.pvtRegionIndex();
-            storage[conti0EqIdx + waterCompIdx] *=
+            if (waterEnabled)
+            storage[conti0EqIdx + Indices::canonicalToActiveComponentIndex(waterCompIdx)] *=
```
andlaus: nit: indentation?

ewoms/models/blackoil/blackoillocalresidual.hh
```diff
@@ -191,7 +169,7 @@ public:
 
         if (!blackoilConserveSurfaceVolume) {
            for (unsigned phaseIdx = 0; phaseIdx < numPhases; ++ phaseIdx) {
-               unsigned compIdx = FluidSystem::solventComponentIndex(phaseIdx);
+               unsigned compIdx = Indices::canonicalToActiveComponentIndex(FluidSystem::solventComponentIndex(phaseIdx));
```
andlaus: s/compIdx/activeSolventCompIdx/. maybe the name "solvent" is too misleading here, because it is *not* referring to the component of the black-oil solvent extension but to the dominating hydrocarbon component of a given phase (i.e. the oil component for the oil phase and the gas component for the gas phase). better names highly appreciated.

ewoms/models/blackoil/blackoilmodel.hh
```diff
@@ -211,6 +211,7 @@ class BlackOilModel
 
     // if compositionSwitchIdx is negative then this feature is disabled in Indices
     static const bool compositionSwitchEnabled = (Indices::compositionSwitchIdx >= 0);
+    static const bool waterEnabled = (Indices::waterSaturationIdx >= 0);
```
andlaus: relying on negative indices for this has always seemed to be a kludge to me. possibly this PR is a good opportunity to make the Indices class to export `{water,oil,gas}Enabled` attributes.

ewoms/models/blackoil/blackoiltwophaseindices.hh
```diff
@@ -35,7 +35,7 @@ namespace Ewoms {
  *
  * \brief The primary variable and equation indices for the black-oil model.
  */
-template <unsigned numSolventsV, unsigned numPolymersV, unsigned PVOffset>
+template <unsigned numSolventsV, unsigned numPolymersV, unsigned PVOffset, bool gasoil>
```
andlaus: it may be smarter to pass the (canonical) index of the disabled phase here?

ewoms/models/blackoil/blackoiltwophaseindices.hh
```diff
@@ -64,9 +64,9 @@ struct BlackOilTwoPhaseIndices
      * \brief Index of the switching variable which determines the composition of the
      *        hydrocarbon phases.
      *
-     * \note For Two-Phase models this is disabled.
+     * \note For Two-Phase water oil models this is disabled.
```
andlaus: nit, typo: capitalization (yes, it was there before your PR)

ewoms/models/blackoil/blackoiltwophaseindices.hh
```diff
@@ -90,6 +93,14 @@ struct BlackOilTwoPhaseIndices
...
     static const int contiPolymerEqIdx = contiSolventEqIdx + numPolymers;
     // numSolvents-1 continuity equations follow
+
+    // mapping from canonical to active component index
+    static constexpr int canonicalToActive_[] =
+    {
+        0, // oil comp
+        1, // water comp
+        1 // gas comp
+    };
```
andlaus: this seems to be a bit weird and it won't work for the (future) water-gas case. Most likely, this can be replaced using ifs in the `canonicalToActiveComponentIndex()` method without a performance penalty because the conditions of these ifs can be evaluated at compile time.

totto82:
Thanks for the review.

totto82:
This compiles on my  Ubuntu 16.04 with gcc 5.4, but not on Jenkins with gcc 4.9.2. 

> 
>  In file included from /var/lib/jenkins/workspace/opm-simulators-PR-builder/examples/flow.cpp:42:0:
> /var/lib/jenkins/workspace/opm-simulators-PR-builder/mpi/install/include/ewoms/models/blackoil/blackoiltwophaseindices.hh:90:31: note: 'static constexpr unsigned int Ewoms::BlackOilTwoPhaseIndices<numSolventsV, numPolymersV, PVOffset, disabledCanonicalCompIdx>::canonicalToActiveComponentIndex(unsigned int) [with unsigned int numSolventsV = 0u; unsigned int numPolymersV = 0u; unsigned int PVOffset = 0u; unsigned int disabledCanonicalCompIdx = 2u]' is not usable as a constexpr function because:
>      static constexpr unsigned canonicalToActiveComponentIndex(unsigned compIdx)
> 

Any clue how to fix this?

bska:
> This compiles on my Ubuntu 16.04 with gcc 5.4, but not on Jenkins with gcc 4.9.2.
>
>> ```
>> note: 'static constexpr unsigned int
>> Ewoms::BlackOilTwoPhaseIndices<>::canonicalToActiveComponentIndex(unsigned int)' is not usable as a constexpr function
>> ```

It's because of  OPM/opm-common@4d31c7e.  GCC 4.9 does not support the flag `-std=c++14` (it uses `-std=c++1y` to activate limited support for C++14.)  GCC 5.4 on the other hand **does** support the flag `-std=c++14` meaning that your code is compiled as C++11 on Jenkins and as C++14 on your machine.

The rules for `constexpr` functions were significantly relaxed between C++11 and C+/ +14.  In 11, a `constexpr` function essentially has to look like
```C++
constexpr T f(args)
{ return expression(args); }
```
i.e., its body must consist entirely of a single `return` expression.  In 14, `constexpr` functions have much greater flexibility and can contain control flow, multiple statements, local variables &c.  Additional details at [cppreference.com](http://en.cppreference.com/w/cpp/language/constexpr).

I don't see an immediate way of restructuring the function in question, but you will need to do so if you want to mark it as `constexpr` and have it compile as C++11.

andlaus:
I agree with @bska's analysis and like to add that `assert()` is also not allowed in constexpr methods. (IIRC this even applies to c++-14 and c++-17.) Probably it didn't bite because you only compiled this in release mode where the asserts go away.

regarding a fix: probably it is sufficient to remove the `constexpr` and add a `const` qualifier. While I don't expect this to have any influence on performance, this should be tested nevertheless, though. if there is a performance impact, the asserts can be removed and the ifs can be replaced by `?` operators, or we can go back to your original solution.

PR #222, opened 2017-09-15
Bug fix for pvsratevector indices in setMassRate function

kel85uk:
The setMassRate function accesses the wrong index of the pvsRateVector.


andlaus:
oops, you're completely right, merging.

as far as I can see, this bug only bit if conti0EqIdx != 0. did this surface in the context of a non-isothermal simulation?

kel85uk:
@andlaus : No, not quite. I just noticed that the indices were not offset correctly when I was examining the code to use to add the source term for my salt model.

andlaus:
ok, cool. :) please keep reporting the bugs and oddities you find.

PR #226, opened 2017-10-05
ebos: simplify passing of pre-parsed ECL decks

andlaus:
This gets rid of some special-purpose code in generic places (i.e. the `SimulatorParameter` class) and no special hacks to the property and parameter system are required anymore.

### DIFF
--- a/ebos/eclbasegridmanager.hh
+++ b/ebos/eclbasegridmanager.hh
@@ -94,6 +94,22 @@ public:
     }
 
     /*!
+     * \brief Set the Opm::EclipseState and the Opm::Deck object which ought to be used
+     *        when the grid manager is instantiated.
+     *
+     * This is basically an optimization: In cases where the ECL input deck must be
+     * examined to decide which simulator ought to be used, this avoids having to parse
+     * the input twice. When this method is used, the caller is responsible for lifetime
+     * management of these two objects, i.e., they are not allowed to be deleted as long
+     * as the grid manager object is alive.
+     */
+    static void setExternalDeck(Opm::Deck* deck, Opm::EclipseState* eclState)
+    {
+        externalDeck_ = deck;
+        externalEclState_ = eclState;
+    }
+
+    /*!
      * \brief Create the grid for problem data files which use the ECL file format.
      *
      * This is the file format used by the commercial ECLiPSE simulator. Usually it uses
@@ -131,29 +147,33 @@ public:
         caseName_ = rawCaseName;
         std::transform(caseName_.begin(), caseName_.end(), caseName_.begin(), ::toupper);
 
-        if (myRank == 0)
-            std::cout << "Reading the deck file '" << fileName << "'" << std::endl;
+        if (!externalDeck_) {
+            if (myRank == 0)
+                std::cout << "Reading the deck file '" << fileName << "'" << std::endl;
 
-        if( ! simulator.simulatorParameter().first )
-        {
             Opm::Parser parser;
             typedef std::pair<std::string, Opm::InputError::Action> ParseModePair;
             typedef std::vector<ParseModePair> ParseModePairs;
+
             ParseModePairs tmp;
-            tmp.push_back(ParseModePair(Opm::ParseContext::PARSE_RANDOM_SLASH, Opm::InputError::IGNORE));
-            tmp.push_back(ParseModePair(Opm::ParseContext::PARSE_MISSING_DIMS_KEYWORD, Opm::InputError::WARN));
-            tmp.push_back(ParseModePair(Opm::ParseContext::SUMMARY_UNKNOWN_WELL, Opm::InputError::WARN));
-            tmp.push_back(ParseModePair(Opm::ParseContext::SUMMARY_UNKNOWN_GROUP, Opm::InputError::WARN));
+            tmp.emplace_back(Opm::ParseContext::PARSE_RANDOM_SLASH, Opm::InputError::IGNORE);
+            tmp.emplace_back(Opm::ParseContext::PARSE_MISSING_DIMS_KEYWORD, Opm::InputError::WARN);
+            tmp.emplace_back(Opm::ParseContext::SUMMARY_UNKNOWN_WELL, Opm::InputError::WARN);
+            tmp.emplace_back(Opm::ParseContext::SUMMARY_UNKNOWN_GROUP, Opm::InputError::WARN);
             Opm::ParseContext parseContext(tmp);
 
-            deck_.reset( new Opm::Deck(parser.parseFile(fileName , parseContext)) );
-            eclState_.reset(new Opm::EclipseState(deck(), parseContext));
+            internalDeck_.reset(new Opm::Deck(parser.parseFile(fileName , parseContext)));
+            internalEclState_.reset(new Opm::EclipseState(*internalDeck_, parseContext));
+
+            deck_ = &(*internalDeck_);
+            eclState_ = &(*internalEclState_);
         }
-        else
-        {
-            deck_ = simulator.simulatorParameter().first;
-            assert( simulator.simulatorParameter().second );
-            eclState_ = simulator.simulatorParameter().second;
+        else {
+            assert(externalDeck_);
+            assert(externalEclState_);
+
+            deck_ = externalDeck_;
+            eclState_ = externalEclState_;
         }
 
         asImp_().createGrids_();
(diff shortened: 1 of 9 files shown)


ebos/eclbasegridmanager.hh
```diff
@@ -94,6 +94,22 @@ public:
...
+     * \brief Set the Opm::EclipseState and the Opm::Deck object which ought to be used
+     *        when the grid manager is instantiated.
+     *
+     * This is basically an optimization: In cases where the ECL input deck must be
+     * examined to decide which simulator ought to be used, this avoids having to parse
+     * the input twice. When this method is used, the caller is responsible for lifetime
+     * management of these two objects, i.e., they are not allowed to be deleted as long
+     * as the grid manager object is alive.
+     */
+    static void setExternalDeck(Opm::Deck* deck, Opm::EclipseState* eclState)
```
atgeirr: Why does this have to be static, is it not possible to just use this to set the deck_ and eclState_ members?

andlaus: this is because it must already be known whether to parse the deck file or use a pre-parsed version at the time the object is constructed. (you might argue that an empty object should be constructed, a grid manager object that does not manage a grid is IMO semantically more dubious than a static method!?)

andlaus:
ok, now this PR does not break the opm-simulators master anymore. if jenkins is happy and lacking objections other than static variables being bad in principle, I'll merge it tomorrow.

andlaus:

atgeirr:
> lacking objections other than static variables being bad in principle, I'll merge it tomorrow.

Since you are so eager to merge: are there other urgent changes that depend on this?

andlaus:
> Since you are so eager to merge: are there other urgent changes that depend on this?

nope, as far as I can see this does not break anything in downstream. If the new deprecation warning is supposed to be avoided, OPM/opm-simulators#1287 (or similar) needs to be merged.

andlaus:
(I do not want the SimulatorParameter slip into the release without big warning signs around it because IMO `simulator().simulatorParams().first` is quite a bit more obscure than using `externalDeck_` within the grid manager.)

atgeirr:
> (I do not want the SimulatorParameter slip into the release without big warning signs around it because IMO simulator().simulatorParams().first is quite a bit more obscure than using externalDeck_ within the grid manager.)

I see that, but why not change the SimulatorParameter type to be a struct with Deck and EclipseState (shared pointer) members rather than a pair? I'd think `simulator().simulatorParams().deck` would be very understandable?

andlaus:
This is possible but it is IMO still inferior to using static members, because it is IMO still less explicit than registering externally provided data before instantiation of the simulator and the scheme enforces exactly one type for the `SimulatorData` class for the whole simulation. the latter quickly leads to a maintenance nightmare if the mechanism was used in more than one place (which, granted, it currently is not).

blattms:
So can never have two Simulators in one program using different decks? 
Is e.g. optimizing one deck parameter on a multicore machine (e.g. ech thread runs one simulation with a different deck) an unlikely scenario?

andlaus:
that what you ask for is equally easy to do with the proposed scheme:

```c++
Deck sequentialDeck; // = ...
EclipseState sequentialEclState(sequentialDeck);

GridManager::setExternalDeck(sequentialDeck, sequentialEclState);
Simulator sequentialSim;

// repeat the exercise with a deck optimized for parallel execution
```

in the current scheme the type of SimulatorParamter gets fixed, so if e.g. the black-oil model wants `SimulatorParameters` to be `std::vector<double>` and the `EclProblem` wants it to be `std::pair<int, double>` you will run into trouble. Obviously this example does not make too much sense, but I hope it illustrates the point.

atgeirr:
> the scheme enforces exactly one type for the SimulatorData (*SimulatorParameter* I assume) class for the whole simulation

No, it only weakly enforces type, as for example `simulator().simulatorParams().deck` only requires that deck exists, it would be easy to make a struct supplying the union of all such data assumed present by any involved classes (and if you forgot one, the compile error "SimulatorParams has no member foobar" would tell you).

> if e.g. the black-oil model wants SimulatorParameters to be std::vector<double> and the EclProblem wants it to be std::pair<int, double> you will run into trouble.

If those vectors and pairs were given reasonable names it would be no problem, they would all be passed through the struct. Assuming that SimulatorParameter == a particular class is a no-go, assuming that it has a particular member is ok (as usual for template args).

Also, to your example:
```
Deck sequentialDeck; // = ...
EclipseState sequentialEclState(sequentialDeck);

GridManager::setExternalDeck(sequentialDeck, sequentialEclState);
Simulator sequentialSim;

// repeat the exercise with a deck optimized for parallel execution
```
I assume the final line encapsulates `GridManager::setExternalDeck(theOtherDeck, theOtherEclState)`, right? At that point, would not that "contaminate" the first sim? For example, when the first sim asks for the schedule at a certain timestep, would it not get the one from `theOtherEclState` rather than from `sequentialEclState`?

andlaus:
> No, it only weakly enforces type, as for example `simulator().simulatorParams().deck` only requires that deck exists, it would be easy to make a struct supplying the union of all such data assumed present by any involved classes (and if you forgot one, the compile error "SimulatorParams has no member foobar" would tell you).

yes, but then you need to write a separate class for each type tag and the fact that these weak types must be compatible (i.e., no clashes in the names of the members are allowed) remains. No `std::pair<A, B>` and `std::pair<C, D>` in the same simulation...

> If those vectors and pairs were given reasonable names it would be no problem, they would all be passed through the struct.

I consider the proposed new mechanism with static variables considerably more bullet-proof, explicit and simpler to use. I don't want to stick to the "second best" solution just because it was implemented first.

> I assume the final line encapsulates `GridManager::setExternalDeck(theOtherDeck, theOtherEclState`), right?

no, it is simply a copy-paste-and-rename job: 

```c++
Deck sequentialDeck; // = ...
EclipseState sequentialEclState(sequentialDeck);

GridManager::setExternalDeck(sequentialDeck, sequentialEclState);
Simulator sequentialSim;

Deck parallelDeck; // = ...
EclipseState parallelEclState(parallelDeck);

GridManager::setExternalDeck(parallelDeck, parallelEclState);
Simulator parallelSim;
```

this is the case because after the simulator has been created, the external deck pointers can be modified without changing the state of the simulator. the only thing the calling code needs to ensure is that these objects are not deleted.

atgeirr:
> this is the case because after the simulator has been created, the external deck pointers can be modified without changing the state of the simulator. the only thing the calling code needs to ensure is that these objects are not deleted.

That is confusing behaviour! (Proof: it fooled me!)


> yes, but then you need to write a separate class for each type tag and the fact that these weak types must be compatible (i.e., no clashes in the names of the members are allowed) remains. No std::pair<A, B> and std::pair<C, D> in the same simulation...

I don't understand why you need to write separate classes? By default it can be empty, and only when you want to use a class that requires for example a deck, do you need to provide it. So for those cases when you would like to call setExternalDeck(), you would instead need to pass a struct containing a `deck` member. Seems to me quite equivalent?

> I don't want to stick to the "second best" solution just because it was implemented first.

Neither do I, but we disagree about which is best...

> I'll merge this PR in the afternoon if nobody can come up with a reasonable use case that can be covered with the SimulatorParameter mechanism but not with the new one (assuming there are no other technical objections)...

Can I ask you to please abstain from doing this until we have settled the issue properly?

andlaus:
> I don't understand why you need to write separate classes? By default it can be empty, and only when you want to use a class that requires for example a deck, do you need to provide it.

if these "weak" types are compatible by accident, then they can be reused of course. that's *by accident*. Also, there is no good way to find out which of these parameters are actually necessary and which are not.

> So for those cases when you would like to call setExternalDeck(), you would instead need to pass a struct containing a deck member. Seems to me quite equivalent?

Yes, everything can probably also be done with the `SimulatorParameter` mechanism, given enough determination to do so. That is not the point, though: The reason why I think it should be replaced is that it implies maintaining a second code path in the startup routines that is currently incomplete (e.g. `Ewoms::start()` is unaware), all possible code paths need to be tested for incompatibilities, and -- IMO -- `SimulatorParameters` feels quite awkward in conjunction with the rest of the code base. the only reason in favor of the `SimulatorParameter` mechanism that I can see is that it avoids static variables. To me, this in itself is not a sufficient reason because there already are plenty of static attributes (as well as some global variables related to the property introspection functionality) in the code base.

> Can I ask you to please abstain from doing this until we have settled the issue properly?

Both solutions work and come with a specific set of advantages and disadvantages and I think that there thus is no "proper" resolution of this issue because it is a matter of taste. That said, I can assure you that the static variable route is much more similar to the rest of the eWoms and opm-material code base and that the `SimulatorParameter` mechanism feels like a "pink sheep" to me. also, since nothing seems to get broken by merging this PR, I do not really see a reason to withhold the merge much longer.

tl;dr: let's get it in and continue discussing if this is a topic which is really *that* important! (IMO to many bits have been flipped because of it already.)

dr-robertk:
I have no preference. The SimulatorParameter was introduced to overcome the chicken-egg problem with the Simulator class initialization.

andlaus:
> I have no preference. The SimulatorParameter was introduced to overcome the chicken-egg problem with the Simulator class initialization.

I hope you did not interpret my comments above differently from how I intended them: I'm grateful for the work you did and I'm glad that it was merged. I just would like to get this code a bit refactored to make it feel more integrated before the release.

dr-robertk:
I actually did not follow the discussion closely. I'm sure you'll do the right thing ;-).

andlaus:
ok, cool; let's merge. the pesky deprecation warnings in `flow`, can be avoided by merging OPM/opm-simulators#1287

PR #227, opened 2017-10-11
remove the deprecated SimulatorParameter mechanism

andlaus:
since OPM/opm-simulators#1287 has been merged there are no "in tree" OPM upstreams which use that mechanism anymore.


andlaus:
note: this PR *must not* be merged before OPM/opm-simulators#1299

PR #229, opened 2017-10-17
add trilinos to debian packaging

akva2:
after reorganizations in the build system, this reared its head.


andlaus:
Since this seems pretty simple, I suppose this is merge ready? I merge it right now, but I'm a bit wondering why debian package picks up the zoltan dependency up transitively from opm-grid.

PR #230, opened 2017-10-24
Drsdt work in progress

hnil:
This is to discuss how to incorporate DRSDT.  This depend on branch hnil_work in opm-simulators and opm-materials. It brakes some of the codes in eworms at the moment.


ebos/eclproblem.hh
```diff
@@ -479,16 +479,24 @@ public:
...
 
+
+
         if (updateHysteresis_())
             this->model().invalidateIntensiveQuantitiesCache(/*timeIdx=*/0);
-
-        this->model().updateMaxOilSaturations();
+        // NB update MaxOilSaturation is called here instead of endEpisode to get correct output for the lagged evalauation
+        // used in eclipse
+        this->model().updateMaxOilSaturations();// this should be removed when cellValues is properly functions
```
andlaus: I think it would be better to move the maximum oil saturation storage and the code which you put into `model().updateCellValues()` into `updateHysteresis()` and rename `updateHysteresis()` to e.g. `updateSolutionParameters()`. these arrays should then also be stored in the problem instead of in the model.

ebos/eclproblem.hh
```diff
@@ -565,6 +573,8 @@ public:
         auto& simulator = this->simulator();
         const auto& eclState = simulator.gridManager().eclState();
         int episodeIdx = simulator.episodeIndex();
+        // NB hysteresis variable is not updated  here since the lagged evaluation of this quantities are used
+
```
andlaus: I don't get this comment...

ewoms/models/blackoil/blackoilintensivequantities.hh
```diff
@@ -115,16 +116,21 @@ public:
...
         const auto& priVars = elemCtx.primaryVars(dofIdx, timeIdx);
+        const auto& priVars0 = elemCtx.primaryVars(dofIdx, timeIdx+1);
 
         unsigned globalSpaceIdx = elemCtx.globalSpaceIndex(dofIdx, timeIdx);
         unsigned pvtRegionIdx = priVars.pvtRegionIndex();
         fluidState_.setPvtRegionIndex(pvtRegionIdx);
 
         // extract the water and the gas saturations for convenience
+
+        // time index here is only used to deside if one crate a iday or not timeIdx==1 will
```
andlaus: typo: decide

also, what do you want to say here?

ewoms/models/blackoil/blackoilintensivequantities.hh
```diff
@@ -115,16 +116,21 @@ public:
 
         const auto& problem = elemCtx.problem();
         const auto& priVars = elemCtx.primaryVars(dofIdx, timeIdx);
+        const auto& priVars0 = elemCtx.primaryVars(dofIdx, timeIdx+1);
```
andlaus: probably you don't need the old primary variables, because you can get the previous Rs (and Rv) value from the problem (currently: model)

ewoms/models/blackoil/blackoilintensivequantities.hh
```diff
@@ -196,12 +202,25 @@ public:
...
             if (FluidSystem::enableDissolvedGas()) {
-                const Evaluation& RsSat =
+                //const Evaluation& RsSat =
+                Evaluation RsSat =
                     FluidSystem::saturatedDissolutionFactor(fluidState_,
                                                             oilPhaseIdx,
                                                             pvtRegionIdx,
                                                             SoMax);
-                fluidState_.setRs(RsSat);
+                Evaluation Rs_tmp = RsSat;
```
andlaus:
```
RsSat = Opm::max(Opm::min(RsSat, problem.lastRs() + dRsDt), problem.lastRs() - dRsDt);
```

(same for Rv)

ewoms/models/blackoil/blackoilintensivequantities.hh
```diff
@@ -220,7 +239,11 @@ public:
         else if (priVars.primaryVarsMeaning() == PrimaryVariables::Sw_po_Rs) {
             // if the switching variable is the mole fraction of the gas component in the
```
andlaus: that comment is now misleading: The black-oil model has used Rs instead of the mole fraction of the gas component in the oil phase as a primary variable for a long time. mea culpa!

ewoms/models/blackoil/blackoilintensivequantities.hh
```diff
@@ -220,7 +239,11 @@ public:
...
             // oil phase, we can directly set the composition of the oil phase
-            const auto& Rs = priVars.makeEvaluation(Indices::compositionSwitchIdx, timeIdx);
```
andlaus: you can go back to a const reference as you do not modify this variable here (and as far as I can see, modifying it is not necessary)

ewoms/models/blackoil/blackoilintensivequantities.hh
```diff
@@ -346,6 +369,30 @@ public:
 #endif
     }
 
+    /*! function to update Rs with rate limmitation this is used if keyword DRSDT is pressent in configuration file
+     */
+    Evaluation rateLimmitedDissolvedGasUpdate(const Evaluation& RsE,const Evaluation& So, const ElementContext& elemCtx, unsigned dofIdx, unsigned timeIdx)
```
andlaus: better do this as a part of [`BlackOilNewtonMethod::updatePrimaryVariables_()`](https://github.com/OPM/ewoms/blob/master/ewoms/models/blackoil/blackoilnewtonmethod.hh#L147)

(okay, `flow` does not use that method yet, but this is a reason for a cleanup IMO.)

ewoms/models/blackoil/blackoilmodel.hh
```diff
@@ -216,11 +219,18 @@ class BlackOilModel
...
     BlackOilModel(Simulator& simulator)
         : ParentType(simulator)
     {}
-
+    enum CellValuesName : unsigned {
+        prevSo = 0,
+        prevRs = 1,
+        maxOil = 2,
+        numCellValues,
+    };
```
andlaus: I don't think that this is a good idea. stuff like

```
model.getValues(Model::prevRs);
```

makes the code harder to understand than

```
model.prevRs()
```

because it relies on "magic" constants.

anyway, I think that this change is orthogonal to the DRSDT stuff and should thus go to a separate PR.

ewoms/models/blackoil/blackoilmodel.hh
```diff
@@ -242,6 +252,11 @@ public:
     void finishInit()
     {
         maxOilSaturation_.resize(this->numGridDof(), 0.0);
+        cellValues_.resize(numCellValues);
```
andlaus: the name `cell` should not be used on this level of abstraction: eWoms models can be used for discretization other than element centered FV. For example, there is a test which uses it in conjunction with a vertex-centered FV discretization. (also `cell` is `element` in DUNE-lingo, but I don't think this is very important.)

ewoms/models/blackoil/blackoilmodel.hh
```diff
@@ -384,9 +399,17 @@ public:
...
 
+        if (cellValues_.size()>0){
+            for(unsigned i=0;i<numCellValues;++i){
+                if(cellValues_.size() >0 ){
+                    outstream << cellValues_[dofIdx][i];
+                }
+            }
+        }
         if (maxOilSaturation_.size() > 0)
             outstream << maxOilSaturation_[dofIdx] << " ";
```
andlaus: isn't the maximum oil saturation a part of `cellValues_`?

ewoms/models/blackoil/blackoilmodel.hh
```diff
@@ -525,6 +560,55 @@ public:
...
+    /*!
+     * \brief Update the maximum oil saturation observed during the simulation for all
+     *        elements.
+     *
+     * This method must be called manually by the problem because depending on the exact
+     * simulation it sometimes needs to be called after a time step or before an episode
+     * starts.
+     */
+    //void updateCellValues(ElementContext& elemCtx,& gridManager)
+    void updateCellValues(ElementContext& elemCtx,const GridView& gridView)
```
andlaus: As mentioned above, this should probably go to a renamed `EclProblem::updateHysteresis_();`

ewoms/models/blackoil/blackoilprimaryvariables.hh
```diff
@@ -334,7 +334,7 @@ public:
      *
      * \return true Iff the interpretation of one of the switching variables was changed
      */
-    bool adaptPrimaryVariables(const Problem& problem, unsigned globalDofIdx, Scalar eps = 0.0)
+    bool adaptPrimaryVariables(const Problem& problem, unsigned globalDofIdx,const ElementContext& elemCtx,Scalar eps = 0.0)
```
andlaus: I think you don't need an element context here.

ewoms/models/blackoil/blackoilprimaryvariables.hh
```diff
@@ -381,18 +381,33 @@ public:
                 Scalar po = (*this)[Indices::pressureSwitchIdx];
                 Scalar T = asImp_().temperature_();
                 Scalar SoMax = problem.model().maxOilSaturation(globalDofIdx);
```
andlaus: add

```
Scalar RsPrev = problem.previousRs(globalDofIdx);
```

here!?

ewoms/models/blackoil/blackoilprimaryvariables.hh
```diff
@@ -381,18 +381,33 @@ public:
...
-                Scalar RsSat = FluidSystem::oilPvt().saturatedGasDissolutionFactor(pvtRegionIdx_,
+                //Scalar RsSat= 0.0;
+                Scalar Rs = FluidSystem::oilPvt().saturatedGasDissolutionFactor(pvtRegionIdx_,
                                                                                    T,
                                                                                    po,
                                                                                    So2,
                                                                                    SoMax);
+
+                if(FluidSystem::enableRateLimmitedDissolvedGas()>Opm::FluidSystems::None){
```
andlaus: better do the `Opm::max(Opm::min(Rs, prevRs + maxDelta), prevRs - maxDelta))` dance here.

ewoms/models/blackoil/blackoilprimaryvariables.hh
```diff
@@ -465,20 +480,41 @@ public:
             Scalar SoMax = std::max(So, problem.model().maxOilSaturation(globalDofIdx));
             Scalar RsSat =
                 FluidSystem::oilPvt().saturatedGasDissolutionFactor(pvtRegionIdx_, T, po, So, SoMax);
+            Scalar RsMax = RsSat;
             Scalar Rs = (*this)[Indices::compositionSwitchIdx];
-            if (Rs > RsSat*(1.0 + eps)) {
+
+            // Rs updates is only immited in the case off all
+            if(FluidSystem::enableRateLimmitedDissolvedGas()>Opm::FluidSystems::Free){
```
andlaus: ditto

ebos/eclproblem.hh
hnil: My first thought was the oposite to collect things which change in the model.. I am still trying to think how we can isolate changes since this will simplify adjoint.

ebos/eclproblem.hh
hnil: The point is that we do hysteresis explicitely with lagged evaluationi, which is not obvious if you do not know that this is common .. This means that that "conseptually" the calculation of using the prevois state in the next step should have smax on the prevois time. The proper definition of state is val(t),hyst_var(t-dt).... This is also how i defined it for drst, ie rs(t-dt)

ewoms/models/blackoil/blackoilintensivequantities.hh
hnil: My point is here was that we should separate the use of time index and the initialize as adi, since this is two different thins. Forexample: to calculate residuals without forming matrix, adjoint where other derivatives maybe of interest, to make code more clear...

ewoms/models/blackoil/blackoilintensivequantities.hh
hnil: Everything should be the same for Rv. We just have to find the right structure first. We can make the explicit function names, but I think the underlying structure should be..

andlaus: I was already thinking about DRVDT. But yes, one step at a time...

ewoms/models/blackoil/blackoilintensivequantities.hh
andlaus: you mean making it possible to define `Evaluations` where the derivatives are w.r.t. the primary variables of a time step different than the current one? if you need this, I agree that this feature should be added...

(Deciding what the derivatives mean at a later point is difficult to do with the operator overloading approach.)

ebos/eclproblem.hh
andlaus: currently, hysteresis is not visible at this stage because at the beginning of any given time step, a well defined set of capillary pressure and relperm curves is used, i.e. the hysteresis parameters are basically just constants that are put into these curves.

Also, be aware that the *values* hysteresis curves should be exactly identical for the "lagged" and "immediate" parameter update approaches (I think). for the latter, only the kinks of these curves gets moved to the current position which is not very good for Newton convergence.

ebos/eclproblem.hh
andlaus: In my mind, as a general rule, everything which does not directly depend on the solution that is to be found for the current time level belongs into the problem, stuff that depends on that solution should be part of the model. I'm thinking in "forward-mode", though.

ewoms/models/blackoil/blackoilprimaryvariables.hh
```diff
@@ -465,20 +480,41 @@ public:
...
+                RsMax = FluidSystem::rateLimmitedUpdate(So,So0,RsMax,Rs0,dt);
+                assert(RsMax_tmp>=RsMax);
+
+            }
+            if (Rs > RsMax*(1.0 + eps)) {
                 // the gas phase appears, i.e., switch the primary variables to { Sw, po,
                 // Sg }.
                 setPrimaryVarsMeaning(Sw_po_Sg);
-                (*this)[Indices::compositionSwitchIdx] = 0.0; // hydrocarbon gas saturation
+                (*this)[Indices::compositionSwitchIdx] = eps; // hydrocarbon gas saturation
```
totto82: Why is eps added here? 
We used eps her before, but removed it recently because it didn't help the convergence.

totto82:
superseded by #255

PR #233, opened 2017-10-27
Schedule: explicitly instantiated

joakim-hove:
Followup to: https://github.com/OPM/opm-parser/pull/1138

### DIFF
--- a/ebos/eclbasegridmanager.hh
+++ b/ebos/eclbasegridmanager.hh
@@ -38,6 +38,7 @@
 #include <opm/parser/eclipse/EclipseState/EclipseState.hpp>
 #include <opm/parser/eclipse/EclipseState/Grid/EclipseGrid.hpp>
 #include <opm/parser/eclipse/EclipseState/Schedule/Schedule.hpp>
+#include <opm/parser/eclipse/EclipseState/SummaryConfig/SummaryConfig.hpp>
 
 #if HAVE_MPI
 #include <mpi.h>
@@ -103,10 +104,12 @@ public:
      * management of these two objects, i.e., they are not allowed to be deleted as long
      * as the grid manager object is alive.
      */
-    static void setExternalDeck(Opm::Deck* deck, Opm::EclipseState* eclState)
+    static void setExternalDeck(Opm::Deck* deck, Opm::EclipseState* eclState, Opm::Schedule* schedule, Opm::SummaryConfig* summaryConfig)
     {
         externalDeck_ = deck;
         externalEclState_ = eclState;
+        externalSchedule_ = schedule;
+        externalSummaryConfig_ = summaryConfig;
     }
 
     /*!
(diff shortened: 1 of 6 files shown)


joakim-hove:
@andlaus : The current PR does not compile:
```txt
hove@joakim:~/work/OPM/ewoms/build$ make
[  0%] Making "tests" data available in output tree
[ 94%] Built target datafiles
[ 94%] Built target tests
[ 94%] Built target examples
[ 96%] Patching Makefile to be DUNE compatible
[ 96%] Built target dune-compat
[ 98%] Linking CXX executable bin/ebos
CMakeFiles/ebos.dir/ebos/ebos.cc.o: In function `Ewoms::EclBaseGridManager<Ewoms::Properties::TTag::EclProblem>::EclBaseGridManager(Ewoms::Simulator<Ewoms::Properties::TTag::EclProblem>&)':
ebos.cc:(.text._ZN5Ewoms18EclBaseGridManagerINS_10Properties4TTag10EclProblemEEC2ERNS_9SimulatorIS3_EE[_ZN5Ewoms18EclBaseGridManagerINS_10Properties4TTag10EclProblemEEC5ERNS_9SimulatorIS3_EE]+0x1ea): undefined reference to `Ewoms::EclBaseGridManager<Ewoms::Properties::TTag::EclProblem>::externalSchedule_'
collect2: error: ld returned 1 exit status
CMakeFiles/ebos.dir/build.make:114: recipe for target 'bin/ebos' failed
make[2]: *** [bin/ebos] Error 1
CMakeFiles/Makefile2:2591: recipe for target 'CMakeFiles/ebos.dir/all' failed
make[1]: *** [CMakeFiles/ebos.dir/all] Error 2
Makefile:138: recipe for target 'all' failed
make: *** [all] Error 2
```
Any tips on where to fix?

andlaus:
no objections (actually, "no opinion" would be more appropriate), i.e., merge whenever. I recommend waiting until the 2017.10 release has been finalized, though.

andlaus:
for the compilation issue, you need to add an external definition of the `externalSchedule_` variable to the `EclBaseGridManager` class:

```c++
template <class TypeTag>
class EclBaseGridManager {
  // ...
  static Opm::Schedule *externalSchedule_;
};

template <class TypeTag>
Opm::Schedule EclBaseGridManager<TypeTag>::externalSchedule_ = nullptr;
```

that's pretty annoying, but for non-primitive types, c++ requires this for some reason.

joakim-hove:
I'm sorry - but this is above my C++ Pay grade - av bit more details?

andlaus:
note the last two lines which define the static `externalDeck_` variable of the template class, i.e., these two lines are basically what you would put into a separate compile unit if you declare an external variable in plain-old C. C++ requires to also define static variables of classes that way, but the trouble is that in a template context, such a thing is impossible to put into a single compile unit because a definition would be required for each possible instantiation of the template (which are literally infinite), so for template classes these definitions can be put into the header file and the linker then deals with the duplicates. At least that's how I understand the mechanism.

andlaus:
I was wondering: wouldn't it be better if `EclipseState` continued to provide a `schedule()` method and just making the instantiation of `Opm::Schedule` outside of `EclipseState` easy? i.e., to essentially see `EclipseState` convenience class that provides a collection of independent objects. That way, no downstream code would not need to be changed.

joakim-hove:
I>  was wondering: wouldn't it be better if EclipseState continued to provide a schedule() method and just making the instantiation of Opm::Schedule outside of EclipseState easy? 

Yes - I have certainly thought of that; getting all the downstream fixes correct has been quite flickety. The intention of all this fiddling is to facilitate runtime changes to the Schedule sections - i.e. adjusting rates, adding wells and so on. I conjecture that your suggestion will lead to a large amount of misunderstandings & bugs due to the "*Which `Schedule` instance is this?*" question.

PR #234, opened 2017-11-02
Fix OpenMP

andlaus:
This makes multi-threaded linearization with OpenMP  work. (note that the linear solver is a quite different beast.)

this issue was only uncovered because the dune buildsystem branches (OPM/opm-common#277 et al) support enabling OpenMP without too much hassle. (I always had major headaches to get OpenMP into working conditions with the current build system.)

### DIFF
--- a/ewoms/common/parametersystem.hh
+++ b/ewoms/common/parametersystem.hh
@@ -398,7 +398,7 @@ void printUsage(const std::string& progName, const std::string& errorMsg = "",
  *         not be read.
  */
 template <class TypeTag>
-std::string parseCommandLineOptions(int argc, char **argv, bool handleHelp = true)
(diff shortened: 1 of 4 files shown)


andlaus:
okay, this PR requires some downstream mob-up because of changing the `argv` argument of `setupParameters()` from `char**` to `const char**`. OPM/opm-simulators#1346 does this mop-up. if somebody demands to spin off the first patch of this PR: not a big deal

andlaus:
okay, the mop-up has been merged before the actual spill happened. to rescue the build, I'll merge.

atgeirr:
Oops, sorry!

andlaus:
no worries, the build on master was only broken for about half an hour.

PR #236, opened 2017-11-20
Use the Equil initializer directly

totto82:
Do not relay on opm-core objects like phaseUsage and BlackoilState but
instead use the initializer directly.

Depents on OPM/opm-core#1187

### DIFF
--- a/ebos/eclequilinitializer.hh
+++ b/ebos/eclequilinitializer.hh
@@ -36,9 +36,8 @@
 // the ordering of these includes matters. do not touch it if you're not prepared to deal
 // with some trouble!
 #include <dune/grid/cpgrid/GridHelpers.hpp>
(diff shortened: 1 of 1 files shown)


ebos/eclequilinitializer.hh
```diff
@@ -122,31 +121,13 @@ public:
...
-                            gridManager.eclState(),
-                            simulator.problem().gravity()[dimWorld - 1],
-                            opmBlackoilState,
-                            enableSwatinit);
+
+        Opm::EQUIL::DeckDependent::InitialStateComputer<FluidSystem> initialState(equilMaterialLawManager,
+                                                                                  gridManager.eclState(),
+                                                                                  equilGrid,
+                                                                                  simulator.problem().gravity()[dimWorld - 1],
+                enableSwatinit);
```
andlaus: indentations screw-up?

PR #238, opened 2017-11-24
fix the build condition for ebos

andlaus:
Another shenanigan of the current default OPM build system:

this produced a cmake syntax error if ewoms was compiled with only the mandatory opm modules available because some *_FOUND variables expand to the empty string instead of to "0" in this case.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -130,7 +130,7 @@ set(CMAKE_PROJECT_NAME "${PROJECT_NAME}")
 EwomsAddApplication(ebos
                     SOURCES ebos/ebos.cc
                     EXE_NAME ebos
-                    CONDITION ${OPM_GRID_FOUND} AND ${OPM_PARSER_FOUND} AND ${OPM_CORE_FOUND})
+                    CONDITION OPM_GRID_FOUND AND OPM_PARSER_FOUND AND OPM_CORE_FOUND)
 
 if(OPM_GRID_FOUND AND OPM_CORE_FOUND AND OPM_PARSER_FOUND)
   install(TARGETS ebos DESTINATION bin)


andlaus:

blattms:
For the record: This has absolutely nothing to do with the OPM build system, but is rather a bug in your CMake code. *_FOUND variables are only set if a module is found by convention.

blattms:
BTW I am not sure whether UPPER_CASE_MODULE_found will be set forever. You might want to switch to case-sensitive-module_FOUND.

andlaus:
> *_FOUND variables are only set if a module is found by convention.

yeah, but cmake's `find_package_handle_standard_args()` macro sets it IIRC and I'm pretty sure that this used to work as it was.

> BTW I am not sure whether UPPER_CASE_MODULE_found will be set forever. You might want to switch to case-sensitive-module_FOUND.

Thanks for the advice. I was already aware of that but I've decided to scale down my personal involvement in the current build system for the master branches to the bare minimum, so I don't see it as my responsibility to fix somebody's mess if he decides to remove the ALL_UPPERCASE version of these variables. (the branches which use the dune build system used mixed-case variables from day 1, BTW.)

blattms:
Please (re)read
```
cmake --help-command find_package
cmake --help-module FindPackageHandleStandardArgs
```
and wonder about the inconsisties of CMake...

atgeirr:
> and wonder about the inconsisties of CMake...

While I still think it was an improvement to transition away from autotools, it is clear that CMake will eventually be replaced by something *much* simpler and better... We'll ride that wave when it comes, in 2027 or so. By then, I hope C++ will have a module system that can simplify what a build system needs to do significantly. By the time I am ready to retire (in 2043 or so) I hope to be able to annoy newbies with "in the old days you needed to compile and link like this" horror stories.

PR #239, opened 2017-11-28
Fix drsdt

hnil:
Drsdt relatert fanges. Introducing cellValues. Fin bug in somax for solvent ++


andlaus:
does this supersede #230? if yes, would you please close #230?

ewoms/models/blackoil/blackoilmodel.hh
```diff
@@ -567,7 +680,8 @@ private:
         priVars.setPvtRegionIndex(regionIdx);
     }
 
-    std::vector<Scalar> maxOilSaturation_;
+    //std::vector<Scalar> maxOilSaturation_;
+    std::vector< std::array<Scalar,numCellValues> > cellValues_;//Rs value of the prevois time stem used for ratelimmited dissolution
 };
```
totto82: Is it necessary to introduce a array of cellvalues in order to implement DRSDT? If not, I think such change belongs in a separate PR.

ewoms/models/blackoil/blackoilprimaryvariables.hh
```diff
@@ -462,15 +480,30 @@ public:
...
+                RsMax = FluidSystem::rateLimmitedUpdate(So,So0,RsMax,Rs0,dt);
+                assert(RsMax_tmp>=RsMax);
+
+            }
+            if (Rs > RsMax*(1.0 + eps)) {
                 // the gas phase appears, i.e., switch the primary variables to { Sw, po,
                 // Sg }.
                 setPrimaryVarsMeaning(Sw_po_Sg);
-                (*this)[Indices::compositionSwitchIdx] = 0.0; // hydrocarbon gas saturation
+                (*this)[Indices::compositionSwitchIdx] = eps; // hydrocarbon gas saturation
```
totto82: Is this eps introduced to stabilize the DRSDT implementation or is this a suggestion to improve stability in general? I tested this  throughly a few months ago, and ended up removing it, but maybe I was wrong? In any case if not necessary for DRSDT I suggest you remove it from this PR since this may affect performance unrelated to DRSDT

andlaus: mathematically speaking it should be more stable without the epsilon: without it you go to the kink and switch to the other branch of the function, while with it you go *beyond* the kink and switch to the other branch of the function. If the solution which is closer to the kink than the epsilon value, this can lead to osscilations?!

totto82:
superseded by #255

PR #240, opened 2017-11-29
BUGFIX Communicate the default thpress correctly 

totto82:
Verified on Model 2 rel 0 with 1, 2, 4 and 8 cores.  
THPRESS and end results are identical in eyeball norm between the runs.

### DIFF
--- a/ebos/eclthresholdpressure.hh
+++ b/ebos/eclthresholdpressure.hh
@@ -226,8 +226,8 @@ private:
 
         // make sure that the threshold pressures is consistent for parallel
         // runs. (i.e. take the maximum of all processes)
-        for (unsigned i = 0; i < thpres_.size(); ++i)
-            thpres_[i] = gridView.comm().max(thpres_[i]);
+        for (unsigned i = 0; i < thpresDefault_.size(); ++i)
+            thpresDefault_[i] = gridView.comm().max(thpresDefault_[i]);
     }
 
     // internalize the threshold pressures which where explicitly specified via the


totto82:

ebos/eclthresholdpressure.hh
```diff
@@ -166,17 +166,14 @@ private:
...
-        // difference between two EQUIL regions.
+        // loop over the whole grid including the ghosts cells
+        // and compute the maximum gravity adjusted pressure difference between two EQUIL regions.
         auto elemIt = gridView.template begin</*codim=*/ 0>();
         const auto& elemEndIt = gridView.template end</*codim=*/ 0>();
         ElementContext elemCtx(simulator_);
         for (; elemIt != elemEndIt; ++elemIt) {
 
             const auto& elem = *elemIt;
-            if (elem.partitionType() != Dune::InteriorEntity)
```
andlaus: I guess this should stay. as far as I can see, ghost cells should never be center element? (they can be neighbours, though.)

ebos/eclthresholdpressure.hh
```diff
@@ -226,8 +223,8 @@ private:
 
         // make sure that the threshold pressures is consistent for parallel
         // runs. (i.e. take the maximum of all processes)
-        for (unsigned i = 0; i < thpres_.size(); ++i)
-            thpres_[i] = gridView.comm().max(thpres_[i]);
+        for (unsigned i = 0; i < thpresDefault_.size(); ++i)
+            thpresDefault_[i] = gridView.comm().max(thpresDefault_[i]);
```
andlaus: oops, that was embarrassing. thanks!

ebos/eclthresholdpressure.hh
totto82: You are right this doesn't change anything. I will remove it.

andlaus:
thanks for that catch!

PR #241, opened 2017-11-29
Prepare for using the initial solution from ebos directly 

totto82:
- Let ebos handle the swatinit scaling
- Keep the initial solution a bit longer so flow can output it. 
- Stop using the equilGrid in the init code. 
(The equilGrid can most probably be removed now. But not in this PR.) 

Note. Merge together with associated PR in opm-simulators.

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -119,9 +119,6 @@ NEW_PROP_TAG(DisableWells);
 // print statements in debug mode.
 NEW_PROP_TAG(EnableDebuggingChecks);
 
-// If this property is set to false, the SWATINIT keyword will not be handled by ebos.
-NEW_PROP_TAG(EnableSwatinit);
-
 // Set the problem property
 SET_TYPE_PROP(EclBaseProblem, Problem, Ewoms::EclProblem<TypeTag>);
 
@@ -238,8 +235,6 @@ SET_BOOL_PROP(EclBaseProblem, DisableWells, false);
 // By default, we enable the debugging checks if we're compiled in debug mode
 SET_BOOL_PROP(EclBaseProblem, EnableDebuggingChecks, true);
 
-// ebos handles the SWATINIT keyword by default
-SET_BOOL_PROP(EclBaseProblem, EnableSwatinit, true);
 } // namespace Properties
 
 /*!
@@ -556,6 +551,10 @@ public:
             // write the summary information after each time step
             summaryWriter_.write(wellManager_);
         }
+
+        // we no longer need the initial soluiton
+        if (this->simulator().episodeIndex() == 0)
+            initialFluidStates_.clear();
     }
 
     /*!
@@ -972,26 +971,8 @@ public:
         // the initial solution.
         thresholdPressures_.finishInit();
 
-        // apply SWATINIT if requested by the programmer. this is only necessary if
-        // SWATINIT has not yet been considered at the first time the EquilInitializer
-        // was used, i.e., only if threshold pressures are enabled in addition to
-        // SWATINIT.
-        const auto& deck = this->simulator().gridManager().deck();
-        const auto& eclState = this->simulator().gridManager().eclState();
-        int numEquilRegions = eclState.getTableManager().getEqldims().getNumEquilRegions();
-        bool useThpres = deck.hasKeyword("THPRES") && numEquilRegions > 1;
-        bool useSwatinit =
-            GET_PROP_VALUE(TypeTag, EnableSwatinit) &&
-            eclState.get3DProperties().hasDeckDoubleGridProperty("SWATINIT");
-
-        if (useThpres && useSwatinit)
-            applySwatinit();
-
         // release the memory of the EQUIL grid since it's no longer needed after this point
         this->simulator().gridManager().releaseEquilGrid();
-
-        // the fluid states specifying the initial condition are also no longer required.
-        initialFluidStates_.clear();
     }
 
     /*!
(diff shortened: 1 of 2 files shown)


andlaus:
this looks good and makes things simpler, so I'm all for merging. Nevertheless, I've got three minor questions/demands:

* can you please squash these commits together? I'd like to keep the git history for the master branch clean, i.e., debug commits  should not be in it IMO
* can this be merged without preconditions or does it need patches for upstream modules?
* does this affect the results or the performance of flow in any significant way?

andlaus:
The second point is already answered by your PR description: OPM/opm-simulators#1355 is needed. I somehow missed this, sorry.

totto82:
> can you please squash these commits together? I'd like to keep the git history for the master branch clean, i.e., debug commits should not be in it IMO

That was my plan. I wasn't 100% sure about the last commit so I kept the option to revert it easily. 
Can EquilGrid be completely removed now? I have tested that this patch works with MPI. 

> can this be merged without preconditions or does it need patches for upstream modules?

NO

> does this affect the results or the performance of flow in any significant way?

I guess the time for the initialization is halved, but other than that NO.

andlaus:
> Can EquilGrid be completely removed now?

if the hydrostatic code from opm-core now works with a distributed grid, yes. does it?

ebos/eclproblem.hh
```diff
@@ -556,6 +551,10 @@ public:
             // write the summary information after each time step
             summaryWriter_.write(wellManager_);
         }
+
+        // we no longer need the initial soluiton
+        if (this->simulator().episodeIndex() == 0)
+            initialFluidStates_.clear();
```
andlaus: why did you move this from `initialSolutionApplied()` to here?

totto82: Its needed for output in flow (check the associated PR in opm-simulators).

totto82:
> if the hydrostatic code from opm-core now works with a distributed grid, yes. does it?

Yes it does. Could you check if it work with Alugrid as well? Or is this tested automatically.

ebos/eclproblem.hh
andlaus: hm, I had a brief look and it looks like one can just use the fluid state from the `IntensiveQuantities` there? the solution should not have changed compared to the initial one, so if there is a difference in the initial fluid state and `iq.fluidState()` that would be a bug as far as I can see.

andlaus:
> Yes it does.

cool, then all these `equilGrid` hacks can be removed!

> Could you check if it work with Alugrid as well? Or is this tested automatically.

It's not tested automatically. I'll give it a shot and report back.

andlaus:
Okay, I've tested: It currently does not compile with alugrid. That's because `Opm::initStateEquil()` uses `Opm::UgGridHelpers` extensively and this is neither available nor easy to provide for AluGrid. As far as I can see, it is not very difficult to replace these constructs by DUNE equivalents in the `Opm::initStateEquil()` code, but that would kill compatibility with the old `UnstructuredGrid` approach. (Not sure if this is acceptable at this stage.)

It also did not compile with `PolyhedralGrid`, but this seemed to be for different reasons related to some issues within `PolyhedralGrid`.

totto82:
OK. I will revert the removal of the EquilGrid in this PR since it is not the main target of this PR.

andlaus:
> OK. I will revert the removal of the EquilGrid in this PR since it is not the main target of this PR.

as you prefer. I think that the issue with the alternative grids should not block the EquilGrid removal, at least not if the init code is scheduled to be adapted to properly work with Dune grids in the not-too-distant future...

ebos/eclproblem.hh
totto82: You are almost right. The state (rs, rv) is changed for cells with only water.

andlaus: > The state (rs, rv) is changed for cells with only water.

I see. since Rs and Rv does not matter in this case, wouldn't it be better to output the initial solution which is actually used by the simulator? (granted, that would almost certainly require another update of the reference data for the unit tests.)

totto82: Yes, it doesn't matter for the results, but our customer wanted it this way. I think the solution I suggest here is as clean as it gets without moving the output code to ewoms.

totto82:
> as you prefer. I think that the issue with the alternative grids should not block the EquilGrid removal, at least not if the init code is scheduled to be adapted to properly work with Dune grids in the not-too-distant future...

I will go for the current solution then. ie no changes to current PR. 
It seems the polyhedral grid is not working for flow with the current master anyway. And ebos itself seem to compile  and run fine.

ebos/eclproblem.hh
andlaus: yeah, it's not a big deal. the question I'm more concerned about is: isn't what we currently do misleading?

Also, I've just looked a bit closer: is `initialFluidStates_` even accessible from outside of the problem class? I did not see any accessor method...

ebos/eclproblem.hh
```diff
@@ -1026,48 +1007,9 @@ public:
...
-            priVars0 = priVars1;
-        }
-
-        // the solution (most likely) has changed, so we need to invalidate the cache for
-        // intensive quantities
-        this->simulator().model().invalidateIntensiveQuantitiesCache(/*timeIdx=*/0);
+    // temporary solution to facilitate output of initial state from flow
+    const ScalarFluidState& initialFluidState(unsigned globalDofIdx ) const {
+        return initialFluidStates_[globalDofIdx];
     }
```
totto82: @andlaus Here it is.

ebos/eclproblem.hh
andlaus: forget the second comment: I was on master instead of on your branch...

ebos/eclproblem.hh
andlaus: yes. That was a case of "lost in branches" for me.

PR #242, opened 2017-11-29
[bugfix] make code compile when Scalar != double.

dr-robertk:
This fixes two minor issues when Scalar is not double.


andlaus:
thanks. for the polymer tables its probably slightly more explicit to use `std::vector<double>` instead of `auto`, but I'm okay with it as-is. Let's merge.

PR #243, opened 2017-12-01
replace the own copy of the ConditionalStorage class by the one in opm-common

andlaus:
the opm-common variant is also a bit more fool proof...


andlaus:

andlaus:
no difference on CI. merging!

PR #245, opened 2017-12-06
black-oil: do not write the saturated oil formation factor to VTK anymore

andlaus:
this quantity is pretty useless, and the analogous one for wet gas was never written...


andlaus:
jenkins build this please

andlaus:
jenkins build this please

andlaus:
since both CIs are okay with this, there is no reason to hold back IMO.

PR #247, opened 2017-12-08
Add activeToCanonicalComponentIndex map

totto82:
The inverse map may also be useful.


andlaus:
looks good. don't we need this for `BlackOilTwoPhaseIndices` as well?

andlaus:
forget about it. I didn't look well enough. meging.

PR #253, opened 2017-12-15
Start using the BlackoilFluidState

totto82:
Some follow up based on OPM/opm-material#268


andlaus:
jenkins build this with downsteams opm-simulators=1373 opm-data=290 please

andlaus:
since jenkins is happy, let's not hold this back. thanks for the patch.

PR #254, opened 2017-12-18
Self sustained equil init

andlaus:
This copies the code which calculated the hydrostatic initial conditions over from opm-core and cleans it up a bit at that opportunity. note that this changes the tolerance for the initial saturations from 1e-6 to 1e-10 which requires a new reference solution for SPE5CASE1.


andlaus:
jenkins build this with downstreams opm-data=292 please

andlaus:
jenkins is happy. if there are no objections, I'll merge this on Wednesday...

totto82:
Looks good.

andlaus:
if you need this today, e.g., because it could lead to conflicts for stuff you are currently working on, you have my blessing for merging it right now.

totto82:
@atgeirr is back tomorrow. Lets wait for him.

andlaus:
Ah, I forgot: With this, the UgGridHelpers  stuff can finally be replaced in that code by direct usage of the DUNE  grid interface without breaking anything. this would enable alternative  grids like Dune::PolyhedralGrid or Dune::ALUGrid again. Unfortunately, it is quite a bit of non-trivial work.

atgeirr:
A few comments:
 - The old code should have been moved over first with no changes, preserving commit history (good for crediting and of course precise git blame when things don't work...)
 - The cleanup should have been a separate PR. Now it is hard to tell what are the cleanups (or perhaps those are the commits after the first one?) 
 - Then finally the change to the tolerance should have been a separate PR (as that is from my understanding the only change here that actually causes different behaviour).
 - Changing grid interface here should be in a separate PR!

andlaus:
> The old code should have been moved over first with no changes, preserving commit history (good for crediting and of course precise git blame when things don't work...)

the "move without changes" thing is done by the first commit (almost no changes; except those needed to keep it compiling). for the history: @akva2: you have a script to do this right? 

> The cleanup should have been a separate PR. Now it is hard to tell what are the cleanups (or perhaps those are the commits after the first one?)

as usual, I tried to make the PR fully bisectable. since the first commit is basically just a verbatim copy of the old code, you need to look at the individual commits. If you insist, I don't have a big problem in doing it in two steps, though.

> Changing grid interface here should be in a separate PR!

the stuff in this one basically boils down to replacing `UgGridHelpers::numCells(grid)` by `grid.size(0)`. are you sure this is necessary? Be aware that grid usage here will be cleaned up more thoroughly later anyway.

akva2:
i have a couple of scripts. i'd prefer instructions rather than trying to teach how to use my hackery if that works for you. it takes some adaption to get it right for various situations that may pop up.

andlaus:
> i have a couple of scripts. i'd prefer instructions rather than trying to teach how to use my hackery if that works for you. it takes some adaption to get it right for various situations that may pop up.

okay: I need the following files and their associated history to be moved from opm-core to eWoms:

```
opm/core/simulator/EquilibrationHelpers.hpp
opm/core/simulator/initStateEquil.hpp
opm/core/simulator/initStateEquil_impl.hpp
opm/core/utility/RegionMapping.hpp
```

please send me a pointer to the branch in your github fork once you're done. I intend to rebase the present PR on your branch.

akva2:
okay. i will have to bring over various tests (test_equil, test_regionmapping, test_satfunc) and one example, compute_initial_state. ewoms is not prepared for unit tests however...

akva2:
i see you change paths. do you want it reflected at that path in entire history, or do you want to do the moves afterwards ?

andlaus:
> okay. i will have to bring over various tests  (test_equil, test_regionmapping, test_satfunc) and one example,  compute_initial_state. ewoms is not prepared for unit tests however.

the test can probably be omitted, they are probably quite a bit too much bound to unrelated opm-core infrastructure and the flow tests in opm-simulater code a quite thoroughly than the opm-core unittests anyway?

if there are good reasons to keep them anyway, I'll adapt the tests to get rid of the Boost unit test framework.

akva2:
i have no idea but in general gut reaction is that removing unit tests sounds like a step backwards.

andlaus:
> i see you change paths. do you want it reflected at that path in entire history, or do you want to do the moves afterwards ?

I think it is better to have the history verbatim and change the paths afterwards.

akva2:
okay. once i get an answer on the test thing, i will do this. won't be today though.

andlaus:
> i have no idea but in general gut reaction is that removing unit tests sounds like a step backwards.

I think it depends on how much added value these tests provide over those flow tests in opm-simulators? If that amount is too low, they just slow refactorings down.

atgeirr:
> I think it depends on how much added value these tests provide over those flow tests in opm-simulators? If that amount is too low, they just slow refactorings down.

I think that in general we should aim for *more* unit(-ish) tests rather than less, so I think those that are still relevant (meaning that the code they test is moved here) should be brought over. If any classes remain in the end that are only needed by their tests we can of course remove them...

andlaus:
yes, I agree that we should *test* more thoroughly. IMO this implies better code coverage in tests, but not necessarily more tests!

but I guess there is no harm if you salvage them @akva2. Things that turn out to be useless, can still be deleted within eWoms...

akva2:
import_equil@akva2/ewoms
remove_equil@akva2/opm-core

akva2:
and since this was completely straight forward, here's the how:

filter-branch.sh
```bash
#!/bin/bash

SCRIPTDIR=`dirname $0`
FILELIST=`git ls-files -c | grep -Fxf $SCRIPTDIR/file_list -v | tr '\n' ' '`
rm $FILELIST

exit 0
```

parent-filter.rb:
```ruby
#!/usr/bin/ruby
old_parents = gets.chomp.gsub('-p ', ' ')

if old_parents.empty? then
  new_parents = []
else
  new_parents = `git show-branch --independent #{old_parents}`.split
end

puts new_parents.map{|p| '-p ' + p}.join(' ')
```

in same folder as filter-branch.sh stick a file called file_list with the files to keep, one per line

then do the filtering:
1) git clone https://github.com/OPM/opm-core opm-core-filtered
2) git clone ewoms
2) cd opm-core-filtered
3) git filter-branch --tree-filter path/to/filter-branch.sh --parent-filter path/to/parent_filter.rb --prune-empty
4) cd ../ewoms
5) git pull ../opm-core-filtered. edit commit message to what you want.

atgeirr:
Thanks @akva2 for the history preservation, code archeologists everywhere are happy...!

@andlaus you can keep this open as a discussion forum if you want, but I assume that you will open new PRs?

andlaus:
okay, I rebased on top of your branch, now the history is included. I deleted the old unit tests because they are IMO too closely tied to the opm-core infrastructure and the value which they provide over the flow tests from opm-simulators is probably not worth the two or three working days to beat them into shape. (if somebody disagrees, I'm happy to accept patches which restore them.)

joakim-hove:
> and the value which they provide over the flow tests from opm-simulators is probably not worth

Well - tests which are near the code they are testing are certainly more valuable than tests which are far away; i.e. in another repository.

andlaus:
> Well - tests which are near the code they are testing are certainly more valuable than tests which are far away; i.e. in another repository.

do you volunteer to beat them into shape?

joakim-hove:
> do you volunteer to beat them into shape?

No - and quite honestly I think that kind of extortion on your behalf is not ok; I just merely pointed that your actions are reducing the test coverage and quality of testing in OPM - I think that is a move in the wrong direction.

andlaus:
Don't get me wrong, I'd also prefer to keep them; I just think that the cost/benefit ratio to do so is too high.

andlaus:
> @andlaus [...] I assume that you will open new PRs?

a  new PR is not required, I rebased on top of @akva2's branch, so the history is conserved here, too.

atgeirr:
> Don't get me wrong, I'd also prefer to keep them; I just think that the cost/benefit ratio to do so is too high.

At least test_equil and  test_regionmapping apply directly to code that has been moved here, so they should be working. I do not understand what the cost is, other than minor changes corresponding to the modifications you did?

andlaus:
>>  Don't get me wrong, I'd also prefer to keep them; I just think that the cost/benefit ratio to do so is too high.
>
>At least test_equil and test_regionmapping apply directly to code that has been moved here, so they should be working. I do not understand what the cost is, other than minor changes corresponding to the modifications you did?

let me explain: the problem is that these unit tests include quite a bit of stuff from opm-core just to set up the infrastructure around the test (`UnstructuredGrid` anyone?) and also, they use boost::unit_test_framework and I would really like to avoid depending on boost modules that require libraries because they are a pain in the butt with the ABI change done in GCC 6. If you disagree with the last point try to compile the opm-parser unit tests with clang on Ubuntu 16.04.

Moreover, as I see this, this PR is just a prelude to a more thorough cleanup of these classes. Once that happens, the unit tests will basically need to be rewritten anyway.

finally, I'm pretty sure that the stuff  in opm-core must be kept for `flow_legacy`, so the unit tests would stick around anyway.

akva2:
the equil tests are a bit meh due to the the unreadable to mere mortals type mess;
```
In file included from /home/akva/kode/opm/sibling/ewoms/ewoms/io/basegridmanager.hh:30:0,
                 from /home/akva/kode/opm/sibling/ewoms/ebos/eclbasegridmanager.hh:30,
                 from /home/akva/kode/opm/sibling/ewoms/ebos/eclcpgridmanager.hh:30,
                 from /home/akva/kode/opm/sibling/ewoms/ebos/eclproblem.hh:52,
                 from /home/akva/kode/opm/sibling/ewoms/tests/test_equil.cpp:21:
/home/akva/kode/opm/sibling/ewoms/ewoms/common/propertysystem.hh: In instantiation of ‘const bool Ewoms::Properties::propertyDefinedOnSelf<Opm::FluidSystems::BlackOil<double>, Opm::FluidSystems::BlackOil<double>, Ewoms::Properties::PTag::FluidSystem>::value’:
/home/akva/kode/opm/sibling/ewoms/ewoms/common/propertysystem.hh:984:90:   required from ‘struct Ewoms::Properties::GetProperty<Opm::FluidSystems::BlackOil<double>, Ewoms::Properties::PTag::FluidSystem>’
/home/akva/kode/opm/sibling/ewoms/ebos/equil/initstateequil.hh:929:58:   required from ‘class Ewoms::EQUIL::DeckDependent::InitialStateComputer<Opm::FluidSystems::BlackOil<double> >’
/home/akva/kode/opm/sibling/ewoms/tests/test_equil.cpp:423:73:   required from here
/home/akva/kode/opm/sibling/ewoms/ewoms/common/propertysystem.hh:832:9: error: no type named ‘SelfType’ in ‘class Opm::FluidSystems::BlackOil<double>’
         ! std::is_base_of<PropertyUndefined,

[ 96%] Building CXX object CMakeFiles/test_equil.dir/tests/test_equil.cpp.o
In file included from /home/akva/kode/opm/sibling/ewoms/ewoms/io/basegridmanager.hh:30:0,
                 from /home/akva/kode/opm/sibling/ewoms/ebos/eclbasegridmanager.hh:30,
                 from /home/akva/kode/opm/sibling/ewoms/ebos/eclcpgridmanager.hh:30,
                 from /home/akva/kode/opm/sibling/ewoms/ebos/eclproblem.hh:52,
                 from /home/akva/kode/opm/sibling/ewoms/tests/test_equil.cpp:21:
/home/akva/kode/opm/sibling/ewoms/ewoms/common/propertysystem.hh: In instantiation of ‘struct Ewoms::Properties::GetProperty<Opm::FluidSystems::BlackOil<double>, Ewoms::Properties::PTag::FluidSystem>::SearchSplicesThenChildren_<Opm::FluidSystems::BlackOil<double>, void>’:
/home/akva/kode/opm/sibling/ewoms/ewoms/common/propertysystem.hh:981:66:   required from ‘struct Ewoms::Properties::GetProperty<Opm::FluidSystems::BlackOil<double>, Ewoms::Properties::PTag::FluidSystem>::GetEffectiveTypeTag_<Opm::FluidSystems::BlackOil<double>, false>’
/home/akva/kode/opm/sibling/ewoms/ewoms/common/propertysystem.hh:984:90:   required from ‘struct Ewoms::Properties::GetProperty<Opm::FluidSystems::BlackOil<double>, Ewoms::Properties::PTag::FluidSystem>’
/home/akva/kode/opm/sibling/ewoms/ebos/equil/initstateequil.hh:929:58:   required from ‘class Ewoms::EQUIL::DeckDependent::InitialStateComputer<Opm::FluidSystems::BlackOil<double> >’
/home/akva/kode/opm/sibling/ewoms/tests/test_equil.cpp:423:73:   required from here
/home/akva/kode/opm/sibling/ewoms/ewoms/common/propertysystem.hh:977:83: error: no type named ‘ChildrenTuple’ in ‘class Opm::FluidSystems::BlackOil<double>’
     { typedef typename SearchTypeTagTuple_<typename CurTree::ChildrenTuple>::type type; };
```

but the two others were straight forward to update; https://pastebin.com/Qz6XGUbq

this disclosed that the satfunc test doesn't actually need the equil stuff it included, so we can undo the move from opm-core if we want.

andlaus:
@akva2:  unfortunately `test_equil` seems to me like the only unit test which has some additional value over the full `flow` integration tests :/. thanks, I'll have a look. what do you need the property system for, here?

@atgeirr:
> At least test_equil and test_regionmapping apply directly to code that has been moved here, so they should be working.

if the boost issue is solved, maybe. Even then, at least `test_regionmapping` does not provide much value -- besides `test_satfunc` seemingly does not have anything to do with the EQUIL code --, because I think `RegionMapping` should go to the chopping block since it obscures things quite a bit more than necessary. at least this applies to the eWoms version of that class.

andlaus:
> the satfunc test doesn't actually need the equil stuff it included, so we can undo the move from opm-core if we want.

yes, please! I'll rebase once you're done.

joakim-hove:
> if the boost issue is solved, maybe. Even then, at least test_regionmapping does not provide much value -- besides test_satfunc seemingly does not have anything to do with the EQUIL code --, because I think RegionMapping should go to the chopping block since it obscures things quite a bit more than necessary. at least this applies to the eWoms version of that class.


image

akva2:
i need access to Ewoms::EQUIL::phasePressures etc and i thus have to include initstateequil.hh. and that triggers a lot of type system issues. that spawned
```[ 96%] Building CXX object CMakeFiles/test_equil.dir/tests/test_equil.cpp.o
In file included from /home/akva/kode/opm/sibling/ewoms/ebos/equil/initstateequil.hh:35:0,
                 from /home/akva/kode/opm/sibling/ewoms/tests/test_equil.cpp:21:
/home/akva/kode/opm/sibling/ewoms/ewoms/common/propertysystem.hh:147:27: error: ‘FluidSystem’ is not a member of ‘Ewoms::Properties::PTag’
 #define PTAG(PropTagName) Ewoms::Properties::PTag::PropTagName
                           ^
/home/akva/kode/opm/sibling/ewoms/ewoms/common/propertysystem.hh:487:47: note: in expansion of macro ‘PTAG’
     ::Ewoms::Properties::GetProperty<TypeTag, PTAG(PropTagName)>::p::type
                                               ^
/home/akva/kode/opm/sibling/ewoms/ebos/equil/initstateequil.hh:929:22: note: in expansion of macro ‘GET_PROP_TYPE’
     typedef typename GET_PROP_TYPE(TypeTag, FluidSystem) FluidSystem;
with a lot of similar friends
```

i included eclproblem.hh to get a bit further, that gives the one above.

andlaus:
First, thanks for digging into this, it's appreciated!

> i included eclproblem.hh to get a bit further, that gives the one above.

did you pass `TTAG(EclProblem)` as the TypeTag argument for the `InitialStateComputer` class? the template parameters are currently quite a mess...

akva2:
i did what explained plus replaced the Opm namespace with the Ewoms namespace and that was it.

andlaus:
> i did what explained plus replaced the Opm namespace with the Ewoms namespace and that was it.

then that's probably the reason for the error...

akva2:
i don't understand what 'that' means. i included the new, equilvalent headers and replaced the old symbols. i am completely ignorant to how these things work, i just did the monkey work :)

andlaus:
> i don't understand what 'that' means.

'that' means passing a `MaterialLawManager` instead of a `TypeTag` to `InitialStateComputer`.  this was a change which I thought will help to considerably clean up the equil code in the future, because passing TypeTags around is how everything else within eWoms works. I did not consider that it will cause problems with legacy code when doing it. sorry for that.

> i just did the monkey work :)

if *that* is your idea of monkey work, I'd like to be a monkey, too ;)

andlaus:
PR rebased on top of the rewritten import branches.

akva2:
as for passing MaterialLawManager as template parameter to InitialStateComputer, that doesn't help as it's the mere inclusion of initialstateequil.hh that triggers the FluidSystem member issue. if i include eclproblem and does that, i get rid of the FluidSystem error as before, but i get the same SelfType error.

atgeirr:
Thanks for taking this seriously, @akva2 and @andlaus. I think that test_equil is a fairly important test, as it tests the equilibration in isolation, without any "interference" from other parts of the simulator. @joakim-hove I have some trouble interpreting your comment other than as a shoe commercial... could you explain?

alfbr:
> shoe commercial... 

My wild guess is a reference to "just do it!" :-)

totto82:
I agree that test_equil or something similar should be ported together with the initialization code. I tried to remove most of its dependency to other opm-core files, but maybe I missed something. How hard is it to remove the Boost dependency? Can't we just write our own CHECK_CLOSE etc.

akva2:
rewriting the test macros is trivial. what i got stuck at is setting up the ewoms machinery, as the current tests rely on UnstructuredGrid stuff which is incompatible. we only need a cart-grid so it should be doable through cpgrid.

andlaus:
I've spend most of the last two working days on fixing that unit test. with the latest version of this branch it now works and does not depend on opm-core and boost anymore. I still think that the value of this test is dubious, though.

andlaus:
> current tests rely on UnstructuredGrid stuff which is incompatible

"luckily" this is not the case because `UnstructuredGrid` is now part of opm-grid even though it is still included via `opm/core/...`. I suppose it is time to clean up the mess that is the location of header files. (e.g. opm-simulators and opm-common have a similar situations.)

that said, the equil test does not use  `UnstructuredGrid` for much, so the effort required to remove it would hopefully not be too big.

andlaus:

akva2:
right, i thought it was problematic due to the Grid typetag stuff (which i don't really understand how works but i tried to monkey some other setups..).

andlaus:
you're right; the type tag stuff was problematic (even more so, the parameter system because it did not allow to runtime parmeters after they were accessed for the first time). Initially I used a separate type tag for each test case, but that lead to unacceptable compile times, so I've modified the parameter system instead...

andlaus:
jenkins build this with downstreams opm-data=292 please

andlaus:
jenkins build this with downstreams opm-data=292 please

andlaus:
jenkins build this with downstreams opm-data=292 please

andlaus:
after a little bit of fighting with the build system, jenkins is now happy. are there still any objections to merging?

atgeirr:
> are there still any objections to merging?

No I guess this can be merged now. Just be aware that if this code cannot be used by the legacy sim, we have not quite resolved all that we set out to do...

atgeirr:
Ouch, noticed now that the opm-data PR has a conflict! Can you resolve that quickly, @andlaus? Otherwise I'll have to revert here temporarily and I'd like to avoid that.

andlaus:
> Ouch, noticed now that the opm-data PR has a conflict! Can you resolve that quickly, @andlaus? Otherwise I'll have to revert here temporarily and I'd like to avoid that.

done. this hiccup was due to unfortunate timing.

andlaus:
> Just be aware that if this code cannot be used by the legacy sim, we have not quite resolved all that we set out to do...

when it comes to the opm-core elimination, true. for `flow_legacy` I would prefer not to touch the current code in any significant way anyway, though: IMO, while changing stuff like the header locations is not a problem, changing, e.g. the grid interface would miss the point of `flow_legacy`.

atgeirr:
Thanks for the opm-data fix, while we were broken for some time I am glad we avoided reverting.

> when it comes to the opm-core elimination, true.

For that particular aspect, the ugly yet simple solution is to just move the stuff from opm-core to opm-simulators. In the longer run I think we should make the initialization code independent enough to expose it as an easy-to-use library. That would be valuable to software pre-processing or creating cases for example.

PR #255, opened 2017-12-19
Add support for DRSDT

totto82:
This is a rebased, squashed and cleaned version of #239 

@hnil Did I miss something? 
@andlaus Can you review this? 

Tested on SPE1CASE1, SPE1CASE2, SPE1CASEINF and SPE1SPEMID

TODO: test it on the bigmodels to make sure they doesn't regress. 

Merged together with OPM/opm-material#269

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -477,6 +477,9 @@ public:
             Scalar initialDt = EWOMS_GET_PARAM(TypeTag, Scalar, InitialTimeStepSize);
             dt = std::min(dt, initialDt);
         }
+        // The updateCell values is called before dt is updated to since it is based on the prevois step.
+        // this put the model state set values which should be from the previous step
+        this->model().updateCellValues();
 
         if (nextEpisodeIdx < numReportSteps) {
             simulator.startNextEpisode(episodeLength);
@@ -485,8 +488,6 @@ public:
 
         if (updateHysteresis_())
             this->model().invalidateIntensiveQuantitiesCache(/*timeIdx=*/0);
-
-        this->model().updateMaxOilSaturations();
 
         if (GET_PROP_VALUE(TypeTag, EnablePolymer))
             updateMaxPolymerAdsorption_();
--- a/ewoms/models/blackoil/blackoilintensivequantities.hh
+++ b/ewoms/models/blackoil/blackoilintensivequantities.hh
@@ -59,6 +59,7 @@ class BlackOilIntensiveQuantities
     typedef typename GET_PROP_TYPE(TypeTag, IntensiveQuantities) Implementation;
 
     typedef typename GET_PROP_TYPE(TypeTag, Scalar) Scalar;
+    typedef typename GET_PROP_TYPE(TypeTag, Model) Model;
     typedef typename GET_PROP_TYPE(TypeTag, Evaluation) Evaluation;
     typedef typename GET_PROP_TYPE(TypeTag, FluidSystem) FluidSystem;
     typedef typename GET_PROP_TYPE(TypeTag, MaterialLaw) MaterialLaw;
@@ -199,7 +200,13 @@ public:
                                                             oilPhaseIdx,
                                                             pvtRegionIdx,
                                                             SoMax);
-                fluidState_.setRs(RsSat);
+                if(FluidSystem::enableRateLimitedDissolvedGas() > Opm::FluidSystems::None){
+                    // limit the updated Rs value by a given rate pr volume
+                    const Evaluation& RsLimited = rateLimitedDissolvedGasUpdate(RsSat, So, elemCtx, dofIdx, timeIdx);
+                    fluidState_.setRs(RsLimited);
+                }else{
+                  fluidState_.setRs(RsSat);
+                }
             }
             else
                 fluidState_.setRs(0.0);
@@ -219,7 +226,12 @@ public:
             // if the switching variable is the mole fraction of the gas component in the
             // oil phase, we can directly set the composition of the oil phase
             const auto& Rs = priVars.makeEvaluation(Indices::compositionSwitchIdx, timeIdx);
-            fluidState_.setRs(Rs);
+            //if(FluidSystem::enableRateLimitedDissolvedGas() == Opm::FluidSystems::All){
+            //    const auto& RsLimited = rateLimitedDissolvedGasUpdate(Rs, So, elemCtx,dofIdx,timeIdx);
+            //    fluidState_.setRs(RsLimited);
+            //} else {
+                fluidState_.setRs(Rs);
+            //}
 
             if (FluidSystem::enableVaporizedOil()) {
                 // the gas phase is not present, but we need to compute its "composition"
@@ -344,6 +356,18 @@ public:
 #endif
     }
 
+    /*! function to update Rs with rate limmitation this is used if keyword DRSDT is present in configuration file
+     */
+    Evaluation rateLimitedDissolvedGasUpdate(const Evaluation& RsE,const Evaluation& So, const ElementContext& elemCtx, unsigned dofIdx, unsigned timeIdx)
+    {
+        assert(timeIdx==0);
+        unsigned globalSpaceIdx = elemCtx.globalSpaceIndex(dofIdx, timeIdx);
+        const Scalar& Rs0 = elemCtx.model().cellValues(globalSpaceIdx,Model::rsPrev);
+        const Scalar& So0 = elemCtx.model().cellValues(globalSpaceIdx,Model::soPrev);
+        const double dt = elemCtx.simulator().timeStepSize();
+        return FluidSystem::rateLimitedUpdate(So, So0, RsE, Rs0, dt);;
+
+    }
     /*!
      * \copydoc ImmiscibleIntensiveQuantities::fluidState
      */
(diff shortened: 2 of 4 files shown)


ewoms/models/blackoil/blackoilintensivequantities.hh
```diff
@@ -219,7 +226,12 @@ public:
             // if the switching variable is the mole fraction of the gas component in the
             // oil phase, we can directly set the composition of the oil phase
             const auto& Rs = priVars.makeEvaluation(Indices::compositionSwitchIdx, timeIdx);
-            fluidState_.setRs(Rs);
+            //if(FluidSystem::enableRateLimitedDissolvedGas() == Opm::FluidSystems::All){
+            //    const auto& RsLimited = rateLimitedDissolvedGasUpdate(Rs, So, elemCtx,dofIdx,timeIdx);
+            //    fluidState_.setRs(RsLimited);
+            //} else {
+                fluidState_.setRs(Rs);
+            //}
```
totto82: @hnil Why was this commented out?

ebos/eclproblem.hh
```diff
@@ -477,6 +477,9 @@ public:
             Scalar initialDt = EWOMS_GET_PARAM(TypeTag, Scalar, InitialTimeStepSize);
             dt = std::min(dt, initialDt);
         }
+        // The updateCell values is called before dt is updated to since it is based on the prevois step.
+        // this put the model state set values which should be from the previous step
```
andlaus: I don't think that the model should be concerned with "cell values". for one thing what are the "celll values" that are supposed to be updated, for another, models are not specific for element-centered FV. Move the contents of this method to EclProblem, i.e. this call boils down to `updateElementValues_()`  (more descriptive name, please!) in `beginTimeStep()`

ewoms/models/blackoil/blackoilintensivequantities.hh
```diff
@@ -199,7 +200,13 @@ public:
                                                             oilPhaseIdx,
                                                             pvtRegionIdx,
                                                             SoMax);
-                fluidState_.setRs(RsSat);
+                if(FluidSystem::enableRateLimitedDissolvedGas() > Opm::FluidSystems::None){
```
andlaus: I am quite sure that the fluid system should not need to care if the gas-in-oil or the oil-in-gas is rate limited or not, i.e. please move this flag to the model.

ewoms/models/blackoil/blackoilintensivequantities.hh
```diff
@@ -199,7 +200,13 @@ public:
...
+                    // limit the updated Rs value by a given rate pr volume
+                    const Evaluation& RsLimited = rateLimitedDissolvedGasUpdate(RsSat, So, elemCtx, dofIdx, timeIdx);
```
andlaus: I don't think that the `IntensiveQuantities` need to concern itself with the rate limits. the only place where this is needed is the update of the primary variables.

ewoms/models/blackoil/blackoilintensivequantities.hh
andlaus: I think the old version is correct, as the IQs should not need to consider rate limits. that said, please remove commented out code instead of leaving it in place (commented out code just clutters the the files and makes them less maintainable. IMO)

ewoms/models/blackoil/blackoilintensivequantities.hh
```diff
@@ -344,6 +356,18 @@ public:
 #endif
     }
 
+    /*! function to update Rs with rate limmitation this is used if keyword DRSDT is present in configuration file
+     */
+    Evaluation rateLimitedDissolvedGasUpdate(const Evaluation& RsE,const Evaluation& So, const ElementContext& elemCtx, unsigned dofIdx, unsigned timeIdx)
```
andlaus: the IQs do not concern themselfs with updating the primary variables, so this method probably does not belong here.

ewoms/models/blackoil/blackoilmodel.hh
```diff
@@ -220,7 +223,12 @@ public:
     BlackOilModel(Simulator& simulator)
         : ParentType(simulator)
     {}
-
+    enum CellValuesName : unsigned {
+        soPrev = 0,
+        rsPrev = 1,
+        soMax = 2,
+        numCellValues,
+    };
```
andlaus: as @totto82 already mentioned in a comment related to these PRs: changing the approach of how to access the data of the model is out of scope of this PR and should thus be discussed separately. Please remove (for now)

ewoms/models/blackoil/blackoilmodel.hh
```diff
@@ -241,7 +249,11 @@ public:
      */
     void finishInit()
     {
-        maxOilSaturation_.resize(this->numGridDof(), 0.0);
+        cellValues_.resize(numCellValues);
+        cellValues_.resize(this->numGridDof());
+        for(unsigned i=0; i < cellValues_.size() ; ++i){
+            cellValues_[i].fill( 0.0);
+        }
```
andlaus: "cell values" stuff -> please move to a separate PR

ewoms/models/blackoil/blackoilmodel.hh
```diff
@@ -384,8 +396,13 @@ public:
         outstream << priVars.primaryVarsMeaning() << " ";
         outstream << priVars.pvtRegionIndex() << " ";
 
-        if (maxOilSaturation_.size() > 0)
-            outstream << maxOilSaturation_[dofIdx] << " ";
+        if (cellValues_.size()>0){
```
andlaus: same

ewoms/models/blackoil/blackoilmodel.hh
```diff
@@ -422,8 +439,13 @@ public:
         unsigned pvtRegionIdx;
         instream >> pvtRegionIdx;
 
-        if (maxOilSaturation_.size() > 0)
-            instream >> maxOilSaturation_[dofIdx];
+        if (cellValues_.size()>0){
```
andlaus: ditto

ewoms/models/blackoil/blackoilmodel.hh
```diff
@@ -477,14 +499,26 @@ public:
      * match the results of the 'flow' and ECLIPSE 100 simulators.
      */
     Scalar maxOilSaturation(unsigned globalDofIdx) const
```
andlaus: as discussed earlier, this function should not be here. could you please move it to `EclProblem` and add a dummy function to `BlackOilProblem` that just returns 1?

ewoms/models/blackoil/blackoilmodel.hh
```diff
@@ -477,14 +499,26 @@ public:
...
+    { cellValues_[globalDofIdx][soMax] = value; }
+
+    Scalar cellValues(unsigned globalDofIdx,CellValuesName cv) const
+    { return cellValues_[globalDofIdx][cv]; }
+
+    /*!
+     * \brief Sets an elements maximum oil phase saturation observed during the
+     *        simulation (used for restarting from UNRST-files).
+     */
+    void setCellValue(const Scalar& value, unsigned globalDofIdx, CellValuesName cv)
```
andlaus: "cell values" stuff

ewoms/models/blackoil/blackoilmodel.hh
```diff
@@ -494,34 +528,25 @@ public:
      * simulation it sometimes needs to be called after a time step or before an episode
      * starts.
      */
-    void updateMaxOilSaturations()
+    void updateCellValues()
```
andlaus: updating quasi-static data after a time step is the domain of the problem: this method does not make any sense for e.g. vertex-centered FV discretizations

ewoms/models/blackoil/blackoilprimaryvariables.hh
```diff
@@ -381,14 +382,21 @@ public:
                 Scalar po = (*this)[Indices::pressureSwitchIdx];
                 Scalar T = asImp_().temperature_();
                 Scalar SoMax = problem.model().maxOilSaturation(globalDofIdx);
-                Scalar RsSat = FluidSystem::oilPvt().saturatedGasDissolutionFactor(pvtRegionIdx_,
+                Scalar Rs = FluidSystem::oilPvt().saturatedGasDissolutionFactor(pvtRegionIdx_,
                                                                                    T,
                                                                                    po,
                                                                                    So2,
                                                                                    SoMax);
```
andlaus: probably, it is better to add RsMax and RsMin and use

```
Scalar RsMax = std::min(RsSat, RsOld + RsMaxDelta);
Scalar RsMin = std::max(0.0, RsOld - RsMaxDelta);
```

where `RsOld` and `RsMaxDelta` are retrieved from the problem like `SoMax`.

andlaus:
I have a bit more general question related to DRSDT: how should composition changes due to pressure changes in the saturated zone be handled? e.g., pressure increases in an area where all three phases are present and the increase of the saturated Rs increases the value set by DRSDT.

ewoms/models/blackoil/blackoilintensivequantities.hh
totto82: I just wasn't sure why the commented out code was added in the first place.

andlaus:
Ah, I forgot: With this, the `UgGridHelpers` stuff can finally be replaced in that code by direct usage of the DUNE grid interface without breaking anything. this would enable alternative grids like `Dune::PolyhedralGrid` or `Dune::ALUGrid` again. Unfortunately, it is quite a bit of non-trivial work.

PR #257, opened 2017-12-19
BlackOilPrimaryVariables: remove recursion adaptPrimaryVariables()

andlaus:
if the interpretation of the primary variables has been changed and the new variables have been set correctly, there must not be any further changes without an update.


andlaus:

andlaus:
jenkins is happy and I've run the base case of model 2 without observing any ill-effect. merging.

PR #262, opened 2018-01-04
Add ecl output from opm-output

totto82:
The new eclwriter output and restart using the eclIO from opm-output

Adds opm-output as optional dependency 

All tests in opm-simulator pass and MPI restart works. (Test should be added later) 
Convergence and results for Norne, Model 2 rel 0 + rel 5 is unaffected. 

Standard initialization is done prior to restart in order
to compute correct initial thpressure values. This is not
necessary if thpressures are written to the restart file

TODO: 
- [x] Guard output with HAVE_OPM_OUTPUT macro
- [X] Investigate performance change


totto82:
I want the build system to set HAVE_OPM_OUTPUT to 1 if opm-output is found. How can I do that?

andlaus:
the easiest way is to use the [dune build system](https://github.com/OPM/ewoms/blob/master/INSTALL) because you are already done there. for the traditional OPM build system, @akva2 or @blattms can probably help out.

akva2:
add it to the config vars in ewoms-prereqs.h

ebos/collecttoiorank.hh
```diff
@@ -245,64 +238,69 @@ namespace Ewoms
...
                 {
+                    // add missing data to global cell data
+                    for (const auto& pair : localCellData_) {
+                        const std::string& key = pair.first;
+                        std::size_t container_size = globalSize;
+                        auto ret = globalCellData_.insert(key, pair.second.dim,
+                                                std::vector<double>(container_size),
+                                                pair.second.target);
+                        assert(ret.second);
+                        DUNE_UNUSED_PARAMETER(ret.second); //dummy op to prevent warning with -DNDEBUG
```
andlaus: IMO it is better to add the OPM_OPTIM_UNUSED attribute to `ret` variable.

ebos/collecttoiorank.hh
```diff
@@ -314,25 +312,29 @@ namespace Ewoms
             {
                 // we should only get one link
                 if( link != 0 ) {
-                    OPM_THROW(std::logic_error,"link in method pack is not 0 as expected");
+                    OPM_THROW(std::logic_error,"link in method pack is not 0 as execpted");
```
andlaus: typo: expected should be correct here?

ebos/collecttoiorank.hh
```diff
@@ -344,65 +346,55 @@ namespace Ewoms
 
         protected:
             template <class Vector>
-            void write( MessageBufferType& buffer, const IndexMapType& localIndexMap, const Vector& data ) const
+            void write( MessageBufferType& buffer, const IndexMapType& localIndexMap,
```
andlaus: please pick one style: one argument per line or one line for all arguments.

ebos/collecttoiorank.hh
```diff
@@ -423,6 +415,11 @@ namespace Ewoms
...
         }
 
+        const Opm::data::Solution& globalCellData() const
+        {
+            return globalCellData_;
+        }
+
         bool isIORank() const
         {
             return isIORank_;
```
andlaus: possibly it is better to use `return toIORankComm_.rank() == 0;` here. (that is unrelated to this PR, yes.)

ebos/collecttoiorank.hh
```diff
@@ -433,13 +430,32 @@ namespace Ewoms
             return toIORankComm_.size() > 1;
         }
 
+        int localIdxToglobalIdx(const unsigned localIdx) {
```
andlaus: capitalize `global` here. Also, opening brackets of functions and classes  generally are on a separate line in eWoms...

ebos/ecloutputblackoilmodule.hh
```diff
@@ -75,89 +61,200 @@ class EcfvDiscretization;
...
     enum { gasCompIdx = FluidSystem::gasCompIdx };
     enum { oilCompIdx = FluidSystem::oilCompIdx };
 
-    typedef typename ParentType::ScalarBuffer ScalarBuffer;
+    typedef std::vector<Scalar> ScalarBuffer;
 
 public:
     EclOutputBlackOilModule(const Simulator& simulator)
-        : ParentType(simulator)
+        :simulator_(simulator)
```
andlaus: pedantic nit: white space between `:` and `simulator_`

ebos/ecloutputblackoilmodule.hh
```diff
@@ -75,89 +61,200 @@ class EcfvDiscretization;
...
-                             "in the ECL output files");
-        EWOMS_REGISTER_PARAM(TypeTag, bool, EclOutputWriteOilSaturationPressure,
-                             "Include the saturation pressure of oil in the "
-                             "ECL output files");
-    }
-
     /*!
      * \brief Allocate memory for the scalar fields we would like to
-     *        write to the VTK file.
+     *        write to eclIO
```
andlaus: s/eclIO/ECL output files/

ebos/ecloutputblackoilmodule.hh
```diff
@@ -75,89 +61,200 @@ class EcfvDiscretization;
...
+
+        sSol_.clear();
+        cPolymer_.clear();
+        soMax_.clear();
+        pcSwMdcOw_.clear();
+        krnSwMdcOw_.clear();
+        pcSwMdcGo_.clear();
+        krnSwMdcGo_.clear();
+        bubbelPointPressure_.clear();
+        dewPointPressure_.clear();
```
andlaus: are the clears actually needed? I believe that the resizes below already do the trick...

ebos/ecloutputblackoilmodule.hh
```diff
@@ -75,89 +61,200 @@ class EcfvDiscretization;
...
+        if (densityOutput_() && rstKeywords["DEN"] > 0) {
+            rstKeywords["DEN"] = 0;
+            for (unsigned phaseIdx = 0; phaseIdx < numPhases; ++ phaseIdx) {
+                if (!FluidSystem::phaseIsActive(phaseIdx))
+                    continue;
+                density_[phaseIdx].resize(bufferSize,0.0);
+            }
+        }
+        if (viscosityOutput_()) {
+            const bool has_vwat = (rstKeywords["VISC"] > 0) || (rstKeywords["VWAT"] > 0);
```
andlaus: camelCase please!

ebos/ecloutputblackoilmodule.hh
```diff
@@ -75,89 +61,200 @@ class EcfvDiscretization;
...
+            {
+                rstKeywords["VWAT"] = 0;
+                viscosity_[waterPhaseIdx].resize(bufferSize,0.0);
+            }
+            if (FluidSystem::phaseIsActive(oilPhaseIdx) && has_voil > 0)
+            {
+                rstKeywords["VOIL"] = 0;
+                viscosity_[oilPhaseIdx].resize(bufferSize,0.0);
+            }
+            if (FluidSystem::phaseIsActive(gasPhaseIdx) && has_vgas > 0)
```
andlaus: is there a reason not to fold `FluidSystem::phaseIsActive()` into the `has*` variables? Also, you don't need the `> 0` for boolean variables...

ebos/ecloutputblackoilmodule.hh
```diff
@@ -75,89 +61,200 @@ class EcfvDiscretization;
...
+        // Now: Output the same as legacy
+        soMax_.resize(bufferSize,0.0);
+        pcSwMdcOw_.resize(bufferSize,0.0);
+        krnSwMdcOw_.resize(bufferSize,0.0);
+        pcSwMdcGo_.resize(bufferSize,0.0);
+        krnSwMdcGo_.resize(bufferSize,0.0);
+
+        if (rstKeywords["PBPD"] > 0)  {
+            rstKeywords["PBPD"] = 0;
+            bubbelPointPressure_.resize(bufferSize,0.0);
```
andlaus: typo: s/bubbel/bubble/

ebos/ecloutputblackoilmodule.hh
```diff
@@ -75,89 +61,200 @@ class EcfvDiscretization;
...
+        }
+
+        //Warn for any unhandled keyword
+        if (log) {
+            for (auto& keyValue : rstKeywords) {
+                if (keyValue.second > 0) {
+                    std::string logstring = "Keyword '";
+                    logstring.append(keyValue.first);
+                    logstring.append("' is unhandled for output to file.");
+                    Opm::OpmLog::warning("Unhandled output keyword", logstring);
```
andlaus: This is somewhat unrelated: does `OpmLog` work with plain ebos? if no, I guess we'll need to make it work.

ebos/ecloutputblackoilmodule.hh
```diff
@@ -75,89 +61,200 @@ class EcfvDiscretization;
...
+                }
+            }
+        }
+
+        failed_cells_pb_.clear();
```
andlaus: camelCase please

ebos/ecloutputblackoilmodule.hh
```diff
@@ -166,158 +263,660 @@ public:
...
+            if (pressuresOutput_() && oilPressure_.size() > 0) {
+                oilPressure_[globalDofIdx] = Toolbox::value(fs.pressure(oilPhaseIdx));
+                Opm::Valgrind::CheckDefined(oilPressure_[globalDofIdx]);
 
-                    pressure_[phaseIdx][globalDofIdx] = Toolbox::value(fs.pressure(phaseIdx));
-                    Opm::Valgrind::CheckDefined(pressure_[phaseIdx][globalDofIdx]);
-                }
             }
-            if (gasDissolutionFactorOutput_()) {
+            if (gasDissolutionFactorOutput_() && gasDissolutionFactor_.size() > 0) {
```
andlaus: hm. when is `gasDissolutionFactorOutput_()` true, but `gasDissolutionFactor_.size() > 0` not?

ebos/ecloutputblackoilmodule.hh
```diff
@@ -166,158 +263,660 @@ public:
...
+
+            if (solventOutput_() && sSol_.size() > 0) {
+                sSol_[globalDofIdx] = intQuants.solventSaturation().value();
+            }
+
+            if (polymerOutput_() && cPolymer_.size() > 0) {
+                cPolymer_[globalDofIdx] = intQuants.polymerConcentration().value();
+            }
+
+            if (bubbelPointPressureOutput_() && bubbelPointPressure_.size() > 0 )
```
andlaus: s/bubbel/bubble/

ebos/ecloutputblackoilmodule.hh
```diff
@@ -166,158 +263,660 @@ public:
...
+                    dewPointPressure_[globalDofIdx] = Toolbox::value(FluidSystem::dewPointPressure(fs, intQuants.pvtRegionIndex()));
+                }
+                catch (const Opm::NumericalProblem& e) {
+                    const auto globalIdx = elemCtx.simulator().gridManager().grid().globalCell()[globalDofIdx];
+                    failed_cells_pd_.push_back(globalIdx);
+                }
+            }
+
+            if (soMaxOutput_() && soMax_.size() > 0)
+                soMax_[globalDofIdx] = elemCtx.simulator().model().maxOilSaturation(globalDofIdx);
```
andlaus: once #265 is merged, this will become `elemCtx.simulator().problem().maxOilSaturation(globalDofIdx);`

ebos/ecloutputblackoilmodule.hh
```diff
@@ -166,158 +263,660 @@ public:
...
+                                krnSwMdcOw_[globalDofIdx],
+                                globalDofIdx);
+                    matLawManager->gasOilHysteresisParams(
+                                pcSwMdcGo_[globalDofIdx],
+                                krnSwMdcGo_[globalDofIdx],
+                                globalDofIdx);
+                }
+            }
+
+            // hack to make the intial output of rs and rv Ecl compatible.
```
andlaus: I'm not sure if we should do this: ECL seems to output something different than what it uses as initial condition, so it might be a good idea to do the right thing, spare ourselfs the hassle and not replicate that bug. Any opinions @alfbr?

ebos/ecloutputblackoilmodule.hh
```diff
@@ -166,158 +263,660 @@ public:
...
+                                                                                        gasPhaseIdx,
+                                                                                        intQuants.pvtRegionIndex());
+            }
+
+        }
+
+    }
+
+
+    void outputErrorLog(const Simulator& simulator) {
```
andlaus: brace on separate line please.

ebos/ecloutputblackoilmodule.hh
```diff
@@ -166,158 +263,660 @@ public:
...
+        const size_t max_num_cells_faillog = 20;
```
andlaus: camelCase please (and below as well)

ebos/ecloutputblackoilmodule.hh
```diff
@@ -166,158 +263,660 @@ public:
...
+
+        int pb_size = failed_cells_pb_.size(), pd_size = failed_cells_pd_.size();
+        std::vector<int> displ_pb, displ_pd, recv_len_pb, recv_len_pd;
+        const auto& comm = simulator.gridManager().grid().comm();
```
andlaus: you can simplify this to `simulator.gridView().comm()`

ebos/ecloutputblackoilmodule.hh
```diff
@@ -166,158 +263,660 @@ public:
...
+            recv_len_pd.resize(comm.size());
+        }
+
+        comm.gather(&pb_size, recv_len_pb.data(), 1, 0);
+        comm.gather(&pd_size, recv_len_pd.data(), 1, 0);
+        std::partial_sum(recv_len_pb.begin(), recv_len_pb.end(), displ_pb.begin()+1);
+        std::partial_sum(recv_len_pd.begin(), recv_len_pd.end(), displ_pd.begin()+1);
+        std::vector<int> global_failed_cells_pb, global_failed_cells_pd;
+
+        if ( comm.rank() == 0 )
```
andlaus: it's probably a good idea to put this into an `isIoRank_()` method.

ebos/ecloutputblackoilmodule.hh
```diff
@@ -166,158 +263,660 @@ public:
...
+            if (global_failed_cells_pd.size() > max_num_cells_faillog) {
+                errlog << ", ...";
+            }
+            errlog << "]";
+            Opm::OpmLog::warning("Dew point numerical problem", errlog.str());
         }
     }
 
     /*!
      * \brief Add all buffers to the VTK output writer.
```
andlaus: that comment is (and has been) incorrect

ebos/ecloutputblackoilmodule.hh
```diff
@@ -166,158 +263,660 @@ public:
...
+        }
+
+        if ( sol.has("KRNSW_GO" ) ) {
+            krnSwMdcGo_[elemIdx] = sol.data("KRNSW_GO")[globalDofIndex];
+        }
+    }
+
+
+    template <class FluidState>
+    void writeToFluidState(FluidState& fs, unsigned elemIdx) const {
```
andlaus: this function name is probably slightly misleading because it makes you think that it involves I/O. rename to `assignToFluidState()`?

also: opening brace on a separate line please.

ebos/eclproblem.hh
```diff
@@ -206,7 +209,7 @@ SET_BOOL_PROP(EclBaseProblem, EnableVtkOutput, false);
 SET_BOOL_PROP(EclBaseProblem, EnableEclOutput, true);
 
 // also enable the summary output.
-SET_BOOL_PROP(EclBaseProblem, EnableEclSummaryOutput, true);
+SET_BOOL_PROP(EclBaseProblem, EnableEclSummaryOutput, false);
```
andlaus: is this intentional? does this property still have an effect after this PR? if (yes, no), please remove the property altogether.

ebos/eclwriter.hh
```diff
@@ -58,297 +59,168 @@ NEW_PROP_TAG(EnableEclOutput);
...
 
-    typedef BaseOutputWriter::ScalarBuffer ScalarBuffer;
-    typedef BaseOutputWriter::VectorBuffer VectorBuffer;
-    typedef BaseOutputWriter::TensorBuffer TensorBuffer;
+    typedef std::vector<Scalar> ScalarBuffer;
 
-    friend class EclWriterHelper<TypeTag, GridManager>;
 
 public:
     EclWriter(const Simulator& simulator)
```
andlaus: is this class actually used after the PR?

ewoms/disc/common/fvbaseproblem.hh
```diff
@@ -623,22 +623,21 @@ public:
      */
     void writeOutput(bool verbose = true)
     {
+        if (!enableVtkOutput_())
```
andlaus: Hm, not sure if this class should concern itself with VTK or any other specific file format. I'll have to think about how to properly deal with this, though...

totto82:
> add it to the config vars in ewoms-prereqs.cmake

Thanks.

ebos/collecttoiorank.hh
```diff
@@ -122,10 +122,6 @@ namespace Ewoms
                     {
                         const int id = distributedGlobalIndex_[ localIndexMap_[ i ] ];
                         indexMap[ i ] = globalPosition_[ id ] ;
-#ifndef NDEBUG
```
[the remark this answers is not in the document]
totto82: I loop over the overlap region. I.e. I may insert the same cell twice. The overlap region is added for the MPI restart.

ebos/collecttoiorank.hh
```diff
@@ -192,32 +184,33 @@ namespace Ewoms
 
             {
                 std::set< int > send, recv;
+
+                typedef typename GridManager::EquilGrid::LeafGridView EquilGridView;
+                const EquilGridView equilGridView = gridManager.equilGrid().leafGridView() ;
```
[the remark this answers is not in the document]
totto82: It is used here to give the global grid. I guess we can clean this up at a later stage. (If desired)

ebos/collecttoiorank.hh
totto82: Do you mean removing the isIORank_ completely?

ebos/ecloutputblackoilmodule.hh
totto82: I don't know. Not tried to output PBPD with Ebos

ebos/ecloutputblackoilmodule.hh
totto82: Never. I will clean up this. I guess the same comments applied to the other outputs as well.

ebos/eclwriter.hh
totto82: Yes, it is used as a layer between eclproblem and ecloutputblackoilmodule. 
Is EclOuput a better name?

ewoms/disc/common/fvbaseproblem.hh
totto82: Will you fix this later?

ebos/eclwriter.hh
andlaus: if it is used, and writes output files, it can stay as it is. If you prefer `EclOutput` or `EclOutputWriter`, I'm fine with renaming it, too.

ewoms/disc/common/fvbaseproblem.hh
andlaus: yes, but that will be a story for another PR. let's go with what you propose for now.

ebos/collecttoiorank.hh
andlaus: The attribute, yes. it seems to be quite pointless to me.

ebos/ecloutputblackoilmodule.hh
andlaus: `OpmLog` is currently not touched by eWoms at all so it will do whatever is the default. I'll put this on my todo list, for now let's keep you patch as-is.

atgeirr: If the program you are using sets up no logging, then the message will go to the great void in the sky. Using the OpmLog::setupSimpleDefaultLogging() method is enough to get everything printed to the terminal (for more sophisticated behavior one may of course customize logging extensively, as done by Flow).

ebos/ecloutputblackoilmodule.hh
alfbr: Hack is in the eye of the beholder. In this case I believe we are looking at what values to put Rs and Rv when they basically do not exist. Currently they are set similarly as other simulators. Changing how those values are set in flow does not belong in this PR in my opinion, and is a separate discussion.

andlaus: > Hack is in the eye of the beholder. In this case I believe we are looking at what values to put Rs and Rv when they basically do not exist.

yes, they "exist" but they are never used as the initial condition by the simulator yet they are marketed as those . "hack" is the right word for this IMO.

the question I have is whether we should even re-implement such stupid bugs.

totto82: I my opinion the real "hack" is that we use RS=RSSAT for the water only case. This output "hack" just make sure that the initial values are not affected by the "water only hack". I.e. We make sure that the output of the initial  solution actually is correct. I disagree that it is a bug.

andlaus: okay, so it does not affect the total mass of the conserved quantities in the initial solution because of zero saturation? if that's the case, I agree that the ECL behaviour is not a bug, but I still think that what's written should be left to the discretion of the implementation.

totto82:
Output is now about 3 times faster than the old code.

andlaus:
cool! is this number for model 2 and/or Norne?

totto82:
Both.

totto82:

andlaus:
I think this is ready to be merged. any objections?

andlaus:
the answer seems to be "no". thanks for the work!

PR #264, opened 2018-01-04
consistently rename "heat conduction" to "thermal conduction" and use "solid energy" laws

andlaus:
according to wikipedia the term "heat" is the energy transferred due to a temperature gradient, i.e., it only makes sense if such a gradient is present and this is not necessary for the storage term.

this means that technically the term "heat conductivity" is meaningful, but "thermal conductivity" is IMO more consistent.

this has partially already been done in opm-material and eWoms it was pretty inconsistent, so it also depends on OPM/opm-material#273.

### DIFF
--- a/ewoms/io/vtkenergymodule.hh
+++ b/ewoms/io/vtkenergymodule.hh
@@ -42,7 +42,7 @@ NEW_TYPE_TAG(VtkEnergy);
 
 // create the property tags needed for the energy module
 NEW_PROP_TAG(VtkWriteSolidInternalEnergy);
-NEW_PROP_TAG(VtkWriteHeatConductivity);
+NEW_PROP_TAG(VtkWriteThermalConductivity);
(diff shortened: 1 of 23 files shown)


andlaus:
jenkins build this with downstreams opm-material=273 please

andlaus:
jenkins is happy and I'd like to have this soon to as a precursor for the energy extension black-oil model. If nobody objects, I'll merge this in the afternoon. (I also won't mind if anybody merges it before.)

atgeirr:
I don't object. The only thing that made me pause was "SolidEnergyLaw", what exactly is that meant to cover?

andlaus:
that represents the internal energy stored by the porous medium's solid matrix. for geological applications "solid" == "rock", but calling the it "rock" looks weird in e.g. medical applications.

most commonly, the solid energy law is some constant heat capacity of solid times temperature, but in principle it can be anything.

andlaus:
Ok, let's get this done!

PR #265, opened 2018-01-05
Implement DRSDT and DRVDT

andlaus:
this is basically a simpler alternative to #255. also, it includes moving the observed maximum of the oil saturation to the problem where it IMO belongs.

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -358,6 +358,26 @@ public:
         if (!deck.hasKeyword("NOGRAV") && EWOMS_GET_PARAM(TypeTag, bool, EnableGravity))
             this->gravity_[dim - 1] = 9.80665;
 
+        // deal with DRSDT
+        maxDRsDt_ = 0.0;
+        maxDRs_ = -1.0;
+        if (deck.hasKeyword("DRSDT")) {
+            const auto& drsdtKeyword = deck.getKeyword("DRSDT");
+            maxDRsDt_ = drsdtKeyword.getRecord(0).getItem("DRSDT_MAX").getSIDouble(0);
+            size_t numDof = this->model().numGridDof();
+            lastRs_.resize(numDof, 0.0);
+        }
+
+        // deal with DRVDT
+        maxDRvDt_ = 0.0;
+        maxDRv_ = -1.0;
+        if (deck.hasKeyword("DRVDT")) {
+            const auto& drvdtKeyword = deck.getKeyword("DVSDT");
+            maxDRvDt_ = drvdtKeyword.getRecord(0).getItem("DRVDT_MAX").getSIDouble(0);
+            size_t numDof = this->model().numGridDof();
+            lastRv_.resize(numDof, 0.0);
+        }
+
         initFluidSystem_();
         updateElementDepths_();
         readRockParameters_();
@@ -495,6 +515,15 @@ public:
      */
     void beginTimeStep()
     {
+        if (!lastRs_.empty()) {
+            // DRSDT is enabled
+            maxDRs_ = maxDRsDt_*this->simulator().timeStepSize();
+        }
+
+        if (!lastRv_.empty())
+            // DRVDT is enabled
+            maxDRv_ = maxDRvDt_*this->simulator().timeStepSize();
+
         if (!GET_PROP_VALUE(TypeTag, DisableWells)) {
             wellManager_.beginTimeStep();
         }
@@ -540,6 +569,8 @@ public:
         // we no longer need the initial soluiton
         if (this->simulator().episodeIndex() == 0)
             initialFluidStates_.clear();
+
+        updateCompositionChangeLimits_();
     }
 
     /*!
@@ -985,6 +1016,8 @@ public:
 
         // release the memory of the EQUIL grid since it's no longer needed after this point
         this->simulator().gridManager().releaseEquilGrid();
+
+        updateCompositionChangeLimits_();
     }
 
     /*!
@@ -1009,6 +1042,30 @@ public:
             for (unsigned eqIdx = 0; eqIdx < numEq; ++ eqIdx)
                 rate[eqIdx] /= this->model().dofTotalVolume(globalDofIdx);
         }
+    }
+
+    /*!
+     * \brief Returns the maximum value of the gas dissolution factor at the current time
+     *        for a given degree of freedom.
+     */
+    Scalar maxGasDissolutionFactor(unsigned globalDofIdx OPM_UNUSED) const
+    {
+        if (lastRs_.empty() || maxDRs_ < 0.0)
+            return std::numeric_limits<Scalar>::max()/2;
+
+        return lastRs_[globalDofIdx] + maxDRs_;
+    }
+
+    /*!
+     * \brief Returns the maximum value of the oil vaporization factor at the current
+     *        time for a given degree of freedom.
+     */
+    Scalar maxOilVaporizationFactor(unsigned globalDofIdx OPM_UNUSED) const
+    {
+        if (lastRv_.empty() || maxDRv_ < 0.0)
+            return std::numeric_limits<Scalar>::max()/2;
+
+        return lastRv_[globalDofIdx] + maxDRv_;
     }
 
     /*!
@@ -1064,6 +1121,70 @@ private:
             const unsigned int elemIdx = elemMapper.index(element);
 
             elementCenterDepth_[elemIdx] = cellCenterDepth( element );
+        }
+    }
+
+    // update the parameters needed for DRSDT and DRVDT
+    void updateCompositionChangeLimits_()
+    {
+        // update the "last Rs" values for all elements, including the ones in the ghost
+        // and overlap regions
+        if (!lastRs_.empty()) {
+            ElementContext elemCtx(this->simulator());
+            const auto& gridManager = this->simulator().gridManager();
+            auto elemIt = gridManager.gridView().template begin</*codim=*/0>();
+            const auto& elemEndIt = gridManager.gridView().template end</*codim=*/0>();
+            for (; elemIt != elemEndIt; ++elemIt) {
+                const Element& elem = *elemIt;
+
+                elemCtx.updatePrimaryStencil(elem);
+                elemCtx.updatePrimaryIntensiveQuantities(/*timeIdx=*/0);
+
+                unsigned compressedDofIdx = elemCtx.globalSpaceIndex(/*spaceIdx=*/0, /*timeIdx=*/0);
+                const auto& iq = elemCtx.intensiveQuantities(/*spaceIdx=*/0, /*timeIdx=*/0);
+                const auto& fs = iq.fluidState();
+
+                typedef typename std::decay<decltype(fs) >::type FluidState;
+
+                lastRs_[compressedDofIdx] =
+                    Opm::BlackOil::template getRs_<FluidSystem,
+                                                   Scalar,
+                                                   FluidState>(fs,
+                                                               iq.pvtRegionIndex());
+            }
+        }
+
+        // update the "last Rv" values for all elements, including the ones in the ghost
+        // and overlap regions
+        if (!lastRv_.empty()) {
+            ElementContext elemCtx(this->simulator());
+            const auto& gridManager = this->simulator().gridManager();
+            auto elemIt = gridManager.gridView().template begin</*codim=*/0>();
+            const auto& elemEndIt = gridManager.gridView().template end</*codim=*/0>();
+            for (; elemIt != elemEndIt; ++elemIt) {
+                const Element& elem = *elemIt;
+
+                elemCtx.updatePrimaryStencil(elem);
+                elemCtx.updatePrimaryIntensiveQuantities(/*timeIdx=*/0);
+
+                unsigned compressedDofIdx = elemCtx.globalSpaceIndex(/*spaceIdx=*/0, /*timeIdx=*/0);
+                const auto& iq = elemCtx.intensiveQuantities(/*spaceIdx=*/0, /*timeIdx=*/0);
+                const auto& fs = iq.fluidState();
+
+                typedef typename std::decay<decltype(fs) >::type FluidState;
+
+                lastRv_[compressedDofIdx] =
+                    Opm::BlackOil::template getRv_<FluidSystem,
+                                                   Scalar,
+                                                   FluidState>(fs,
+                                                               iq.pvtRegionIndex());
+            }
+        }
+
+        if (!lastRs_.empty() || !lastRv_.empty()) {
+            // we need to invalidate the intensive quantities cache here because the Rs
+            // values of the intensive quantities are potentially different now.
+            this->model().invalidateIntensiveQuantitiesCache(/*timeIdx=*/0);
         }
     }
 
@@ -1617,6 +1738,13 @@ private:
     std::vector<Scalar> polymerConcentration_;
     std::vector<Scalar> solventSaturation_;
 
+    std::vector<Scalar> lastRs_;
+    Scalar maxDRsDt_;
+    Scalar maxDRs_;
+
+    std::vector<Scalar> lastRv_;
+    Scalar maxDRvDt_;
+    Scalar maxDRv_;
 
     EclWellManager<TypeTag> wellManager_;
 
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -92,6 +92,7 @@
 
 #include <vector>
 #include <string>
+#include <algorithm>
 
 namespace Ewoms {
 template <class TypeTag>
@@ -385,6 +386,12 @@ public:
             maxDRsDt_ = drsdtKeyword.getRecord(0).getItem("DRSDT_MAX").getSIDouble(0);
             size_t numDof = this->model().numGridDof();
             lastRs_.resize(numDof, 0.0);
+
+            std::string drsdtFlag =
+                drsdtKeyword.getRecord(0).getItem("Option").getTrimmedString(0);
+            std::transform(drsdtFlag.begin(), drsdtFlag.end(), drsdtFlag.begin(), ::toupper);
+
+            dRsDtOnlyFreeGas_ = (drsdtFlag == "FREE");
         }
 
         // deal with DRVDT
@@ -1198,11 +1205,14 @@ private:
 
                 typedef typename std::decay<decltype(fs) >::type FluidState;
 
-                lastRs_[compressedDofIdx] =
-                    Opm::BlackOil::template getRs_<FluidSystem,
-                                                   Scalar,
-                                                   FluidState>(fs,
-                                                               iq.pvtRegionIndex());
+                if (!dRsDtOnlyFreeGas_ || fs.saturation(gasPhaseIdx) > 1e-5)
+                    lastRs_[compressedDofIdx] =
+                        Opm::BlackOil::template getRs_<FluidSystem,
+                                                       Scalar,
+                                                       FluidState>(fs,
+                                                                   iq.pvtRegionIndex());
+                else
+                    lastRs_[compressedDofIdx] = 1e4;
             }
         }
 
@@ -1817,6 +1827,7 @@ private:
     std::vector<Scalar> solventSaturation_;
 
     bool drsdtActive_; // if no, VAPPARS *might* be active
+    bool dRsDtOnlyFreeGas_; // apply the DRSDT rate limit only to cells that exhibit free gas
     std::vector<Scalar> lastRs_;
     Scalar maxDRsDt_;
     Scalar maxDRs_;
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -1211,7 +1211,7 @@ private:
                                                        FluidState>(fs,
                                                                    iq.pvtRegionIndex());
                 else
-                    lastRs_[compressedDofIdx] = 1e4;
+                    lastRs_[compressedDofIdx] = std::numeric_limits<Scalar>::infinity();
             }
         }
 
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -358,10 +358,16 @@ public:
         if (!deck.hasKeyword("NOGRAV") && EWOMS_GET_PARAM(TypeTag, bool, EnableGravity))
             this->gravity_[dim - 1] = 9.80665;
 
+        // this is actually not fully correct: the latest occurence of VAPPARS and DRSDT
+        // or DRVDT up to the current time step in the schedule section counts, presence
+        // of VAPPARS alone is not sufficient to disable DR[SV]DT. TODO: implment support
+        // for this in opm-parser's Schedule object"
+        bool hasVappars = deck.hasKeyword("VAPPARS");
+
         // deal with DRSDT
         maxDRsDt_ = 0.0;
         maxDRs_ = -1.0;
-        if (deck.hasKeyword("DRSDT")) {
+        if (!hasVappars && deck.hasKeyword("DRSDT")) {
             const auto& drsdtKeyword = deck.getKeyword("DRSDT");
             maxDRsDt_ = drsdtKeyword.getRecord(0).getItem("DRSDT_MAX").getSIDouble(0);
             size_t numDof = this->model().numGridDof();
@@ -371,7 +377,7 @@ public:
         // deal with DRVDT
         maxDRvDt_ = 0.0;
         maxDRv_ = -1.0;
-        if (deck.hasKeyword("DRVDT")) {
+        if (!hasVappars && deck.hasKeyword("DRVDT")) {
             const auto& drvdtKeyword = deck.getKeyword("DVSDT");
             maxDRvDt_ = drvdtKeyword.getRecord(0).getItem("DRVDT_MAX").getSIDouble(0);
             size_t numDof = this->model().numGridDof();
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -523,7 +523,7 @@ public:
         }
 
         bool doInvalidate = updateHysteresis_();
-        doInvalidate = updateMaxOilSaturation_();
+        doInvalidate = doInvalidate || updateMaxOilSaturation_();
 
         if (GET_PROP_VALUE(TypeTag, EnablePolymer))
             updateMaxPolymerAdsorption_();
@@ -542,12 +542,11 @@ public:
      */
     void beginTimeStep()
     {
-        if (!lastRs_.empty()) {
+        if (drsdtActive_)
             // DRSDT is enabled
             maxDRs_ = maxDRsDt_*this->simulator().timeStepSize();
-        }
 
-        if (!lastRv_.empty())
+        if (drvdtActive_)
             // DRVDT is enabled
             maxDRv_ = maxDRvDt_*this->simulator().timeStepSize();
 
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -1204,7 +1204,7 @@ private:
 
                 typedef typename std::decay<decltype(fs) >::type FluidState;
 
-                if (!dRsDtOnlyFreeGas_ || fs.saturation(gasPhaseIdx) > 1e-5)
+                if (!dRsDtOnlyFreeGas_ || fs.saturation(gasPhaseIdx) > freeGasMinSaturation_)
                     lastRs_[compressedDofIdx] =
                         Opm::BlackOil::template getRs_<FluidSystem,
                                                        Scalar,
@@ -1835,6 +1835,7 @@ private:
     std::vector<Scalar> lastRv_;
     Scalar maxDRvDt_;
     Scalar maxDRv_;
+    constexpr static Scalar freeGasMinSaturation_ = 1e-7;
 
     bool vapparsActive_; // if no, DRSDT and/or DRVDT *might* be active
     std::vector<Scalar> maxOilSaturation_;
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -362,12 +362,25 @@ public:
         // or DRVDT up to the current time step in the schedule section counts, presence
         // of VAPPARS alone is not sufficient to disable DR[SV]DT. TODO: implment support
         // for this in opm-parser's Schedule object"
-        bool hasVappars = deck.hasKeyword("VAPPARS");
+        drsdtActive_ = false;
+        drvdtActive_ = false;
+        vapparsActive_ = false;
+
+        if (deck.hasKeyword("VAPPARS")) {
+            vapparsActive_ = true;
+
+            size_t numDof = this->model().numGridDof();
+            maxOilSaturation_.resize(numDof, 0.0);
+
+            // TODO: update the PVT objects. this is only required if VAPPARS becomes a
+            // fully dynamic keyword.
+        }
 
         // deal with DRSDT
         maxDRsDt_ = 0.0;
         maxDRs_ = -1.0;
-        if (!hasVappars && deck.hasKeyword("DRSDT")) {
+        if (!vapparsActive_ && deck.hasKeyword("DRSDT")) {
+            drsdtActive_ = !vapparsActive_;
             const auto& drsdtKeyword = deck.getKeyword("DRSDT");
             maxDRsDt_ = drsdtKeyword.getRecord(0).getItem("DRSDT_MAX").getSIDouble(0);
             size_t numDof = this->model().numGridDof();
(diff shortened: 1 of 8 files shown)


ebos/eclproblem.hh
```diff
@@ -1051,6 +1150,98 @@ private:
...
+                                                   FluidState>(fs,
+                                                               iq.pvtRegionIndex());
+            }
+        }
+    }
+
+    bool updateMaxOilSaturation_()
+    {
+        // we use VAPPARS
+        if (vapparsActive_) {
```
atgeirr: This loop and those above could be combined into one, since the iteration is rather heavy. The necessary if-s inside would be quite predictable and not cost performance.

ebos/eclproblem.hh
```diff
@@ -1051,6 +1150,98 @@ private:
...
+
+                maxOilSaturation_[compressedDofIdx] = std::max(maxOilSaturation_[compressedDofIdx], So);
+            }
+
+            // we need to invalidate the intensive quantities cache here because the
+            // derivatives of Rs and Rv will most likely have changed
+            return true;
+        }
+
+        return false;
```
atgeirr: The return value here is unnecessary, it is just the vapparsActive_ flag. I think the code calling this would be better without.

ewoms/models/blackoil/blackoilintensivequantities.hh
```diff
@@ -186,51 +186,56 @@ public:
...
                 fluidState_.setRv(0.0);
         }
         else if (priVars.primaryVarsMeaning() == PrimaryVariables::Sw_po_Rs) {
             // if the switching variable is the mole fraction of the gas component in the
+            Scalar RsMax = elemCtx.problem().maxGasDissolutionFactor(globalSpaceIdx);
+
             // oil phase, we can directly set the composition of the oil phase
             const auto& Rs = priVars.makeEvaluation(Indices::compositionSwitchIdx, timeIdx);
-            fluidState_.setRs(Rs);
+            fluidState_.setRs(Opm::min(RsMax, Rs));
```
atgeirr: Why is not the dRsDtOnlyFreeGas_ flag checked here? If true, this should be skipped. Better to check that here with a bool than to depend on arbitrary large (perhaps large enough) values being written to lastRs_.

ewoms/models/blackoil/blackoilintensivequantities.hh
```diff
@@ -244,13 +249,14 @@ public:
             if (FluidSystem::enableDissolvedGas()) {
```
atgeirr: In this else branch (rv is a variable), should the Rv update be limited?

ewoms/models/blackoil/blackoilnewtonmethod.hh
```diff
@@ -158,25 +158,27 @@ protected:
             Scalar delta = update[eqIdx];
 
             // limit changes in water saturation to 20%
-            if (eqIdx == Indices::waterSaturationIdx
```
atgeirr: I have trouble seeing what was actually changed in this section?

ewoms/models/blackoil/blackoilprimaryvariables.hh
```diff
@@ -380,16 +380,18 @@ public:
...
                                                                                    po,
                                                                                    So2,
                                                                                    SoMax);
+
                 setPrimaryVarsMeaning(Sw_po_Rs);
                 if (compositionSwitchEnabled)
-                    (*this)[Indices::compositionSwitchIdx] = RsSat;
-
+                    (*this)[Indices::compositionSwitchIdx] =
+                        std::min(RsMax, RsSat);
```
atgeirr: Open question: we go from having free gas to not having free gas, should we then act depending on the dRsDtOnlyFreeGas_ flag? This code ignores it...

ewoms/models/blackoil/blackoilprimaryvariables.hh
```diff
@@ -450,11 +451,17 @@ public:
             Scalar T = asImp_().temperature_();
             Scalar po = (*this)[Indices::pressureSwitchIdx];
             Scalar So = 1.0 - Sw - solventSaturation();
-            Scalar SoMax = std::max(So, problem.model().maxOilSaturation(globalDofIdx));
+            Scalar SoMax = std::max(So, problem.maxOilSaturation(globalDofIdx));
+            Scalar RsMax = problem.maxGasDissolutionFactor(globalDofIdx);
```
atgeirr: Better: make RsMax min of problem.maxGas...() and RsSat (must also switch order then). Further down then if (Rs > RsMax) is a more logical way to check if gas phase appears (line 464).

ewoms/models/blackoil/blackoilprimaryvariables.hh
```diff
@@ -501,16 +508,18 @@ public:
...
-            Scalar SoMax = problem.model().maxOilSaturation(globalDofIdx);
+            Scalar SoMax = problem.maxOilSaturation(globalDofIdx);
+            Scalar RvMax = problem.maxOilVaporizationFactor(globalDofIdx);
             Scalar RvSat =
                 FluidSystem::gasPvt().saturatedOilVaporizationFactor(pvtRegionIdx_,
                                                                      T,
                                                                      pg,
                                                                      /*So=*/Scalar(0.0),
                                                                      SoMax);
+            RvSat = std::min(RvMax, RvSat);
```
atgeirr: Should use RvMax instead of RvSat here (and change the if on line 522 to use it as well), as suggested for Rs above.

ewoms/models/blackoil/blackoilproblem.hh
```diff
@@ -59,6 +59,33 @@ public:
         : ParentType(simulator)
     {}
 
+    /*!
+     * \brief Returns the maximum value of the gas dissolution factor at the current time
+     *        for a given degree of freedom.
+     *
+     * This is required for the DRSDT keyword.
+     */
+    Scalar maxGasDissolutionFactor(unsigned globalDofIdx OPM_UNUSED) const
```
atgeirr: I do not really understand the relationship between blackoilproblem and eclproblem well enough to see why these functions are needed here?

ebos/eclproblem.hh
```diff
@@ -483,10 +529,8 @@ public:
             simulator.setTimeStepSize(dt);
         }
 
-        if (updateHysteresis_())
-            this->model().invalidateIntensiveQuantitiesCache(/*timeIdx=*/0);
-
-        this->model().updateMaxOilSaturations();
+        bool doInvalidate = updateHysteresis_();
+        doInvalidate = updateMaxOilSaturation_();
```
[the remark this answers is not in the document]
andlaus: for your first comment: yes, you're right. thanks for the catch, I've pushed a fixup.

regarding your second comment: it depends on what eclipse does, I assumed that it updates the limits at the end of each substep. (the difference should not matter too much except if the Rs value goes down in a substep and up again in later ones.)

ebos/eclproblem.hh
```diff
@@ -495,13 +539,25 @@ public:
...
+        if (doInvalidate)
+            this->model().invalidateIntensiveQuantitiesCache(/*timeIdx=*/0);
     }
 
     /*!
      * \brief Called by the simulator before each time integration.
      */
     void beginTimeStep()
     {
+        if (!lastRs_.empty()) {
```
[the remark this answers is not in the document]
andlaus: good question. maybe the VAPPARS stuff. anyway, I've pushed a fixup, thanks.

ebos/eclproblem.hh
```diff
@@ -1002,6 +1062,45 @@ public:
         }
     }
 
+    /*!
+     * \brief Returns the maximum value of the gas dissolution factor at the current time
+     *        for a given degree of freedom.
+     */
+    Scalar maxGasDissolutionFactor(unsigned globalDofIdx) const
+    {
+        if (!drsdtActive_ || maxDRs_ < 0.0)
```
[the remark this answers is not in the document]
andlaus: correct. changed, thanks.

ebos/eclproblem.hh
```diff
@@ -1051,6 +1150,98 @@ private:
...
+                elemCtx.updatePrimaryStencil(elem);
+                elemCtx.updatePrimaryIntensiveQuantities(/*timeIdx=*/0);
+
+                unsigned compressedDofIdx = elemCtx.globalSpaceIndex(/*spaceIdx=*/0, /*timeIdx=*/0);
+                const auto& iq = elemCtx.intensiveQuantities(/*spaceIdx=*/0, /*timeIdx=*/0);
+                const auto& fs = iq.fluidState();
+
+                typedef typename std::decay<decltype(fs) >::type FluidState;
+
+                if (!dRsDtOnlyFreeGas_ || fs.saturation(gasPhaseIdx) > 1e-5)
```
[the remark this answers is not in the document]
andlaus: this is needed for the FREE flag for DRSDT: if phase is present, the increase of Rs is unlimited in the presence of that flag. I think that it is cleaner to do this here and not bother in the intensive quantities. (if one can say "clean" in the context of DRSDT.)

Note that DRVDT does not exhibit an analogous atrocity.

ebos/eclproblem.hh
andlaus: it is not really time critical and will usually be cached, though. anyway, I'll merge this with the DR[SV]DT update code and move the ifs into the loop if you insist, but this will make make the code more complicated and not much faster considering that only one of VAPPARS or DR[SV]DT can be active at the same time.

ebos/eclproblem.hh
andlaus: > The return value here is unnecessary, it is just the vapparsActive_ flag

true. it keeps the symmetry with `updateHysteresis_()`, though.

ewoms/models/blackoil/blackoilintensivequantities.hh
andlaus: less special cases and symmetry with DRVDT? (the real reason is that the "FREE" flag was bolted-on later.)

ewoms/models/blackoil/blackoilintensivequantities.hh
andlaus: no, Rv is dealt with by the `if (FluidSystem::enableVaporizedOil())` branch above. for the else branch: Rs is always zero in the dead oil case, right?

ewoms/models/blackoil/blackoilnewtonmethod.hh
andlaus: join the club ;)

IIRC that's an artefact of when I wanted to limit the composition update. In the end it turned out to be unnecessary. I've removed the change, thanks for the catch.

ewoms/models/blackoil/blackoilprimaryvariables.hh
andlaus: true. respecting the FREE flag here will likely lead to serious convergence issues, though.

@totto82: any idea how to check this in the reference simulator?

ewoms/models/blackoil/blackoilprimaryvariables.hh
andlaus: good idea, changed.

ewoms/models/blackoil/blackoilprimaryvariables.hh
andlaus: agreed, changed.

ewoms/models/blackoil/blackoilproblem.hh
andlaus: basically `BlackOilProblem` is the base class for the black-oil model. in particular it is supposed to be independent of anything that needs ECL stuff. In practice it means that it does not use opm-parser data structures so that -- in principle -- you can a non-ECL black-oil simulator. EclProblem is the implementation of BlackOilProblem which uses ECL weirdness to satisfy the API expected by the model. 

These functions are needed here to avoid all problems which derive from the BlackOilProblem and don't care about this -- ehem, ReservoirProblem -- having to provide them.

totto82:
This changes behavior for Model 2 where DRSDT is not present. Is this expected?

ebos/eclproblem.hh
```diff
@@ -363,6 +364,51 @@ public:
...
+        // for this in opm-parser's Schedule object"
+        drsdtActive_ = false;
+        drvdtActive_ = false;
+        vapparsActive_ = false;
+
+        if (deck.hasKeyword("VAPPARS")) {
+            vapparsActive_ = true;
+
+            size_t numDof = this->model().numGridDof();
+            maxOilSaturation_.resize(numDof, 1.0);
```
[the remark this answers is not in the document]
atgeirr: I won't, but you have to defend your hack! Primarily by explaining it...

ewoms/models/blackoil/blackoilprimaryvariables.hh
```diff
@@ -450,11 +451,17 @@ public:
...
+                FluidSystem::oilPvt().saturatedGasDissolutionFactor(pvtRegionIdx_,
+                                                                    T,
+                                                                    po,
+                                                                    So,
+                                                                    SoMax);
+
             Scalar Rs = (*this)[Indices::compositionSwitchIdx];
-            if (Rs > RsSat*(1.0 + eps)) {
+            if (Rs > std::min(RsMax, RsSat)) {
                 // the gas phase appears, i.e., switch the primary variables to { Sw, po,
```
totto82: This change behavior for the non DRSDT case. Adding an `eps` here after a change of variable stabilized the convergence for some of the Model 2 realizations. I am open to change this, but I don't think it should happen in this PR, and we should test this throughly.  
`std::min(RsMax, RsSat)*(1.0 + eps) `should do the trick. (Also below for Rv)

ebos/eclproblem.hh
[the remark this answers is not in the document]
atgeirr: To the second comment: what I meant was that these things seem to be updated only once every episode (== report timestep requested in schedule) and not every timestep (== ministep or substep), from you comment it looks like you agree it should be updated every timestep. So did I misunderstand what an episode is?

ebos/eclproblem.hh
atgeirr: Ok for now at least. We could always profile it afterwords. It's true that as only one can be active at a time, reducing the impact.

ebos/eclproblem.hh
atgeirr: ... which reminds me that I do not really like that pattern very much, in general I prefer having query methods being const and not changing anything, and update-like method not returning anything (I think Ruby has a very useful convention, with methods ending in ! changing the object and methods ending in ? being queries). But that is not the task of this PR, so I'll let that pass for now...

ewoms/models/blackoil/blackoilproblem.hh
atgeirr: Thanks for the explanation. But these methods are not virtual, and I do not see the B-N trick used either, so how do you ensure you get the most-derived version of the method?

ewoms/models/blackoil/blackoilprimaryvariables.hh
atgeirr: I agree.

ebos/eclproblem.hh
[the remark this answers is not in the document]
atgeirr: I disagree that this is cleaner, because it encodes the binary decision by using some "hopefully large enough" value for lastRs_ instead of preserving this decision as being controlled by a flag. Essentially you are lying through lastRs_ to make things work out right, and I don't think that is a good idea. Are there other reasons for not doing it in the intensive quantities?

ewoms/models/blackoil/blackoilintensivequantities.hh
atgeirr: > no, Rv is dealt with by the `if (FluidSystem::enableVaporizedOil())` branch above

I don't think that is true, in this branch there is no such if?!?

ewoms/models/blackoil/blackoilintensivequantities.hh
atgeirr: > the real reason is that the "FREE" flag was bolted-on later

I can understand that...

But the code would be much better if we did not lie about the states to get that boolean check "for free" here.

andlaus: another reason is that if this is moved to the IQs, the "free" flag would need to be exported by the problem and thus the method must also be provided by the base problem. I think it is simpler as-is, but if you *really* want to do it in the intensive quantities, I'll change it...

ebos/eclproblem.hh
andlaus: yes, if the update method changes something unconditionally, this I will gladly go down this path. the problem is how to do "accumulated" mop-ups if you have several things updates and only sometimes they change something. (here, "mob up" means invalidating the IQ cache. if this is done unconditionally after each update, performance will suffer.)

ebos/eclproblem.hh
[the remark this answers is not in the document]
andlaus: but lastRs_ is not visible outside of the problem object, only the maximum Rs is, and that one does not lie?!

> Are there other reasons for not doing it in the intensive quantities?

see below: keeping the API which needs to be provided by black oil problems a bit simpler.

ebos/eclproblem.hh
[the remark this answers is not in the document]
andlaus: the hack is: if So < SoMax, VAPPARS does not get applied.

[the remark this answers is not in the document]
andlaus: okay, I messed this up; the condition is reversed and you're right. I've added a fixup. oops. thanks.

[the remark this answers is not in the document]
atgeirr: Ok, got it. But this is another case of lying to other parts of the code is it not? What I mean is that somewhere else, you check So < SoMax instead of the two-step procedure of first checking if vappars is active and then checking if So < SoMax, right? So this hack saves you that bit of logic, but at the cost of having the "convention" that with inactive vappars then SoMax is 1. I fear that is bound to backfire at some point. For example: what if we are asked to output SoMax, but without activating vappars? It would be a valid engineering approach to look at that variable in order to decide to use vappars or not.

andlaus:
> This changes behavior for Model 2 where DRSDT is not present. Is this expected?

no. @atgeirr pointed out an issue with VAPPARS which is now fixed. can you re-check?

ewoms/models/blackoil/blackoilproblem.hh
andlaus: the IQs just call the functions of the problem object which they expect. getting the "most derived" class is done implicitly by the property system because you set the `Problem` property to the actual problem class. (IMO B-N just complicates things a lot for no real gain in practice.)

ewoms/models/blackoil/blackoilintensivequantities.hh
atgeirr: It should still be limited, though? If there is a limit, it would prevent Rv from rising too rapidly, resulting in a separate oil phase coming into existence, and I would expect that to be seen here, right?

ewoms/models/blackoil/blackoilprimaryvariables.hh
andlaus: okay, added back. this was basically an artefact of development. thanks for catching it.

ewoms/models/blackoil/blackoilintensivequantities.hh
andlaus: > It should still be limited, though?

if it is a primary variable, no: the limitation is done when adapting the PVs. This is a bit tricky, I know...

andlaus: (note that this applies to the situation before this PR as well, just with always the thermodynamic maximum.)

ebos/eclproblem.hh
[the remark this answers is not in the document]
andlaus: your understanding of the "episode" concept is correct. the question rather is what the reference simulator does. I vaguely remember that there was some weirdness...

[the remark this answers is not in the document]
atgeirr: > your understanding of the "episode" concept is correct

If so, then my comment stands. I find it hard to believe that Eclipse would only update this per report step, but who knows. On its own, the code here looks suspicious (only updating per report step), do we agree on that?

ebos/eclproblem.hh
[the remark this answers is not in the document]
atgeirr: > but lastRs_ is not visible outside of the problem object

For now... It's likely to be broken in the future by anyone who did not read this discussion though!

> only the maximum Rs is, and that one does not lie?!

Will it not be "something hopefully large enough" rather than infinity?

> see below: keeping the API which needs to be provided by black oil problems a bit simpler.

On that: I get now that there is a separation between blackoilproblem and eclproblem, but are there any other blackoilproblem variants? If the answer is no, then for sure this is too early to say  what the interface should be for the "common blackoil" API, as we won't really know until we have such a case.

ebos/eclproblem.hh
atgeirr: I agree that we should delay the cache recalculation until after all the updates. I assume that the invalidation itself is also a somewhat big job (setting a bool for each element?) since otherwise the obvious solution would be for each update function to invalidate the cache if it did anything to ruin it. Is that why you don't want to do that invalidation there? If so, perhaps it would be better with a more coarse-grained cache control (a bool for all of the cells rather than one for each for example).

ewoms/models/blackoil/blackoilproblem.hh
atgeirr: Fine, I agree that the B-N trick is not so nice. I guess this is the strong point of the property system: avoiding B-N yet allowing inheritance. But I think there is still a general, potential problem with this pattern: inside any given problem class, calls to problem class members cannot go to the most derived one, but will always just go to the same class as the caller. Agree?

ewoms/models/blackoil/blackoilintensivequantities.hh
atgeirr: I think I do want this. It is always better to depend on the explicit information, when available (in this case the boolean flag) rather than implicit assumptions and "large enough" values. I'm not saying that this is a rule without exceptions, but it seems to me that the cost here (add a method to the blackoilproblem interface class) is small.

ebos/eclproblem.hh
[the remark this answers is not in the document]
andlaus: > I find it hard to believe that Eclipse would only update this per report step, but who knows.

join the club. IIRC this was the issue with the hysteresis and max oil saturation updates, so I think it is likely that it also applies to the maximum Rs/Rv. we should check this (and re-check the hystersis and max oil stuff), though. @totto82: can you do this?

> On its own, the code here looks suspicious (only updating per report step), do we agree on that?

yes, from a physical POV it is wrong. physics does not define truth here, though.

ebos/eclproblem.hh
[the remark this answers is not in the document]
andlaus: > Will it not be "something hopefully large enough" rather than infinity?

yes, but that's mainly because I don't trust processors with infinity enough. I'm happy to change to infinity it you do.

> are there any other blackoilproblem variants?

yes, the reservoir unit tests. these also work for non-blackoil models like NCP, PVS and flash!

ebos/eclproblem.hh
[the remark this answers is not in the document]
atgeirr: Ok, then we are finally on one page here? I think your original comment

> I assumed that it updates the limits at the end of each substep

confused things a bit, since you seem to assume the opposite in the code...

And yes, let's get this checked before saying it's done...

ebos/eclproblem.hh
andlaus: > I assume that the invalidation itself is also a somewhat big job (setting a bool for each element?

yes, it is somewhat expensive. that said, it is really fast in comparison to having to recalculate all IQs for the next part of the update.

> Is that why you don't want to do that invalidation there?

even though the updates do *not* change the values of the IQs, they affect the derivatives.

> If so, perhaps it would be better with a more coarse-grained cache control (a bool for all of the cells rather than one for each for example).

this would kill performance because during linearization: the IQs would need to be calculated for each element for which a given DOF is in the stencil instead of once. for 3D problems using cartesian grids these would mean 27 updates instead of 1.

one could change the linearization code to precompute the IQs, but that is a somewhat larger endeavour and I'm not sure if it is worth the effort.

ewoms/models/blackoil/blackoilproblem.hh
andlaus: > Agree?

yes. Because of the property system, all users will use the same class though.

ebos/eclproblem.hh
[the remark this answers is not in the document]
atgeirr: > yes, but that's mainly because I don't trust processors with infinity enough. I'm happy to change to infinity it you do.

I think your distrust is fine... I'd rather see the necessary flags in the API to avoid the floating-point comparisons completely (when not needed).

> yes, the reservoir unit tests. these also work for non-blackoil models like NCP, PVS and flash!
 Write

Good, then it is not just a theoretical issue!

ewoms/models/blackoil/blackoilproblem.hh
atgeirr: > yes. Because of the property system, all users will use the same class though

I did not understand how that solves the problem.

If BlackoilProblem::foo() calls bar() or this->bar(), that will *always* be a call to BlackoilProblem::bar() and not EclProblem::bar(), so the ability to overload methods really only applies to methods that are *not called at all* from BlackoilProblem methods.

ebos/eclproblem.hh
atgeirr: > this would kill performance because during linearization: the IQs would need to be calculated for each element for which a given DOF is in the stencil instead of once. for 3D problems using cartesian grids these would mean 27 updates instead of 1.
> 
> one could change the linearization code to precompute the IQs, but that is a somewhat larger endeavour and I'm not sure if it is worth the effort.

I have somehow assumed that this was how it was done! I was wrong then. It does seem to me to be a very logical thing to do, and unless I am wrong we would not have to change any of the linearization code. Would not the only change be to add the pre-computing loop?

andlaus: the basic idea of the DOF-specific IQ cache is that you don't need the update for DOFs which you usually don't access it (e.g. the overlap), and if you need them in some cases, the IQ will still be available. maybe this a bit over-engineered.

> It does seem to me to be a very logical thing to do, and unless I am wrong we would not have to change any of the linearization code. Would not the only change be to add the pre-computing loop?

yes, true. The problem might be slightly worse performance in the parallel case because a-priori it is not clear if the overlap IQs will be needed or not.

ewoms/models/blackoil/blackoilintensivequantities.hh
andlaus: okay, i tried to do this. it also requires modifications to the update proceedure, and the code gets quite a bit more spaghetti that way. do you still want it?

atgeirr: Can you email you the sketched code? I'll take a look.

andlaus:
I've rebased the PR and changed the maximum Rs value for the FREE flag of DRSDT if no gas is present to infinity. it seems to work.

atgeirr:
After lengthy offline discussions about many interesting aspects of ewoms it is time to wind down this. With the latest changes I think the code is acceptable as is, if not ideal from a few perspectives, that might be adressed in a future PR.

What remains is testing. I will start jenkins now, also @totto82 did you rerun model 2?

atgeirr:

andlaus:
testing it on some model were DR[SV]DT has a non-trivial value and one with the FREE flag of DRSDT set would also be beneficial IMO.

atgeirr:
> testing it on some model were DR[SV]DT has a non-trivial value and one with the FREE flag of DRSDT set would also be beneficial IMO.

Yes I hope @hnil can do that tomorrow, before merging this.

totto82:
I have started the testing.

andlaus:
the last patch broke the build because I forgot to make the new constexpr variable static. Fixed, sorry about that.

(take home message: always compile even if the changes are trivial.)

andlaus:

totto82:
> the last patch broke the build because I forgot to make the new constexpr variable static. Fixed, sorry about that.

I have restarted the tests with the fix.

atgeirr:
Could you address this, @andlaus?
```
/var/lib/jenkins/workspace/ewoms-PR-builder/deps/opm-simulators/opm/autodiff/SimulatorFullyImplicitBlackoilEbos.hpp:566:36: error: 'Ewoms::Simulator<Ewoms::Properties::TTag::EclFlowProblem>::Model {aka class Ewoms::BlackOilModel<Ewoms::Properties::TTag::EclFlowProblem>}' has no member named 'setMaxOilSaturation'
             ebosSimulator_.model().setMaxOilSaturation(somax[cellIdx], cellIdx);
```

andlaus:
oops. I forgot to open a PR for the opm-simulators mop-up branch. let's try again...

andlaus:
jenkins build this with downstreams opm-simulators=1381 please

andlaus:
@totto82: this causes minor conflicts with #262 due to moving the maximum oil saturation to the problem. do you want #262 or this one to be merged first? in the former case, I'll deal with the merge conflicts, in the latter you'll have to do this...

totto82:
It isn't a big deal anyhow. But if you think #262 is ready, maybe you could merge that first, since this PR still miss some testing.

andlaus:
the jenkins failures were caused by the fact that the maximum oil saturation was handled slightly differently than before if VAPPARS was not present. this should not really matter, but I've fixed the code to replicate the previous behaviour exactly.

andlaus:
jenkins build this with downstreams opm-simulators=1381 please

totto82:
No change in performance and convergence for model2. Some effect on convergence for norne, but results stays the same. 

diff PR vs Master on Norne
```
< Total time (seconds):         473.457
< Solver time (seconds):        461.516
<  Assembly time (seconds):     222.565 (Failed: 6.4086; 2.87942%)
<  Linear solve time (seconds): 220.052 (Failed: 7.21705; 3.27971%)
<  Update time (seconds):       9.98343 (Failed: 0.316377; 3.16902%)
<  Output write time (seconds): 13.8214
< Overall Well Iterations:      979 (Failed: 17; 1.73647%)
< Overall Linearizations:       1916 (Failed: 55; 2.87056%)
< Overall Newton Iterations:    1577 (Failed: 55; 3.48763%)
< Overall Linear Iterations:    24182 (Failed: 785; 3.24622%)
---
> Total time (seconds):         481.964
> Solver time (seconds):        469.685
>  Assembly time (seconds):     228.467 (Failed: 7.90081; 3.45819%)
>  Linear solve time (seconds): 222.124 (Failed: 8.32269; 3.74687%)
>  Update time (seconds):       10.0268 (Failed: 0.380839; 3.79821%)
>  Output write time (seconds): 14.3242
> Overall Well Iterations:      920 (Failed: 20; 2.17391%)
> Overall Linearizations:       1932 (Failed: 66; 3.41615%)
> Overall Newton Iterations:    1593 (Failed: 66; 4.14313%)
> Overall Linear Iterations:    24249 (Failed: 884; 3.64551%)
57017c57023
< Info              1177
---
> Info              1178
57021c57027
< Problems          5
---
> Problems          6

```

atgeirr:
> Some effect on convergence for norne

Isn't that a bit weird, and therefore worrying? I would expect no change at all from this PR. At best it subtly changes the order of some operations when VAPPARS is involved, at worst we have a subtle bug. Or am I worrying too much about this?

andlaus:
I've fixed the merge conflicts with #262. A nice side effect is that opm-simulators now does not need a mop up PR anymore.

andlaus:

andlaus:
I guess you're worrying too much: In the past we've seen larger changes of the performance for even smaller code changes (like e.g. changing the order of factors in a multiplication)

atgeirr:
> I guess you're worrying too much: In the past we've seen larger changes of the performance for even smaller code changes (like e.g. changing the order of factors in a multiplication)

Yes, we know such things can have an effect. But where is the change in this PR that makes things different? Anyway I'll not follow this to the end of the earth (or Santiago de Compostela, as they would in medieval times).

andlaus:
> But where is the change in this PR that makes things different? 

it could be that the ordering of updated is different and the compiler might thus optimize slightly differently. Just an unscientific theory...

> Anyway I'll not follow this to the end of the earth (or Santiago de Compostela, as they would in medieval times).

@totto82: did you observe positive effects for some other cases? if yes, the Norne behaviour can almost certainly be attributed to noise.

totto82:
All the other cases are unaffected. So it is related to VAPPARS and DRSDT. We have seen smaller changes than this affect the convergence earlier. As I said results are identical. This change gives one less convergence problem. Lets just be happy and move on.

atgeirr:
> All the other cases are unaffected. So it is related to VAPPARS and DRSDT.

Thanks for that verification. Merging!

PR #266, opened 2018-01-05
Packaging: add prereqs file and remove opm-core dependency

akva2:


andlaus:
thanks

debian/control
```diff
@@ -5,7 +5,7 @@ Maintainer: Arne Morten Kvarving <arne.morten.kvarving@sintef.no>
 Build-Depends: debhelper (>= 9), cmake,
  gfortran, libtool, mpi-default-bin, mpi-default-dev, pkg-config,
  libdune-grid-dev (>= 2.2.0), libdune-istl-dev (>= 2.2.0),
- libdune-localfunctions-dev (>= 2.2.0), libopm-core-dev, libopm-material-dev,
+ libdune-localfunctions-dev (>= 2.2.0), libopm-material-dev,
```
andlaus: this should probably be changed to >= 2.4.0, right? I suppose it does not really matter, but still...

akva2: yes. noted, will update it next time files are touched.

PR #267, opened 2018-01-08
fixed: build without opm-core

akva2:
currently the ewoms output code relies on the outdated HAVE_ERT define.
this is only set through the opm-core dependency. if you build without,
you cannot build ewoms. this at least fixes the build in this situation, although we do not get the output code enabled in that case.

### DIFF
--- a/ebos/eclsummarywriter.hh
+++ b/ebos/eclsummarywriter.hh
@@ -71,7 +71,9 @@ class EclSummaryWriter
     typedef typename GET_PROP_TYPE(TypeTag, FluidSystem) FluidSystem;
 
     typedef Ewoms::EclWellManager<TypeTag> WellManager;
+#ifdef HAVE_ERT
     typedef Ewoms::ErtSummary<TypeTag> ErtSummary;
+#endif
 
 
     struct ErtWellInfo {
@@ -129,6 +131,7 @@ public:
             reportIdx += 1;
         }
 
+#ifdef HAVE_ERT
         ErtSummaryTimeStep<TypeTag> ertSumTimeStep(ertSummary_, t, reportIdx);
 
         typedef EclDeckUnits<TypeTag> DeckUnits;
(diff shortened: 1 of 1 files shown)


andlaus:
looks okay, but ebos (and thus everything which needs libecl) should not get compiled if opm-parser (and thus libecl) is not available? cf

https://github.com/OPM/ewoms/blob/master/CMakeLists.txt#L130

akva2:
again. opm-parser *is* available. opm-core is not. the code in ewoms relies on HAVE_ERT, something that is NOT set by opm-parser, nor by libecl. it is only set by opm-core (for backwards compat reasons, it is an outdated define from the old ert days).

*some* parts of the code is protected by HAVE_ERT, but not all. by removing opm-core this issue surfaces. but the real issue is the use of HAVE_ERT.

andlaus:
Hm, looks like the build system currently does not provide any way of checking if libecl is available in the C++ code?

andlaus:
> again. opm-parser is available. opm-core is not

then libecl should be available as well!? opm-parser will not work without it.

andlaus:
> some parts of the code is protected by HAVE_ERT, but not all. by removing opm-core this issue surfaces. but the real issue is the use of HAVE_ERT.

ah, now I see. that was slow in coming. since the code that uses libecl is only touched by the compiler if opm-parser is available, the increminating `#if` statements could either be removed or changed to `#if HAVE_OPM_PARSER`. I think that a good solution would be to export `HAVE_LIBECL` and use that, though.

andlaus:
okay, merging. this PR basically disables ECL output for plain ebos with the default build system, but IMO that's okay.

atgeirr:
> PR basically disables ECL output for plain ebos with the default build system, but IMO that's okay.

Not very ok though, if it means we are going to accept a functionality split going forward!

andlaus:
alternatively these guards could be removed completely or be replaced by HAVE_OPM_PARSER. both options are also fine with me.

also, this code is supposed to bite the dust with #262 anyway.

PR #269, opened 2018-01-08
Add support for PBDV and PDDV in initstateequil.hh

totto82:
Depends on OPM/opm-parser#1182

### DIFF
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -45,6 +45,7 @@ list(APPEND TEST_DATA_FILES
 	${TMP_VTPS}
 	${TMP_VTUS}
 	tests/data/equil_rsvd_and_rvvd.DATA
+        tests/data/equil_pbvd_and_pdvd.DATA
 	tests/data/equil_capillary_overlap.DATA
 	tests/data/equil_deadfluids.DATA
 	tests/data/equil_capillary.DATA
--- a/ebos/equil/equilibrationhelpers.hh
+++ b/ebos/equil/equilibrationhelpers.hh
@@ -241,6 +241,138 @@ private:
     }
 };
 
+/**
+ * Type that implements "dissolved gas-oil ratio"
+ * tabulated as a function of depth policy.  Data
+ * typically from keyword 'PBVD'.
+ */
+template <class FluidSystem>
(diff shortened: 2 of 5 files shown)


andlaus:
from the core eWoms angle, this looks good. so permission to merge at convenience granted.

I cannot say too much about the actual code changes, though. any opinions @bska or @atgeirr?

atgeirr:
> I cannot say too much about the actual code changes, though. any opinions @bska or @atgeirr?

@bska will review and test this for correctness

bska:
> will review and test this for correctness

Initial glance looks good, but I don't understand why you use constant extrapolation for PBVD and linear extrapolation for PDVD.  I would have thought that you'd need to treat those descriptions the same way.

totto82:
> Initial glance looks good, but I don't understand why you use constant extrapolation for PBVD and linear extrapolation for PDVD. I would have thought that you'd need to treat those descriptions the same way.

It based on testing against "the reference simulator", and I agree it is strange. One solution I haven't tested is to use linear extrapolation in one end and constant extrapolation in the other end. I can setup a few more tests to  check this out.

bska:
> I can setup a few more tests to check this out.

If it's not too much trouble, then yes I think that'd be a worthwhile endeavour.

totto82:
> If it's not too much trouble, then yes I think that'd be a worthwhile endeavour.

It was worth it as it reveled a bug. The pressure in the endpoints of the table should be restricted by the cell pressure.

bska:
> The pressure in the endpoints of the table should be restricted by the cell pressure.

Ah, right.  One final nit and then this can be installed.  The documentation for PDVD should refer to "RV" rather than "RS" when describing the input parameters to the function call operator.  Other than that I'm satisfied with the state of this.

totto82:
@bska Thanks for the review.

atgeirr:
I'll start a test, we'll merge when green.

atgeirr:

andlaus:
@totto82: do you consider this ready to be merged?

totto82:
Yes, it is ready when Jenkins is ready.

atgeirr:
Then in it goes! Thanks for the patch, and for the review.

PR #271, opened 2018-01-12
Revert "add back the epsilon value when switching from {oil,gas}-only to both hydrocarbon phases"

andlaus:
this makes flow work for SPE1CASE1 (actually, I do not really know why), but it might lead to different convergence behaviour in some cases.

@totto82: can you test and assess?

### DIFF
--- a/ewoms/models/blackoil/blackoilprimaryvariables.hh
+++ b/ewoms/models/blackoil/blackoilprimaryvariables.hh
@@ -461,7 +461,7 @@ public:
                                                                     SoMax);
 
             Scalar Rs = (*this)[Indices::compositionSwitchIdx];
-            if (Rs > (1.0 + eps)*std::min(RsMax, RsSat)) {
+            if (Rs > std::min(RsMax, RsSat*(1.0 + eps))) {
                 // the gas phase appears, i.e., switch the primary variables to { Sw, po,
                 // Sg }.
                 setPrimaryVarsMeaning(Sw_po_Sg);
@@ -518,7 +518,7 @@ public:
                                                                      SoMax);
 
             Scalar Rv = (*this)[Indices::compositionSwitchIdx];
-            if (Rv > (1.0 + eps)*std::min(RvMax, RvSat)) {
+            if (Rv > std::min(RvMax, RvSat*(1.0 + eps))) {
                 // switch to phase equilibrium mode because the oil phase appears. here
                 // we also need the capillary pressures to calculate the oil phase
                 // pressure using the gas phase pressure


atgeirr:
> this makes flow work for SPE1CASE1 (actually, I do not really know why)

Since this change was made in the original PR to keep behaviour consistent with the existing Flow I am a little skeptical... We must figure out why sufficiently to get both consistent behaviour and a working SPE1CASE1. It's good that you found the source of the trouble!

andlaus:
yes, that's the dilemma I'm worried about, too. Mathematically, not using a "switch hysteresis" should be The Right Thing (TM) to do, right?

totto82:
> yes, that's the dilemma I'm worried about, too. Mathematically, not using a "switch hysteresis" should be The Right Thing (TM) to do, right?

I agree. It is a trick to avoid oscillation between saturated and unsaturated case.

totto82:
Sorry. Hit the wrong button.

atgeirr:
> Mathematically, not using a "switch hysteresis" should be The Right Thing (TM) to do, right?

I guess it could be viewed as a type of damping?

ewoms/models/blackoil/blackoilprimaryvariables.hh
```diff
@@ -461,7 +461,7 @@ public:
                                                                     SoMax);
 
             Scalar Rs = (*this)[Indices::compositionSwitchIdx];
-            if (Rs > (1.0 + eps)*std::min(RsMax, RsSat)) {
+            if (Rs > std::min(RsMax, RsSat)) {
                 // the gas phase appears, i.e., switch the primary variables to { Sw, po,
```
totto82: ` Rs > std::min(RsMax, (1.0 + eps)*RsSat)`

Is inconsistent, but works. Shouldn't give any difference for the base case either. 
The same trick can also be applied to Rv.

andlaus:
> I guess it could be viewed as a type of damping?

not really; the update itself isn't altered by it. In this case, it is even less so because in `flow` the epsilon value is only non-zero in the iteration directly after a PV switch IIRC. maybe I got confused by the frequent back-and-forth here, though...

ewoms/models/blackoil/blackoilprimaryvariables.hh
andlaus: good idea, we have a fallback plan :)

on the other hand, I think that we should try to go the whole way and try to remove the epsilon altogether. Do you remember if doing so leads to any serious regressions or were they just changes that can be attributed to noise?

totto82: It gives almost 10% increase on model 2 rel 5. I agree that we should try to remove it, but implementing and testing alternatives is not a trivial task. I suggest merging the fallback plan and think this through later.

andlaus: okay, I agree that 10% is too much. Let's go with your suggestion.

PR #273, opened 2018-01-17
remove opm-core from upstream and prerequisite lists

akva2:


andlaus:
thanks. merge whenever.

PR #275, opened 2018-01-19
Some changes to facilitate building using a super project 

akva2:
See https://github.com/OPM/opm-common/pull/321


andlaus:
no real objections, merge whenever.

a thing which strikes me as a bit ugly is that the "DEPENDS ewoms_prepare" must be added to all invocations of `opm_add_test()`. possibly it is better to modify the opm_add_test() macro, but there might be difficulties with this that I do not see?

akva2:
i could modify the macro to read a global variable, if you prefer. i just used existing facilities.

andlaus:
both are fine with me, so pick the one which you prefer. isn't there a "PROJECT_NAME" variable so that  opm_add_test() macro could make tests automatically depend on the "${PROJECT_NAME}_prepare" target if it exists?

akva2:
hmm, i like that idea.

PR #276, opened 2018-01-22
refactor the boundary condition handling slightly

andlaus:
instead of passing a "minimal" fluid state that defines the thermodynamic conditions on the domain boundary and the models calculating everything they need based on this, it is now assumed that all quantities needed by the code that computes the boundary fluxes are defined. This simplifies the boundary flux computation code, it allows to get rid of the `paramCache` argument for these methods and to potentially speed things up because quantities do not get re-calculated unconditionally.

on the flipside, this requires slightly more effort to define the conditions at the boundary on the problem level and it makes it less obvious which quantities are actually used. That said, one now has the
freedom to shoot oneself into the foot more easily when specifying boundary conditions and also tools like valgrind or ASAN will normally complain about undefined quantities if this happens.

### DIFF
--- a/ewoms/models/common/darcyfluxmodule.hh
+++ b/ewoms/models/common/darcyfluxmodule.hh
@@ -337,8 +337,7 @@ protected:
     void calculateBoundaryGradients_(const ElementContext& elemCtx,
                                      unsigned boundaryFaceIdx,
                                      unsigned timeIdx,
-                                     const FluidState& fluidState,
-                                     const typename FluidSystem::template ParameterCache<typename FluidState::Scalar>& paramCache)
+                                     const FluidState& fluidState)
     {
         const auto& gradCalc = elemCtx.gradientCalculator();
         Ewoms::BoundaryPressureCallback<TypeTag, FluidState> pressureCallback(elemCtx, fluidState);
@@ -443,9 +442,15 @@ protected:
             }
 
             // take the phase mobility from the DOF in upstream direction
-            if (upstreamDofIdx_[phaseIdx] < 0)
-                mobility_[phaseIdx] =
-                    kr[phaseIdx] / FluidSystem::viscosity(fluidState, paramCache, phaseIdx);
+            if (upstreamDofIdx_[phaseIdx] < 0) {
+                if (interiorDofIdx_ == focusDofIdx)
+                    mobility_[phaseIdx] =
+                        kr[phaseIdx] / fluidState.viscosity(phaseIdx);
+                else
+                    mobility_[phaseIdx] =
+                        Toolbox::value(kr[phaseIdx])
+                        / Toolbox::value(fluidState.viscosity(phaseIdx));
+            }
             else if (upstreamDofIdx_[phaseIdx] != focusDofIdx)
                 mobility_[phaseIdx] = Toolbox::value(intQuantsIn.mobility(phaseIdx));
             else
--- a/ewoms/disc/common/fvbaseextensivequantities.hh
+++ b/ewoms/disc/common/fvbaseextensivequantities.hh
@@ -90,8 +90,7 @@ public:
     void updateBoundary(const Context& context,
                         unsigned bfIdx,
                         unsigned timeIdx,
-                        const FluidState& fluidState OPM_UNUSED,
-                        typename FluidSystem::template ParameterCache<typename FluidState::Scalar>& paramCache OPM_UNUSED)
+                        const FluidState& fluidState OPM_UNUSED)
     {
         unsigned dofIdx = context.interiorScvIndex(bfIdx, timeIdx);
         interiorScvIdx_ = static_cast<unsigned short>(dofIdx);
--- a/ewoms/models/common/forchheimerfluxmodule.hh
+++ b/ewoms/models/common/forchheimerfluxmodule.hh
@@ -278,16 +278,16 @@ protected:
         if (focusDofIdx == i) {
             ergunCoefficient_ =
                 (intQuantsIn.ergunCoefficient() +
-                 Toolbox::value(intQuantsEx.ergunCoefficient()))/2;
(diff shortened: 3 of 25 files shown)


andlaus:

ewoms/models/common/darcyfluxmodule.hh
```diff
@@ -443,9 +442,15 @@ protected:
...
 
             // take the phase mobility from the DOF in upstream direction
-            if (upstreamDofIdx_[phaseIdx] < 0)
-                mobility_[phaseIdx] =
-                    kr[phaseIdx] / FluidSystem::viscosity(fluidState, paramCache, phaseIdx);
+            if (upstreamDofIdx_[phaseIdx] < 0) {
+                if (interiorDofIdx_ == focusDofIdx)
+                    mobility_[phaseIdx] =
+                        kr[phaseIdx] / fluidState.viscosity(phaseIdx);
+                else
```
blattms: Please add a small comment clarifying the difference between the if-else branches

ewoms/models/common/darcyfluxmodule.hh
```diff
@@ -443,9 +442,15 @@ protected:
...
+                    mobility_[phaseIdx] =
+                        Toolbox::value(kr[phaseIdx])
+                        / Toolbox::value(fluidState.viscosity(phaseIdx));
```
blattms: As you are using `Opm::getValue(...)` everywhere else in this patch instead of `Toolbox::value`, what is the reason here to do it differently?

atgeirr:
I assume this does not *currently* affect Flow, since it assumes no-flow boundaries? When/if adding more flexible boundary conditions for Flow this would matter, correct?

> on the flipside, this requires slightly more effort to define the conditions at the boundary on the problem level and it makes it less obvious which quantities are actually used. That said, one now has the freedom to shoot oneself into the foot more easily when specifying boundary conditions and also tools like valgrind or ASAN will normally complain about undefined quantities if this happens.

There are two more ways at least to guard against these problems:
 - ensure that fluid state members *must* be set (not allowing default construction etc.); or
 - initialize all fluid state members to values that are so unreasonable that any use would cause termination (NaN for example), in particular avoiding "reasonable" defaults.

The first may not be feasible to require, if so perhaps the second suggestion should be implemented?

andlaus:
> I assume this does not *currently* affect Flow, since it assumes no-flow boundaries? When/if adding more flexible boundary conditions for Flow this would matter, correct?

yes.

> The first may not be feasible to require, if so perhaps the second suggestion should be implemented?

also yes: the first point is not really possible because it leads to a chicken-and-egg problem because to compute the needed quantities, fluid systems require fluid states.

for the second point, I'm not sure if this does provide any added value compared to just using `Valgrind::SetUndefined(fluidState)` as far as I can see? (this is supposed to be a debugging tool and setting every member to NaN might lead to performance issues for release builds. maybe this should be added to the constructor, not sure if it already is there.)

ewoms/models/common/darcyfluxmodule.hh
andlaus: it is equivalent. Opm::getValue() is shorter though, and it works regardless of the type of the argument passed. we're slowly getting rid of using the toolboxes directly. I don't really want to make a fuzz about this and change the remaining toolbox opportunistically, but if you want, I'll revert this part of the patch...

andlaus:
yay! I managed to shoot myself into the foot for the richards lens problem. freedom feels so nice!

andlaus:

ewoms/models/common/darcyfluxmodule.hh
andlaus: this is the fundamental design needed by dense AD: you need to throw away the derivatives if they are not for the focused-uppon entity. I agree that a comment here would not hurt, but there is a myriad of places where it this also needs to be added and I don't feel like I've currently have the nerves to do this...

andlaus:
now that jenkins is happy, let's merge because I'll (hopefully) need this soon. @blattms: patches for improved comments are always welcome!

atgeirr:
> patches for improved comments are always welcome!

@andlaus I think when a reviewer requests specific improvements, we should usually provide them unless we disagree, rather than pushing the burden to the reviewer.



> I'm not sure if this does provide any added value compared to just using Valgrind::SetUndefined(fluidState) as far as I can see

That may very well be a good practice, but I guess that would only help catching errors when actually testing with valgrind, which does not seem very easy to automate (i.e. on Jenkins).

akva2:
it would not be hard to setup a job that run tests through valgrind (ctest has native support).

andlaus:
> @andlaus I think when a reviewer requests specific improvements, we should usually provide them unless we disagree, rather than pushing the burden to the reviewer.

I meant to say that while I will always agree that literate programming is a very nice thing to get people into a code base, OPM is not following this idiom in any way (to see what I mean, look at the opm-parser code). also, over time too extensive comments tend to become inconsistent with the actual code, i.e. in the best case they make code changes more difficult while in the worst they mislead their readers. So: I disagree that a comment is needed here, and this pattern has not been introduced by the PR, but I do not mind if somebody disagrees with me here as long as he/she is willing to take up the burden of doing the work.

> it would not be hard to setup a job that run tests through valgrind (ctest has native support).

I guess valgrind is too slow for running routinely, but address-sanitizer may be an option?

atgeirr:
> it would not be hard to setup a job that run tests through valgrind (ctest has native support).

Perhaps it's just my limited experience with it, but I used to see lots of false positives that would need manual inspection. Maybe it is better now?

andlaus:
last I've tried (which is a while ago, admittedly), flow worked through a full run of SPE1CASE2 on valgrind without any complaints.

akva2:
i did not make any statement about the usefulness, only that it's not hard to add such a job. i also faintly remember a lot of false or possibly not-so-false positives back from when i last executed the tests under valgrind.

PR #272, opened 2018-01-16
Add FIP, INIT and EGRID output 

totto82:
Writes INIT and EGRID files initially
Adds Fip to summary output if required.
Output Fip to log (.PRT) if Opm-logger is set
Add feature for setting ECL output directory. 

Should be merged together with PR in opm-simulator.

PR #278, opened 2018-01-22
FluidInPlace, INIT and EGRID output to Ebos

totto82:
This PR build on #272 and makes the old PR absolute.  
1) EGRID and INIT output is added to new output module ebos
2) Adds option for double_si restart output
3) Compute region FIP and block values on each core and communicates the results instead of depending on these being computed by opm-output
4) Only compute and communicate values asked for by the deck. 

Depends on OPM/opm-grid#302, OPM/opm-output#235

and needs update in opm-simulator. 

TODO: 
- [x] Double check big models 
- [x] Double check MPI run

### DIFF
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -51,7 +51,7 @@ list(APPEND TEST_DATA_FILES
 	tests/data/equil_capillary.DATA
 	tests/data/equil_capillary_swatinit.DATA
 	tests/data/equil_base.DATA
-	tests/data/equil_livegas.DATAq
+	tests/data/equil_livegas.DATA
 	tests/data/equil_liveoil.DATA
 	tests/data/summary_deck_non_constant_porosity.DATA
 	)
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -51,8 +51,9 @@ list(APPEND TEST_DATA_FILES
 	tests/data/equil_capillary.DATA
 	tests/data/equil_capillary_swatinit.DATA
 	tests/data/equil_base.DATA
-	tests/data/equil_livegas.DATA
+	tests/data/equil_livegas.DATAq
 	tests/data/equil_liveoil.DATA
+	tests/data/summary_deck_non_constant_porosity.DATA
 	)
 
 list(APPEND TEST_SOURCE_FILES)
--- a/ebos/ecloutputblackoilmodule.hh
+++ b/ebos/ecloutputblackoilmodule.hh
@@ -43,8 +43,6 @@
 
 #include <type_traits>
 
-#define ENUM_TO_STR(ENUM) std::string(#ENUM)
-
 namespace Ewoms {
 namespace Properties {
 // create new type tag for the Ecl-output
@@ -87,17 +85,17 @@ class EclOutputBlackOilModule
     struct FIPDataType {
 
         enum FipId {
-            WIP = waterPhaseIdx,
-            OIP = oilPhaseIdx,
-            GIP = gasPhaseIdx,
-            OIPL = 3,
-            OIPG = 4,
-            GIPL = 5,
-            GIPG = 6,
-            PV   = 7, //< Pore volume
-            PAV  = 8
+            WaterInPlace = 0, //WIP
+            OilInPlace = 1, //OIP
+            GasInPlace = 2, //GIP
+            OilInPlaceInLiquidPhase = 3, //OIPL
+            OilInPlaceInGasPhase = 4, //OIPG
+            GasInPlaceInLiquidPhase = 5, //GIPL
+            GasInPlaceInGasPhase = 6, //GIPG
+            PoreVolume   = 7, //PV
+            HydroCarbonPoreVolumeAveragedPressure  = 8  //PAV
         };
-        static const int fipValues = PAV + 1 ;
+        static const int numFipValues = HydroCarbonPoreVolumeAveragedPressure + 1 ;
     };
 
 public:
@@ -160,10 +158,8 @@ public:
         }
 
         // Fluid in place
-        for (int i = 0; i<FIPDataType::fipValues; i++) {
-            //std::cout << stringOfEnumIndex_(i) << std::endl;
+        for (int i = 0; i<FIPDataType::numFipValues; i++) {
             if (!substep || summaryConfig.require3DField(stringOfEnumIndex_(i))) {
-                //std::cout << "required " << stringOfEnumIndex_(i) << std::endl;
                 fip_[i].resize(bufferSize, 0.0);
             }
         }
@@ -484,7 +480,7 @@ public:
                                                                                         intQuants.pvtRegionIndex());
             }
 
-            // FIP
+            // Fluid in Place calculations
 
             // calculate the pore volume of the current cell. Note that the porosity
             // returned by the intensive quantities is defined as the ratio of pore
@@ -496,8 +492,8 @@ public:
             const double pv =
                 elemCtx.simulator().model().dofTotalVolume(globalDofIdx)
                 * intQuants.porosity().value();
-            if (fip_[FIPDataType::PV].size() > 0)
-                fip_[FIPDataType::PV][globalDofIdx] = pv;
+            if (fip_[FIPDataType::PoreVolume].size() > 0)
+                fip_[FIPDataType::PoreVolume][globalDofIdx] = pv;
 
             Scalar fip[FluidSystem::numPhases];
             for (unsigned phase = 0; phase < FluidSystem::numPhases; ++phase) {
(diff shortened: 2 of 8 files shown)


ebos/ecloutputblackoilmodule.hh
```diff
@@ -81,42 +85,118 @@ class EclOutputBlackOilModule
...
+                fip_[i].resize(bufferSize, 0.0);
+                computeFip_ = true;
+            } else {
+                fip_[i].clear();
+            }
+        }
+        if (!substep || summaryConfig.hasKeyword("FPR") || summaryConfig.hasKeyword("FPRP") || summaryConfig.hasKeyword("RPR")) {
+            fip_[FIPDataType::PoreVolume].resize(bufferSize, 0.0);
+            pvHydrocarbon_.resize(bufferSize, 0.0);
+            pPv_.resize(bufferSize, 0.0);
```
andlaus: more descriptive variable names would be a good idea here, e.g. what does the first 'p' mean here? I figure that `pv` stands for pore volume?

ebos/ecloutputblackoilmodule.hh
```diff
@@ -438,8 +518,104 @@ public:
...
+                    if (!FluidSystem::phaseIsActive(phase)) {
+                        continue;
+                    }
+
+                    const double b = Toolbox::value(fs.invB(phase));
+                    const double s = Toolbox::value(fs.saturation(phase));
+                    fip[phase] = b * s * pv;
+                }
+
+                if (FluidSystem::phaseIsActive(oilPhaseIdx) && fip_[FIPDataType::OilInPlace].size() > 0)
```
andlaus: instead of `.size() > 0` it is slightly better to use `!empty()`. for `std::vector` it does not matter, but e.g. for `std::list` it can.

ebos/ecloutputblackoilmodule.hh
```diff
@@ -438,8 +518,104 @@ public:
                                                                                         intQuants.pvtRegionIndex());
             }
 
-        }
+            // Fluid in Place calculations
```
andlaus: maybe the FIP code should be moved into a separate method, e.g. `updateFluidInPlace_()`. if there are difficulties with this, leave it as-is.

ebos/ecloutputblackoilmodule.hh
```diff
@@ -438,8 +518,104 @@ public:
...
+                // Store the pure oil and gas FIP
+                if (FluidSystem::phaseIsActive(oilPhaseIdx) && fip_[FIPDataType::OilInPlaceInLiquidPhase].size() > 0)
+                    fip_[FIPDataType::OilInPlaceInLiquidPhase][globalDofIdx] = fip[oilPhaseIdx];
+
+                if (FluidSystem::phaseIsActive(gasPhaseIdx) && fip_[FIPDataType::GasInPlaceInGasPhase].size() > 0)
+                    fip_[FIPDataType::GasInPlaceInGasPhase][globalDofIdx] = fip[gasPhaseIdx];
+
+                if (FluidSystem::phaseIsActive(oilPhaseIdx) && FluidSystem::phaseIsActive(gasPhaseIdx)) {
+                    // Gas dissolved in oil and vaporized oil
+                    Scalar gipl = Toolbox::value(fs.Rs()) * fip[oilPhaseIdx];
```
andlaus: again, I have problems to determine the meaning of these variables without a manual...

andlaus: Also, it's slighly better to use `Opm::getValue(fs.Rs())` instead of the toolbox directly. this is because `Toolbox<Eval>::value(x)` will only work for `x` of `Eval` type while `Opm::getValue()` should work unconditionally. (Admittedly this is not a problem here.)

ebos/ecloutputblackoilmodule.hh
```diff
@@ -438,8 +518,104 @@ public:
...
+            for( auto& val : blockValues_ ) {
+                const auto& key = val.first;
+                int global_index = key.second - 1;
+                if (global_index == globalIdx) {
+                    if (key.first == "BWSAT") {
+                        val.second = Toolbox::value(fs.saturation(waterPhaseIdx));
+                    } else if (key.first == "BGSAT") {
+                        val.second = Toolbox::value(fs.saturation(gasPhaseIdx));
+                    } else if (key.first == "BPR") {
+                        std::cout << "BPR set "<<Toolbox::value(fs.pressure(oilPhaseIdx)) << std::endl;
```
andlaus: debugging artifact?

ebos/ecloutputblackoilmodule.hh
```diff
@@ -614,11 +790,115 @@ public:
...
+                sol.insert(stringOfEnumIndex_(i),
+                           Opm::UnitSystem::measure::volume,
+                           fip_[i] ,
+                           Opm::data::TargetType::SUMMARY);
+            }
+        }
+    }
+
+    // write Fluid In Place to output log
+    void outputFIPLog(std::map<std::string, double>& miscSummaryData,  std::map<std::string, std::vector<double>>& regionData, const bool substep) {
```
andlaus: can this method be made private?

ebos/ecloutputblackoilmodule.hh
```diff
@@ -614,11 +790,115 @@ public:
...
+                origRegionValues_[i] = regionFipValues[i];
+        }
+
+        // sum all region values to compute the field total
+        std::vector<int> fieldNum(ntFip, 1);
+        ScalarBuffer fieldFipValues(FIPDataType::numFipValues,0.0);
+        bool comunicateSum = false; // the regionValues are already summed over all ranks.
+        for (int i = 0; i<FIPDataType::numFipValues; i++) {            
+            const ScalarBuffer& tmp = FIPTotals_(regionFipValues[i], fieldNum, 1, comunicateSum);
+            fieldFipValues[i] = tmp[0]; //
```
andlaus: empty comment.

ebos/ecloutputblackoilmodule.hh
```diff
@@ -614,11 +790,115 @@ public:
...
+        }
+
+        // compute the hydrocarbon averaged pressure over the regions.
+        ScalarBuffer regPressurePv = FIPTotals_(pPv_, fipnum_, ntFip);
```
andlaus: s/FIPTotals_/fipTotals_/

ebos/ecloutputblackoilmodule.hh
```diff
@@ -767,8 +1055,172 @@ private:
...
+    }
+
+    Scalar pressureAverage_(const Scalar& pressurePvHydrocarbon, const Scalar& pvHydrocarbon, const Scalar& pressurePv, const Scalar& pv, bool hydrocarbon) {
+        if (pvHydrocarbon > 1e-10 && hydrocarbon)
+            return pressurePvHydrocarbon / pvHydrocarbon;
+
+        return pressurePv / pv;
+    }
+
+    void FIPUnitConvert_(ScalarBuffer& fip)
```
andlaus: s/FIPUnitConvert_/fipUnitConvert_/

ebos/ecloutputblackoilmodule.hh
```diff
@@ -767,8 +1055,172 @@ private:
...
+           << std::setw(13) << cip[FIPDataType::WaterInPlace] << "   :" << std::setw(14) << (cip[FIPDataType::GasInPlaceInGasPhase]) << std::setw(14) << cip[FIPDataType::GasInPlaceInLiquidPhase] << std::setw(14) << cip[FIPDataType::GasInPlace] << ":\n"
+           << ":------------------------:------------------------------------------:----------------:------------------------------------------:\n"
+           << ":Originally  in place    :" << std::setw(14) << oip[FIPDataType::OilInPlaceInLiquidPhase] << std::setw(14) << oip[FIPDataType::OilInPlaceInGasPhase] << std::setw(14) << oip[FIPDataType::OilInPlace] << ":"
+           << std::setw(13) << oip[FIPDataType::WaterInPlace] << "   :" << std::setw(14) << oip[FIPDataType::GasInPlaceInGasPhase] << std::setw(14) << oip[FIPDataType::GasInPlaceInLiquidPhase] << std::setw(14) << oip[FIPDataType::GasInPlace] << ":\n"
+           << ":========================:==========================================:================:==========================================:\n";
+        Opm::OpmLog::note(ss.str());
+    }
+
+    std::string stringOfEnumIndex_(int i) {
```
andlaus: that name is a bit too vague for my taste. how about `fipEnumToString_()`?

ebos/ecloutputblackoilmodule.hh
```diff
@@ -767,8 +1055,172 @@ private:
...
+        switch( static_cast<FipId>(i) )
+        {
+        case FIPDataType::WaterInPlace: return "WIP";  break;
+        case FIPDataType::OilInPlace: return "OIP";  break;
+        case FIPDataType::GasInPlace: return "GIP";  break;
+        case FIPDataType::OilInPlaceInLiquidPhase: return "OIPL"; break;
+        case FIPDataType::OilInPlaceInGasPhase: return "OIPG"; break;
+        case FIPDataType::GasInPlaceInLiquidPhase: return "GIPL"; break;
+        case FIPDataType::GasInPlaceInGasPhase: return "GIPG"; break;
+        case FIPDataType::PoreVolume: return "PV"; break;
```
andlaus: you don't need a `break` after `return`.

Also, possibly add `default: return "ERROR";` to the switch statement. On the other hand it may be better to let the compiler complain about missing cases...

ebos/eclproblem.hh
```diff
@@ -1139,12 +1173,13 @@ public:
         return initialFluidStates_[globalDofIdx];
     }
 
-    void setEclIO(std::unique_ptr<Opm::EclipseIO>&& eclIO) {
-        eclWriter_->setEclIO(std::move(eclIO));
-    }
-
     const Opm::EclipseIO& eclIO() const
```
andlaus: with the NNC stuff moved to core ebos, is this still needed? if yes, what's still missing?

ebos/eclwriter.hh
```diff
@@ -199,7 +218,7 @@ public:
         unsigned episodeIdx = simulator_.episodeIndex();
         const auto& gridView = simulator_.gridManager().gridView();
         unsigned numElements = gridView.size(/*codim=*/0);
-        eclOutputModule_.allocBuffers(numElements, episodeIdx, simulator_.gridManager().eclState().getRestartConfig(), true, false);
+        eclOutputModule_.allocBuffers(numElements, episodeIdx, false, false, collectToIORank_);
```
andlaus: what do these boolean statements mean? adding `/*aComment=*/` would probably be a good idea.

ebos/eclproblem.hh
totto82: It is used for getting well data for restart.

ebos/ecloutputblackoilmodule.hh
totto82: It is used by eclwriter.

ebos/eclproblem.hh
andlaus: ok, no problem then

ebos/ecloutputblackoilmodule.hh
andlaus: ok. I just noticed another tiny nit: could you please replace the remaining occurrences of all-caps "FIP" by "Fip"?

atgeirr:
> Note the commit that removes the forced consistency of the initial state explicitly given by the deck. I had to do this to get the tests right. I am quite sure that naive usage of the input data is what users expect to be the default behavior of the simulator. Adding a parameter for this could be an idea if we want to force consistent initial solution.

I expect that would cause a ripple effect across all the simulation results?

atgeirr:
I am happy to see the added test!

In general I am worried about how some of the classes in ewoms have to be written though (such as the EclOutputBlackOilModule class): they require the simulator class, yet perform no simulation. It is only used as a place to get other stuff, such as "is vappars active", the material laws, the EclipseState etc.

I believe that contributes significantly to making it hard to write tests.

andlaus:
> they require the simulator class, yet perform no simulation.

it is probably helpful to think of passing the simulator object as "this class/function depends on something of the simulation run".

> I believe that contributes significantly to making it hard to write tests.

probably. That said it considerably simplifies implementing the functionality itself.

totto82:
> I expect that would cause a ripple effect across all the simulation results?

All tests and field models uses EQUIL initialization where "no forced consistency" is already the default behavior. This just make the direct initialization using PRESSURE, SWAT etc. consistent with the EQUIL initialization.

atgeirr:
> All tests and field models uses EQUIL initialization where "no forced consistency" is already the default behavior. This just make the direct initialization using PRESSURE, SWAT etc. consistent with the EQUIL initialization.

Should be fine, then!

andlaus:
any objections to merging this @atgeirr?

atgeirr:
> any objections to merging this @atgeirr?

No, I just started Jenkins, will update data and merge the lot when green. Probably tomorrow morning though.

CMakeLists_files.cmake
```diff
@@ -51,8 +51,9 @@ list(APPEND TEST_DATA_FILES
 	tests/data/equil_capillary.DATA
 	tests/data/equil_capillary_swatinit.DATA
 	tests/data/equil_base.DATA
-	tests/data/equil_livegas.DATA
```
andlaus: @totto82: keyboard slippage?

PR #279, opened 2018-01-30
Implement blackoil energy

andlaus:
this implements an energy conservation extension for the black-oil model. It requires some small downstream mop-up because the indices class expects an additional template argument which specifies if energy is enabled or not.

I've verified that the performance of `flow` for norne does not change with this. Further testing is highly appreciated, though.

### DIFF
--- a/ebos/eclequilinitializer.hh
+++ b/ebos/eclequilinitializer.hh
@@ -77,12 +77,15 @@ class EclEquilInitializer
     enum { waterCompIdx = FluidSystem::waterCompIdx };
 
     enum { dimWorld = GridView::dimensionworld };
+    enum { enableTemperature = GET_PROP_VALUE(TypeTag, EnableTemperature) };
     enum { enableEnergy = GET_PROP_VALUE(TypeTag, EnableEnergy) };
 
+    // NB: setting the enableEnergy argument to true enables storage of enthalpy and
+    // internal energy!
     typedef Opm::BlackOilFluidState<Scalar,
                                     FluidSystem,
-                                    /*enableTemperature=*/true,
-                                    /*enableEnthalpy=*/enableEnergy> ScalarFluidState;
+                                    enableTemperature,
+                                    enableEnergy> ScalarFluidState;
 
 public:
     template <class EclMaterialLawManager>
@@ -101,9 +104,6 @@ public:
                                                                          eclState,
                                                                          vanguard.grid(),
                                                                          simulator.problem().gravity()[dimWorld - 1]);
-
-        const std::vector<double>& tempiData =
-            eclState.get3DProperties().getDoubleGridProperty("TEMPI").getData();
 
         // copy the result into the array of initial fluid states
         initialFluidStates_.resize(numCartesianElems);
@@ -135,18 +135,12 @@ public:
 
 
             // set the temperature.
-            //
-            // TODO: setting temperature explicitly while computing static equilibirum
-            //       for everything else is a bit inconsistent, i.e.,
-            //       Opm::initStateEquil() should be extended to provide correct initial
-            //       temperatures
-            assert(std::isfinite(tempiData[cartesianElemIdx]));
-            fluidState.setTemperature(tempiData[cartesianElemIdx]);
+            if (enableTemperature || enableEnergy)
+                fluidState.setTemperature(initialState.temperature()[elemIdx]);
 
             // set the phase pressures.
             for (unsigned phaseIdx = 0; phaseIdx < numPhases; ++phaseIdx)
                 fluidState.setPressure(phaseIdx, initialState.press()[phaseIdx][elemIdx]);
-
         }
     }
 
--- a/ebos/eclequilinitializer.hh
+++ b/ebos/eclequilinitializer.hh
@@ -88,15 +88,19 @@ public:
         : simulator_(simulator)
     {
         const auto& vanguard = simulator.vanguard();
+        const auto& eclState = vanguard.eclState();
 
         unsigned numElems = vanguard.grid().size(0);
         unsigned numCartesianElems = vanguard.cartesianSize();
         typedef typename GET_PROP_TYPE(TypeTag, FluidSystem) FluidSystem;
 
         EQUIL::DeckDependent::InitialStateComputer<TypeTag> initialState(materialLawManager,
-                                                                         vanguard.eclState(),
+                                                                         eclState,
                                                                          vanguard.grid(),
                                                                          simulator.problem().gravity()[dimWorld - 1]);
+
+        const std::vector<double>& tempiData =
+            eclState.get3DProperties().getDoubleGridProperty("TEMPI").getData();
 
         // copy the result into the array of initial fluid states
         initialFluidStates_.resize(numCartesianElems);
@@ -127,10 +131,14 @@ public:
                 fluidState.setRv(0.0);
 
 
-            // set the temperature
-            // TODO Get the temperature from the initialState
-            Scalar T = FluidSystem::surfaceTemperature;
-            fluidState.setTemperature(T);
+            // set the temperature.
+            //
+            // TODO: setting temperature explicitly while computing static equilibirum
+            //       for everything else is a bit inconsistent, i.e.,
+            //       Opm::initStateEquil() should be extended to provide correct initial
+            //       temperatures
+            assert(std::isfinite(tempiData[cartesianElemIdx]));
+            fluidState.setTemperature(tempiData[cartesianElemIdx]);
 
             // set the phase pressures.
             for (unsigned phaseIdx = 0; phaseIdx < numPhases; ++phaseIdx)
--- a/ebos/eclequilinitializer.hh
+++ b/ebos/eclequilinitializer.hh
@@ -44,6 +44,7 @@ NEW_PROP_TAG(FluidSystem);
 NEW_PROP_TAG(GridView);
 NEW_PROP_TAG(Scalar);
 NEW_PROP_TAG(MaterialLaw);
+NEW_PROP_TAG(EnableEnergy);
 }
 
 /*!
@@ -65,10 +66,6 @@ class EclEquilInitializer
     typedef typename GET_PROP_TYPE(TypeTag, Scalar) Scalar;
     typedef typename GET_PROP_TYPE(TypeTag, MaterialLaw) MaterialLaw;
 
-    typedef Opm::BlackOilFluidState<Scalar,
-    FluidSystem,
-    /*enableTemperature=*/true> ScalarFluidState;
-
     enum { numPhases = FluidSystem::numPhases };
     enum { oilPhaseIdx = FluidSystem::oilPhaseIdx };
     enum { gasPhaseIdx = FluidSystem::gasPhaseIdx };
@@ -80,6 +77,12 @@ class EclEquilInitializer
     enum { waterCompIdx = FluidSystem::waterCompIdx };
 
     enum { dimWorld = GridView::dimensionworld };
+    enum { enableEnergy = GET_PROP_VALUE(TypeTag, EnableEnergy) };
+
+    typedef Opm::BlackOilFluidState<Scalar,
+                                    FluidSystem,
+                                    /*enableTemperature=*/true,
+                                    /*enableEnthalpy=*/enableEnergy> ScalarFluidState;
 
 public:
     template <class EclMaterialLawManager>
--- a/ebos/eclpeacemanwell.hh
+++ b/ebos/eclpeacemanwell.hh
@@ -116,8 +116,13 @@ class EclPeacemanWell : public BaseAuxiliaryModule<TypeTag>
 
     static const unsigned numModelEq = GET_PROP_VALUE(TypeTag, NumEq);
     static const unsigned conti0EqIdx = GET_PROP_TYPE(TypeTag, Indices)::conti0EqIdx;
+    static const unsigned contiEnergyEqIdx = GET_PROP_TYPE(TypeTag, Indices)::contiEnergyEqIdx;
 
-    typedef Opm::CompositionalFluidState<Scalar, FluidSystem, /*storeEnthalpy=*/false> FluidState;
+    static constexpr unsigned historySize = GET_PROP_VALUE(TypeTag, TimeDiscHistorySize);
+
+    static constexpr bool enableEnergy = GET_PROP_VALUE(TypeTag, EnableEnergy);
+
+    typedef Opm::CompositionalFluidState<Scalar, FluidSystem, /*storeEnthalpy=*/true> FluidState;
     typedef Dune::FieldMatrix<Scalar, dimWorld, dimWorld> DimMatrix;
 
     // all quantities that need to be stored per degree of freedom that intersects the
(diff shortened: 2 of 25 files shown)


andlaus:
jenkins build this with downstreams opm-simulators=1393 please

totto82:
Nice work! I have started to test this, both the thermal part in ebos and to verify that this doesn't effect the standard blackoil simulator. I can also do a review of the code.

andlaus:
> I can also do a review of the code.

yes, please! this branch has been in the works for quite some time and it thus needed to be rebased several times. it is very well possible that I screwed something up during this. Even if you don't find the time for doing a review: thanks for testing :)

ebos/eclequilinitializer.hh
```diff
@@ -80,6 +77,12 @@ class EclEquilInitializer
     enum { waterCompIdx = FluidSystem::waterCompIdx };
 
     enum { dimWorld = GridView::dimensionworld };
+    enum { enableEnergy = GET_PROP_VALUE(TypeTag, EnableEnergy) };
+
+    typedef Opm::BlackOilFluidState<Scalar,
+                                    FluidSystem,
+                                    /*enableTemperature=*/true,
+                                    /*enableEnthalpy=*/enableEnergy> ScalarFluidState;
```
totto82: change comment to enableEnergy?

ebos/eclequilinitializer.hh
```diff
@@ -127,10 +134,14 @@ public:
...
-            Scalar T = FluidSystem::surfaceTemperature;
-            fluidState.setTemperature(T);
+            // set the temperature.
+            //
+            // TODO: setting temperature explicitly while computing static equilibirum
+            //       for everything else is a bit inconsistent, i.e.,
+            //       Opm::initStateEquil() should be extended to provide correct initial
+            //       temperatures
+            assert(std::isfinite(tempiData[cartesianElemIdx]));
+            fluidState.setTemperature(tempiData[cartesianElemIdx]);
```
totto82: Have you tried to implement this in the equilcode? If I remember correctly the infrastructure for temperature is already in place. If the only thing missing is to ask the parser for the temperature, then I think this could be done in this PR.

ebos/eclpeacemanwell.hh
```diff
@@ -319,6 +324,17 @@ public:
...
+        // class contains some "holes" due to alignment
+        Opm::Valgrind::SetDefined(sol[wellGlobalDof]);
+
+        // also apply the initial solution of the well to the "old" time steps
+        for (unsigned timeIdx = 1; timeIdx < historySize; ++timeIdx) {
+            auto& oldSol = const_cast<SolutionVector&>(simulator_.model().solution(timeIdx));
+
+            oldSol[wellGlobalDof] = sol[wellGlobalDof];
+        }
     }
```
totto82: Could you explain what this is for?

ebos/eclproblem.hh
```diff
@@ -276,19 +308,22 @@ class EclProblem : public GET_PROP_TYPE(TypeTag, BaseProblem)
...
-
     typedef BlackOilSolventModule<TypeTag> SolventModule;
     typedef BlackOilPolymerModule<TypeTag> PolymerModule;
 
     typedef Opm::BlackOilFluidState<Scalar,
-            FluidSystem,
-            /*enableTemperature=*/true> InitialFluidState;
+                                    FluidSystem,
+                                    /*enableTemperature=*/true,
+                                    /*enableEnthalpy=*/enableEnergy> InitialFluidState;
```
totto82: enableEnergy?

ebos/eclequilinitializer.hh
andlaus: > Have you tried to implement this in the equilcode?

no, these lines stem from a time where the equil code was still located in opm-core. I just tried to implement your suggestion, but currently it only sets the temperature in the initial solution according to TEMPI. I also tried getting temperature dependent density to work there, but I good stuck because the code seems need a major refactoring in that area. maybe someone more familiar with the static equilibrium stuff should have a look.

ebos/eclequilinitializer.hh
andlaus: yes, changed. thanks. (note: this is an artifact from development.)

ebos/eclproblem.hh
andlaus: yes, thanks for the catch.

ewoms/models/blackoil/blackoilindices.hh
```diff
@@ -83,12 +97,16 @@ struct BlackOilIndices
     static const int compositionSwitchIdx = PVOffset + 2;
```
totto82: Nitpick: Maybe using  PVOffset + (numPhases -1) here as below to avoid confusion.

ewoms/models/blackoil/blackoilmodel.hh
```diff
@@ -121,10 +125,20 @@ public:
...
-// by default, the ECL solvent module is disabled
+// by default, all ECL extension modules are disabled
 SET_BOOL_PROP(BlackOilModel, EnableSolvent, false);
 SET_BOOL_PROP(BlackOilModel, EnablePolymer, false);
-// by default, ebos formulates the conservation equations in terms of mass not surface volumes
+
+//! By default, the blackoil model is isothermal and does not conserve energy
+SET_BOOL_PROP(BlackOilModel, EnableTemperature, false);
+SET_BOOL_PROP(BlackOilModel, EnableEnergy, false);
+
```
totto82: EnableTemperature is used to allow for non isothermal temperature, right. 

In eclproblem (and in the eclequilinitialization) the enableTemperature is hardcoded to true like this: 
```
    typedef Opm::BlackOilFluidState<Scalar,
                                    FluidSystem,
                                    /*enableTemperature=*/true,
                                    /*enableEnthalpy=*/enableEnergy> InitialFluidState;
```

With a EnableTemperature property in place I suggest removing the hardcoded "trues" , and replace it by enableTemperature. And set EnableTempeature to true for ebos.

ewoms/models/blackoil/blackoilindices.hh
```diff
@@ -35,7 +35,7 @@ namespace Ewoms {
  *
  * \brief The primary variable and equation indices for the black-oil model.
  */
-template <unsigned numSolventsV, unsigned numPolymersV, unsigned PVOffset>
+template <bool enableSolventV, bool enablePolymerV, bool enableEnergyV, unsigned PVOffset>
 struct BlackOilIndices
```
totto82: I agree. This seems like a reasonable simplification.

totto82:
Sorry. I had to abort the review yesterday. It am done now. 
Some comments: 
1) There seems to be an issue with unit conversion. I have added it as an issue to opm-parser. OPM/opm-parser#1202, this makes it hard to compare. 
2) There is no ecl temperature output. It is disabled in the current version (by me). Adding it is trivial. I can do it in my output PR so we avoid unnecessary merge issues. 
3) enableTemperature and enableEnergy is not set in ebos, ie. it uses default which is false. Maybe setting this explicitly to false in eclproblem.hh is good, then it is easier for outsiders to notice that there is a energy option. Just change "false" to "true" and recompile ebos. I guess for flow we should add the energy option similar to solvent and polymer, but that's another day. 
4) I can verify that this PR and the one in opm-simulator does not affect the results, convergence nor performance in anyway for flow. i.e. it is safe to merge.

ewoms/models/blackoil/blackoilindices.hh
andlaus: hm, I don't really object but I'm not sure if it is a good idea: First, this should not be done for `waterSaturationIdx` and `pressureSwitchIdx`, I suppose? the reason why I picked `PVOffset + (numPhases-1)` for the extension PV indices instead of simply `compositionSwitchIdx` is that the former can be copy-and-pasted to the twophase index classes while the latter expression is IMO slightly nicer but it does not work for `BlackoilTwoPhaseIndices` because `compositionSwitchIdx` can be disabled there.

totto82: No problem. Leave it as it is. I agree with your reasoning.

ewoms/models/blackoil/blackoilmodel.hh
andlaus: > EnableTemperature is used to allow for non isothermal temperature, right.

that's the idea, yes. So far it is completely untested, though...

> With a EnableTemperature property in place I suggest removing the hardcoded "trues" , and replace it by enableTemperature. And set EnableTempeature to true for ebos.

agreed. changed, (but not yet pushed) thanks. we need to test if this does not affect flow performance to badly because it increases the size if the intensive quantities slightly.

andlaus:
> There is no ecl temperature output. It is disabled in the current version (by me). Adding it is trivial. I can do it in my output PR so we avoid unnecessary merge issues.

that would be very nice. would it be much effort to also add output for the fluid enthalpies and the rock internal energy?

> enableTemperature and enableEnergy is not set in ebos, ie. it uses default which is false. Maybe setting this explicitly to false in eclproblem.hh is good

good point; changed. while at it, I also explicitly disabled the other black-oil extensions.

> I can verify that this PR and the one in opm-simulator does not affect the results, convergence nor performance in anyway for flow. i.e. it is safe to merge.

with EnableTemperature set to true, the size of the intensive quantities object increases slightly even for the base model, i.e., we should verify this again. (it is beneficial though because e.g RTEMPVD is reflected in the output.) my preliminary testing with norne shows that the performance penalty should be at most 1%, though.

andlaus:
okay, I've just pushed the patches that deal with the issues you raised in your review @totto82, i.e. it should be ready for testing.

andlaus:
any objections to merging this? I would like to get this done this week..

@totto82: there might be interference with your output PRs. please indicate which merge order you prefer.

totto82:
I really want the output stuff merged asap. It should be ready, just needs a reference data update. With the output stuff + the temperature offset merged I should be able to test the energy option more throughly.

andlaus:
@totto82: this now applies cleanly again. (There were some small issues with the recent output changes and with headers changing their locations.) also, with OPM/opm-parser#1205 being merged, it should be relatively straight forward to test.

totto82:
Great. I will give it a try.

totto82:
To comments: 

1) Initial solution seems to be slightly wrong. 
2) It seems like ebos does not cool down the reservoir even when injecting cold fluid

andlaus:
did you set the EnableEnergy property to true? so far, ebos is still isothermal only. that said, I noticed that the heat conduction seems to be too large as well. do you get better results if you set THCONR to 0?

totto82:
> did you set the EnableEnergy property to true? so far, ebos is still isothermal only by default. that said, I noticed that the heat conduction seems to be too large as well. do you get better results if you set THCONR to 0?

Yes, I enabledEnergy. I have only tried the base case. I can do some more testing today.

totto82:
THCONR = 0 did do much. My guess is that it is well related. I try to inject 100kK into the reservoir and still the temperature increase is less than 15K.

andlaus:
for the heat conduction I'm really unsure that there are no mistakes when calculating the thermal transmissibilities et cetera. I've verified that the values specified by THCONR in SPE1CASE2_THERMAL  is in the right ballpark (about 5 W/(m K), IIRC), but the thermal flux over the boundary is about 2W/m^2 and seems to be too large, though.

andlaus:
I just had an idea: Maybe the volumetric internal energy of the rock is wrong, what happens if you set SPECROCK to 0 (or very close to it)?

andlaus:
jenkins build this please

andlaus:
jenkins build this opm-simulators=1393 please

andlaus:
jenkins build this opm-simulators=1393 please

andlaus:
jenkins build this opm-simulators=1393 please

totto82:
I have reviewed and tested this. From my point of view, it can be merged. If @atgeirr doesn't object I will merge this later today.

atgeirr:
Let's merge!

ewoms/models/blackoil/blackoiltwophaseindices.hh
```diff
@@ -37,57 +37,74 @@ namespace Ewoms {
  *
  * \brief The primary variable and equation indices for the black-oil model.
  */
-template <unsigned numSolventsV, unsigned numPolymersV, unsigned PVOffset, unsigned disabledCanonicalCompIdx>
+template <bool enableSolventV, bool enablePolymerV, bool enableEnergyV, unsigned PVOffset, unsigned disabledCanonicalCompIdx>
```
GitPaean: This change makes it less flexible in some way. Originally, it supports the situation where `numPolymersV` can be other than 1, like the situation I needs it to be 2.   For sure, I can add more `enable***` flags to the constructor. Or I can just set `numPolymersV` to be 2, which can be easier.

andlaus: You're correct in principle. in practice you would have gotten a *big* headache if you chose num*V to be unequal 1, so these being integers was at least misleading.

do you want to develop a model with two distinct solvent and/or polymer components? if yes, I don't mind if you change them back to integers in that PR.

andlaus: so what's the additional quantity that is conserved in your model? to me, it rather sounds like your stuff is an alternative implementation of the polymer model...

GitPaean: It is the transport equation for the polymer molecular weight. It is a little different polymer model. But with another extra flag, it will not impact the current polymer model. More like, with this flag, we calculate the viscosity this way, otherwise, that way.   With this flag, we have another primary variable and equation, otherwise, no.

andlaus: > It is the transport equation for the polymer molecular weight. 

this is not a physical quantity. It sounds like your new model conserves the mass of two polymer components using the primary variables (total polymer mass, molecular weight of polymer mix)?

GitPaean: > this is not a physical quantity. 

I am not sure what you mean here. You can consider it is some property attached to polymer that can evolve, or you can consider it is a tracer, but we decide to solve it fully implicitly here.

andlaus: I mean what mother nature conserves. This could either be component mass, energy or momentum but not things like pressure, saturations, temperature or molecular weight. To me it sounds a lot like your new model uses two (pseudo-)components for the polymers and conserves the mass of each of them by tracking the total polymer mass plus their composition (via the average polymer mass).

GitPaean: You can consider the total number of the atoms in the polymer (same with mass) that conserves.  Polymer can have bigger molecular and smaller molecular. The bigger molecular can be broken down to smaller molecular. It will not affect the polymer concentration (kg/m^3), while average molecular weight will change when the composition of the polymer changes. 

My current version will be more like a tracer, it is attached to the polymer when injected,  MC will be conserved. M is the molecular weight.

andlaus: Okay, I guess we agree that you should change the polymers template parameter back to an integer. I'm curious about the actual PR.

PR #281, opened 2018-02-05
fixed: remove erroneous brace

akva2:


andlaus:
obvious mistake in build system. merge whenever.

andlaus:
> Too many modules..., sigh.

note that the only reason that this mistake affected each and every module is that this was a bug in the "sibling build" infrastructure which needs to be copy-and-pasted to all modules for chicken-and-eggy reasons. Besides the fact that sibling builds only caused obscure errors for me so far (because it detects siblings modules in unexpected locations when they are supposed to be not detected at all), this is another reason to consider removing sibling builds IMO, it is **not** an argument to reduce the number of modules.

PR #283, opened 2018-02-07
removing the completions in inactive cells after createGrid() in eclbasegridmanager

GitPaean:
reason is that we need to remove some completions in the Well in the Schedule due to processed grid in opm-simulators.

### DIFF
--- a/ebos/eclbasegridmanager.hh
+++ b/ebos/eclbasegridmanager.hh
@@ -194,6 +194,7 @@ public:
         }
 
         asImp_().createGrids_();
+        asImp_().filterCompletions_();
         asImp_().finalizeInit_();
     }
 
@@ -216,6 +217,10 @@ public:
     { return *eclState_; }
 
     const Opm::Schedule& schedule() const {
+        return *schedule_;
+    }
+
+    Opm::Schedule& schedule() {
         return *schedule_;
     }
 
@@ -316,7 +321,7 @@ private:
     static Opm::Deck* externalDeck_;
     static Opm::EclipseState* externalEclState_;
     static Opm::Schedule* externalSchedule_;
-    static Opm::SummaryConfig* externalSummaryConfig_; 
+    static Opm::SummaryConfig* externalSummaryConfig_;
     std::unique_ptr<Opm::Deck> internalDeck_;
     std::unique_ptr<Opm::EclipseState> internalEclState_;
     std::unique_ptr<Opm::Schedule> internalSchedule_;
--- a/ebos/eclcpgridmanager.hh
+++ b/ebos/eclcpgridmanager.hh
@@ -31,6 +31,7 @@
 #include "ecltransmissibility.hh"
 
 #include <dune/grid/CpGrid.hpp>
+#include <dune/grid/cpgrid/GridHelpers.hpp>
 
 #include <dune/grid/common/mcmgmapper.hh>
 
@@ -255,6 +256,16 @@ protected:
         equilCartesianIndexMapper_ = new CartesianIndexMapper(*equilGrid_);
 
         globalTrans_ = nullptr;
+    }
+
+    // removing some completions located in inactive grid cells
+    void filterCompletions_()
+    {
+        assert(grid_);
+        Grid grid = *grid_;
+        grid.switchToGlobalView();
+        const auto eclipseGrid = Opm::UgGridHelpers::createEclipseGrid(grid, this->eclState().getInputGrid());
+        this->schedule().filterCompletions(eclipseGrid);
     }
 
     Grid* grid_;
--- a/ebos/eclalugridmanager.hh
+++ b/ebos/eclalugridmanager.hh
@@ -196,6 +196,11 @@ protected:
             new CartesianIndexMapper(*grid_, cartesianDimension_, cartesianCellId_);
     }
 
+    void filterCompletions_()
+    {
+        // not handling the removal of completions for this type of grid yet.
+    }
+
     Grid* grid_;
     EquilGrid* equilGrid_;
     std::vector<int> cartesianCellId_;
--- a/ebos/eclpolyhedralgridmanager.hh
+++ b/ebos/eclpolyhedralgridmanager.hh
@@ -151,6 +151,11 @@ protected:
         cartesianIndexMapper_ = new CartesianIndexMapper(*grid_);
     }
 
+    void filterCompletions_()
+    {
+        // not handling the removal of completions for this type of grid yet.
+    }
+
     GridPointer grid_;
     CartesianIndexMapperPointer cartesianIndexMapper_;
 };


andlaus:
can't you just do this in the grid manager itself? it is very dangerous to modify the grid after the problem has been instantiated.

andlaus:
okay, I see, it is "only" the schedule. still, doing this as a bolt-on job seems to be quite inelegant. I'll merge if it is the best option, though. is it?

atgeirr:
I think it is the simplest way to allow modification of Schedule, given that the class already has a somewhat complex "either I manage this on my own or I just point to an external Schedule" interior treatment. The proposal hear works fine with either, without any extra complexity. One could argue however that there should be a `setExternalSchedule` method that could be used instead. That would be more general, but less efficient for our current use case (which is just about throwing away some invalid completions).

atgeirr:
...another point: it could be better in the long run to use single shared_ptr instead of "unique_ptr or raw pointer" and explicitly require shared_ptr in the interface (setExternal...), that would make optimization workflows built on "calling Flow" easier to deal with I think, and ensure consistency somewhat better. Not critical now though.

GitPaean:
> okay, I see, it is "only" the schedule. still, doing this as a bolt-on job seems to be quite inelegant. I'll merge if it is the best option, though. is it?

I am not sure it is the `best`.  It is just one simple and straightforward way, right before we need to use the schedule to create all the well related stuff. 

I am not sure before that, did we use `Schedule` to do anything in the `Simulator` or `GridManager`? Or more specifically, the wells? If yes, maybe we should do this before that.

andlaus:
I'd prefer a setter method because then the grid manager is at least aware that the schedule has potentially changed. This assumes that rewriting the schedule is not happening too often and that copying scheduling objects is possible. is it?

andlaus:
> ...another point: it could be better in the long run to use single shared_ptr instead of "unique_ptr or raw pointer" and explicitly require shared_ptr in the interface (setExternal...), that would make optimization workflows built on "calling Flow" easier to deal with I think, and ensure consistency somewhat better. Not critical now though.

no objections from my side!

andlaus:
> I am not sure before that, did we use Schedule to do anything in the Simulator or GridManager? Or more specifically, the wells? If yes, maybe we should do this before that.

it is used by the problem to determine the time step sizes, detect "earthquakes", etc. I'm just a bit worried that being able to modify the engine of the car while driving can have weird effects...

GitPaean:
> can't you just do this in the grid manager itself?

Another suggestion is that we do it in the constructor of  `EclBaseGridManager`, as follows, 

```
        asImp_().createGrids_();
        \\ remove the completions in inactive grid cells
        asImp_().filterCompletions_();
        asImp_().finalizeInit_();

```

Please let me know if this is a better solution to you.

andlaus:
that would definitely qualify as "better option" ;)

ebos/eclbasegridmanager.hh
```diff
@@ -311,6 +314,13 @@ private:
     const Implementation& asImp_() const
     { return *static_cast<const Implementation*>(this); }
 
+    void filterCompletions_()
+    {
+        asImp_().grid().switchToGlobalView();
```
GitPaean: I did not find the what is the default or current value of the `current_view_data_` in the grid, so I am not sure if I should swich/recover the `current_view_data_` here.  Any suggestion here?

GitPaean:
jenkins build this please

GitPaean:
jenkins build this please

ebos/eclbasegridmanager.hh
andlaus: looks good, just one thing: you must move this method to the `EclCpGridManager` (and amend the other grid managers with empty `filterCompletions_()` methods. this is required because `switchToGlobalView()` is `CpGrid` specific.

for your actual question: I don't know, @blattms probably does. maybe the grid should be copied (which is apparently for `CpGrid`) and `switchToGlobalView()` be called on the copy.

GitPaean:
@akva2 it looks like I do not have the privilege for jenkins in ewoms.

akva2:
@GitPaean granted.

ebos/eclbasegridmanager.hh
GitPaean: > this is required because switchToGlobalView() is CpGrid specific.

How we should handle `filterCompletions_` in other types of grid, especially in parallel. like AluGrid? Does it have the same problem in term of some grid cells can be changed to inactive during processing?

GitPaean:
> @GitPaean granted.

Thanks. @akva2

GitPaean:
jenkins build this please

ebos/eclbasegridmanager.hh
andlaus: > How we should handle filterCompletions_ in other types of grid, especially in parallel.

I'm not sure how this should be done. probably it would involve the [cartesianIndex()](https://github.com/OPM/ewoms/blob/master/ebos/eclbasegridmanager.hh#L255) or the `CartesianIndexMapper` directly. maybe @dr-robertk or @blattms have some insights.

short term, I think that it is sufficient if this functionality is not supported on non-`CpGrid`s.

ebos/eclcpgridmanager.hh
```diff
@@ -257,6 +258,14 @@ protected:
         globalTrans_ = nullptr;
     }
 
+    // removing some completions located in inactive grid cells
+    void filterCompletions_()
+    {
+        grid_->switchToGlobalView();
```
GitPaean: It turned out that `switchToGlobalView()` does not do anything here. Before this is already under global view.

But I am not sure if I should remove this line, since we can not guarantee it will be under global view. Without an interface to know the current view, it is really difficult to say what is the best way to do here. I will suggest to add a interface in CpGrid to access the global view directly.

andlaus: if `filterCompletions_()` is called before `loadBalance()` it is guaranteed that the grid has not yet been distributed. If we want to do simulations involving multi-million cell grids in the future, this design may fall on our feet, but let's burn that bridge when we come to it.

the final question is if `Opm::UgGridHelpers::createEclipseGrid()` is available for non-`CpGrid`s...

andlaus:
I'm sufficiently happy to push Green. @GitPaean: do you want do some changes or is it finished from your point of view?

ebos/eclcpgridmanager.hh
GitPaean: > the final question is if Opm::UgGridHelpers::createEclipseGrid() is available for non-CpGrids...

I think it should be done eventually.

GitPaean:
> I'm sufficiently happy to push Green. @GitPaean: do you want do some changes or is it finished from your point of view?

Let us wait a little from the comments from @blattms or @bska .

blattms:
The only safe and sound way to request a view without infecting other code is.
```
do_some_something_on_global_view(CpGrid& grid){
  auto myview = grid; // Copies two pointers
  grid.switchToGlobalView();
  // do stuff
}

**Comment by GitPaean** [comment]
> The only safe and sound way to request a view without infecting other code is.
```
do_some_something_on_global_view(CpGrid& grid){
  auto myview = grid; // Copies two pointers
  grid.switchToGlobalView();
  // do stuff
}

Thanks for the suggestion.

GitPaean:
jenkins build this please

GitPaean:
I am done updating. Merge if you are happy.

atgeirr:
I think you mean `myview.switchToGlobalView()` rather than `grid.....`?
```
do_some_something_on_global_view(CpGrid& grid){
  auto myview = grid; // Copies two pointers
  grid.switchToGlobalView();
  // do stuff
}
```

andlaus:
just to be safe:

andlaus:

ebos/eclcpgridmanager.hh
```diff
@@ -257,6 +258,16 @@ protected:
         globalTrans_ = nullptr;
     }
 
+    // removing some completions located in inactive grid cells
+    void filterCompletions_()
+    {
+        assert(grid_);
+        Grid grid = *grid_;
+        grid.switchToGlobalView();
+        const auto eclipse_grid = Opm::UgGridHelpers::createEclipseGrid(grid, this->eclState().getInputGrid());
```
andlaus: not a blocker, but call this variable `eclipseGrid` please.

andlaus: and it is probably slightly better to make it a const reference. The compiler will determine whether a copy is needed or not the...

GitPaean: Sure. I think it is good to keep consistent style with the current file. But in my opinion, make the variables and functions same style can be confusing sometimes and make the choice of the names more limited.

andlaus: that was quick, thanks. IMO variables and functions() are sufficiently distinguished by the brackets. I agree that consistency with the remaining code is more important, though.

ebos/eclcpgridmanager.hh
```diff
@@ -257,6 +258,16 @@ protected:
...
     }
 
+    // removing some completions located in inactive grid cells
+    void filterCompletions_()
+    {
+        assert(grid_);
+        Grid grid = *grid_;
+        grid.switchToGlobalView();
+        const auto eclipseGrid = Opm::UgGridHelpers::createEclipseGrid(grid, this->eclState().getInputGrid());
+        this->schedule().filterCompletions(eclipse_grid);
```
andlaus: with the rename, Jenkins complains here

GitPaean: Fixed. My bad. I should have made a local build before pushing.

GitPaean:
jenkins build this please

andlaus:
now that Jenkis is happy, let's merge!

PR #284, opened 2018-02-08
Some more cleaning in the output code in opm-output, ewoms and opm-simulator

totto82:
1) Communicate well results ( don't depend on the legacy code for this) 
2) Restore async output capability 
3) Pass RFT output via data::wells not data::cells. Allows for cleaner code and less unnecessary communication 

Depends on OPM/opm-output#250
Needs downstream modification in opm-simulator.


ebos/ThreadHandle.hpp
```diff
@@ -0,0 +1,188 @@
...
+#include <dune/common/exceptions.hh>
+
+#include <thread>
+#include <mutex>
+#include <queue>
+
+namespace Opm
+{
+
+  class ThreadHandle
```
andlaus: it would probably be good to add some documentation comments to this class, move it to a separate file and to remove one level of indentation.

ebos/ThreadHandle.hpp
```diff
@@ -0,0 +1,188 @@
...
+      void run () { }
+      bool isEndMarker () const { return true; }
+    };
+
+    ////////////////////////////////////////////
+    // class ThreadHandleQueue
+    ////////////////////////////////////////////
+    class ThreadHandleQueue
+    {
+    protected:
```
andlaus: same as above. (what is the purpose of this class?) Also, this `protected` block should better be moved to after the `public` one.

ebos/collecttoiorank.hh
```diff
@@ -292,6 +297,8 @@ namespace Ewoms
                         Opm::data::Solution& globalCellData,
                         const std::map<std::pair<std::string, int>, double>& localBlockData,
                         std::map<std::pair<std::string, int>, double>& globalBlockValues,
+                        const Opm::data::Wells& localWellData,
+                        Opm::data::Wells& globalWellData,
```
andlaus: hm,  for my taste the argument list of that function is too long even before this patch. any ideas/thoughts?

ebos/ThreadHandle.hpp
```diff
@@ -0,0 +1,188 @@
+#ifndef OPM_THREADHANDLE_HPP
```
andlaus: the copyright boiler plate is missing. Also, the file should be called "threadhandle.hh" if it does not want to be the lone wolf in eWoms ;) (yeah, the file naming convention should be unified with the other OPM modules, but that's work for another PR.)

tests/test_ecl_output.cpp
```diff
@@ -174,6 +174,69 @@ void test_summary()
     CHECK_CLOSE(roip2, ecl_sum_get_general_var( resp, 1, "ROIP:2" ), 1e-3 );
```
andlaus: that file should probably be renamed to `test_ecl_output.cc`

ebos/ThreadHandle.hpp
totto82: This file is a pure copy from opm-simulator.

atgeirr:
Merge conflict must be resolved.

PR #287, opened 2018-02-14
Add adjoint

hnil:
A test pull request for adding adjoint. The the most critical is if it degrades performance of normal run. Most relevant changes for this is the change for calculating storage term with storage and the adding of the focustimeidx.  This is related to changes in opm simulator.
All test are still not runned.


andlaus:
@hnil: what is the status here? do you want it to be reviewed for merging?

atgeirr:
Replaced by #371.

PR #288, opened 2018-02-19
Fix restart norne

totto82:
TODO
 
- [x] Test it on the big models 
- [x] Verify that this fix Odd's issue.

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -565,7 +565,7 @@ public:
         }
 
         bool doInvalidate = updateHysteresis_();
-        doInvalidate = doInvalidate || updateMaxOilSaturation_();
+        doInvalidate = updateMaxOilSaturation_() || doInvalidate;
 
         if (GET_PROP_VALUE(TypeTag, EnablePolymer))
             updateMaxPolymerAdsorption_();
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -564,8 +564,9 @@ public:
             simulator.setTimeStepSize(dt);
         }
 
-        bool doInvalidate = updateHysteresis_();
-        doInvalidate = updateMaxOilSaturation_() || doInvalidate;
+        const bool invalidateFromHyst = updateHysteresis_();
+        const bool invalidateFromMaxOilSat = updateMaxOilSaturation_();
+        const bool doInvalidate = invalidateFromHyst || invalidateFromMaxOilSat;
 
         if (GET_PROP_VALUE(TypeTag, EnablePolymer))
             updateMaxPolymerAdsorption_();
--- a/ewoms/models/blackoil/blackoilprimaryvariables.hh
+++ b/ewoms/models/blackoil/blackoilprimaryvariables.hh
@@ -267,9 +267,11 @@ public:
 
         bool gasPresent = (fluidState.saturation(gasPhaseIdx) > 0.0);
         bool oilPresent = (fluidState.saturation(oilPhaseIdx) > 0.0);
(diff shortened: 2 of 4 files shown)


totto82:
The PR changes convergence behavior on Norne, but results and performance are identical in eye-ball norm. This PR has no effect on Model 2.

ebos/eclproblem.hh
```diff
@@ -559,7 +559,7 @@ public:
         }
 
         bool doInvalidate = updateHysteresis_();
-        doInvalidate = doInvalidate || updateMaxOilSaturation_();
+        doInvalidate = updateMaxOilSaturation_() || doInvalidate;
```
atgeirr: This looks like not-really-a-change?

ewoms/models/blackoil/blackoilprimaryvariables.hh
```diff
@@ -267,9 +267,11 @@ public:
 
         bool gasPresent = (fluidState.saturation(gasPhaseIdx) > 0.0);
         bool oilPresent = (fluidState.saturation(oilPhaseIdx) > 0.0);
+        static const Scalar thresholdWaterFilledCell = 1.0 - 1e-6;
```
atgeirr: This makes me nervous as always... I assume this is the main reason for the convergence changes?

totto82: Yes. As we discussed on Friday I think the best solution is to store the  primary variable meaning in the restart file. Then we get rid of this line for code.

ebos/eclproblem.hh
totto82: Is the second function in || called if the first is true?

atgeirr: No, and that can be used to optimize away calls. I have complained before about these methods that both do something and return a bool to indicate if they invalidation...

Should updateMaxOilSaturation_() always be called? If so, the old version was buggy (it would not be called if updateHysteresis_() returned true). If not, the new version gives (slightly) worse performance as it calls it unnecessarily.

Assuming that it is correct in this PR, I would personally prefer:
```
const bool invalidateFromHyst = updateHysteresis_();
const bool invalidateFromMaxOilSat = updateMaxOilSaturation_();
const bool doInvalidate = invalidateFromHyst || invalidateFromMaxOilSat;
```
Done that way it is clear that all functions are called, and as a bonus the bools can be const.

totto82: As I understand the code, both most be called, and that is the reason I changed the order. This does affect the norne model, but I didn't notice any significant effect on the results. 
I agree with your personal preference, I just accidentally noticed this while debugging something els. I don't think this change is related to fixing the restart. If you think this needs to be tested more, I will suggest to remove this from this PR and make another one just for this. I will update the code with your suggestions if you think it is sufficiently tested. As I reported earlier this PR has no effect on Model 2 (models without Vappars)

atgeirr: In general I prefer a separate PR, but given that the interaction of this bug (which may or may not be significant at all) with the other changes here may be nontrivial, I think it is ok to do it as part of this PR (including testing).

atgeirr:
jenkins build this please

atgeirr:
jenkins build this please

ebos/eclproblem.hh
atgeirr: This can be merged now, but I had the impression that you wanted to do the minor rewrite of this section first?

totto82: Yes. Give me a minute or two.

totto82:
jenkins build this please

atgeirr:
All green, merging!

PR #290, opened 2018-02-19
Rename test-deck to get consistent case

joakim-hove:
OK: In addition to the description below there seems to have been a change in behaviour related to output path. That is also "fixed" in this PR.

In eclipse all filenames (not include files) must have a fixed case; either lowercase: `eclipse_simulation.data` or alternatively `ECLIPSE_SIMULATION.DATA`. The same case convention will also apply to all result files from this simulation.

Following this PR in libecl: https://github.com/Statoil/libecl/pull/308 the behaviour with respect to case awareness is stricter, and more aligned with Eclipse. If you give full path a file the libecl implementation does not care at all - and you can in principle load a grid from the filename `MixedCase.SMSPEC` - if you are so inclined. But - when using the functions which start with a basename and look in the filesystem for files with the correct extension - like:
```C
ecl_sum = ecl_sum_fread_alloc("path/simulation/CASE1")
```
the case of filenames must be according to eclipse expectations, otherwise the search will fail. The which creates an eclipse summary case has such a mixed case deck name, and when subsequently loading the case it fails - and the whole test fails.


joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

joakim-hove:
@andlaus / @atgeirr : I think this is actually keeping Jenkins hostage; I can not get jenkins to run in this repository. Good if you invoke jenkins, and merge it if you agree with the solution.

atgeirr:
jenkins build this please

atgeirr:
jenkins build this please

joakim-hove:
OK @atgeirr : one more time please?

atgeirr:
jenkins build this please

joakim-hove:
OK - jenkins green; merging.

PR #296, opened 2018-03-07
fixed: fix some condition that was missed in eclipse io update

akva2:
woops...

PR #297, opened 2018-03-08
changing the PVTO in test_ecl_output

GitPaean:
to make the table passing the sanity check.


GitPaean:
closing it since it is included in #296

PR #302, opened 2018-03-19
write the VTK output asynchronously by default for sequential simulations

andlaus:
`Dune::VTKWriter` seems to be on the rather slow side, so threaded output writing leads to a sizeable speedup for the VTK format. For situations where an extra thread is not desireable, asynchronous VTK
output can be disabled using the `--enable-async-vtk-output=false` command-line parameter.

The main complication of this is that `Dune::VTKWriter` cannot be used in conjunction with MPI-parallel simulations because there are unnecessary calls to `MPI_barrier()` in `Dune::VTKWriter::pwrite()`
which lead to all kinds of fun since multiple threads are doing uncoordinated MPI calls in that case. (This is because the main thread needs to call MPI functions during e.g. the linear solver.) That said,
for the MPI-parallel simulations which I've tested, there was no speedup either, so asynchronous output is simply disabled for the MPI case.

### DIFF
--- a/ewoms/io/vtkmultiwriter.hh
+++ b/ewoms/io/vtkmultiwriter.hh
@@ -33,6 +33,7 @@
 #include "vtktensorfunction.hh"
 
 #include <ewoms/io/baseoutputwriter.hh>
+#include <ewoms/parallel/tasklets.hh>
 
 #include <opm/material/common/Valgrind.hpp>
 #include <opm/material/common/Unused.hpp>
@@ -62,6 +63,31 @@ namespace Ewoms {
 template <class GridView, int vtkFormat>
 class VtkMultiWriter : public BaseOutputWriter
 {
+    class WriteDataTasklet : public TaskletInterface
+    {
+    public:
+        WriteDataTasklet(VtkMultiWriter& multiWriter)
+            : multiWriter_(multiWriter)
+        { }
+
+        void run() final
+        {
+            std::string fileName;
+            // write the actual data as vtu or vtp (plus the pieces file in the parallel case)
+            fileName = multiWriter_.curWriter_->write(/*name=*/multiWriter_.curOutFileName_.c_str(),
+                                                      static_cast<Dune::VTK::OutputType>(vtkFormat));
+
+            // determine name to write into the multi-file for the
+            // current time step
+            multiWriter_.multiFile_.precision(16);
+            multiWriter_.multiFile_ << "   <DataSet timestep=\"" << multiWriter_.curTime_ << "\" file=\""
+                                    << fileName << "\"/>\n";
+        }
+
+    private:
+        VtkMultiWriter& multiWriter_;
+    };
+
     enum { dim = GridView::dimension };
 
 #if DUNE_VERSION_NEWER(DUNE_GRID, 2,6)
@@ -87,7 +113,8 @@ public:
     typedef typename VtkWriter::VTKFunctionPtr FunctionPtr;
 #endif
 
-    VtkMultiWriter(const GridView& gridView,
+    VtkMultiWriter(bool asyncWriting,
+                   const GridView& gridView,
                    const std::string& simName = "",
                    std::string multiFileName = "")
         : gridView_(gridView)
@@ -98,6 +125,9 @@ public:
         , elementMapper_(gridView)
         , vertexMapper_(gridView)
 #endif
+        , curWriter_(nullptr)
+        , curWriterNum_(0)
+        , taskletRunner_(/*numThreads=*/asyncWriting?1:0)
     {
         simName_ = (simName.empty()) ? "sim" : simName;
         multiFileName_ = multiFileName;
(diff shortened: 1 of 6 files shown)


andlaus:

andlaus:
@blattms, @dr-robertk: the proper way to fix the MPI issue mentioned above is to do it in dune. if you get a fix merged to Dune, I'll do the necessary changes in eWoms.

atgeirr:
> asynchronous output is simply disabled for the MPI case.

I assume that this is true for VTK output only? So ECL output can still be async?

andlaus:
ECL and VTK output are completely unrelated. That said, since -- in contrast to VTK -- the ECL format does not is not support distributed writing, there are other issues related to the fact that everything must be collected on a single rank, though.

ewoms/io/vtkmultiwriter.hh
```diff
@@ -525,6 +558,8 @@ private:
 
     std::list<ScalarBuffer *> managedScalarBuffers_;
```
atgeirr: I get that this part is not new code, but is there a reason why raw (but still owned it seems) pointers are used here and for curWriter_ rather than some smart pointer? Is the releaseBuffers_() method a bugfix?

andlaus: you are right: `smart_ptr` is probably a good idea. I don't remember why it was done like this. stuff for another PR...

andlaus:
ping. mind to press Green?

atgeirr:
Reading, give me a minute...

ewoms/io/vtkmultiwriter.hh
```diff
@@ -107,13 +135,16 @@ public:
         }
 
         curWriterNum_ = 0;
+        curWriter_ = nullptr;
```
atgeirr: Should be in init list for better exception safety.

ewoms/io/vtkmultiwriter.hh
```diff
@@ -98,6 +125,7 @@ public:
         , elementMapper_(gridView)
         , vertexMapper_(gridView)
 #endif
+        , taskletRunner_(/*numThreads=*/asyncWriting?1:0)
```
atgeirr: What happens with 0 requested threads? Synchronous execution upon dispatch()?

andlaus: yes, synchronous execution.

ewoms/parallel/tasklets.hh
```diff
@@ -138,7 +138,7 @@ public:
         runnerMutex_.lock();
 
         threads_.resize(numWorkers);
-        for (int i = 0; i < numWorkers; ++i)
+        for (unsigned i = 0; i < numWorkers; ++i)
```
atgeirr: Range-for instead of messing wiht indices?

atgeirr:
Looks good, only thing that *must* be fixed is the init list thing, change the for loop commented on if you agree it is better.

ewoms/io/vtkmultiwriter.hh
andlaus: if an exception is thrown in a constructor, the simulator will crash anyway. that's also a reason why there is a two-stage init (i.e., `finalizeInit()` in the problem)

ewoms/parallel/tasklets.hh
andlaus: that would be the first occurrence of range-based for in opm-material and ewoms and probably all of OPM. I think that if it is decided that range-based for should be used, it should be used whenever possible. IMO that should be a higher-level decision and there should be an concerted cleanup effort to convert all locations that make sense.

andlaus:
> only thing that must be fixed is the init list thing

Since it seems impossible that this constructor raises an exception, I cannot see a reason why it should make a difference technically. Anyway, changed.

ewoms/parallel/tasklets.hh
atgeirr: It is not the first in OPM, certainly! In fact it is used all over the place, we agreed a long time ago that was ok (around the time we made gcc 4.8 the minimum compiler perhaps).

The code base is too large to refactor for this needlessly, in my opinion. But when changing code (or of course writing new code) we should make use of the opportunity to use range-for.

atgeirr:
> Since it seems impossible that this constructor raises an exception, I cannot see a reason why it should make a difference technically. Anyway, changed.

An exception could be thrown from here for example:
```
 multiFileName_ = multiFileName;
```

In any case, using the initialization list is the preferred way to do it, so it is good to make the change anyway I think.

atgeirr:
Looks good now, jenkins and merge if green!

atgeirr:
jenkins build this please

ewoms/parallel/tasklets.hh
andlaus: I would not call this change a "refactoring", rather a "cleanup". Anyway, if this is wanted by @totto82 I'll invest a few days after the release into converting opm-material and eWoms.

andlaus:
weird: jenkins says "failed", yet I was unable to find any error message. let's try again.

andlaus:
jenkins build this please

andlaus:
no changes, but jenkins is happy. I propose you press Green...

atgeirr:
If you take a look at Jenkins here: https://ci.opm-project.org/job/ewoms-PR-builder/161/
... you'll see that there was a timeout.

Anyway: merging!

PR #301, opened 2018-03-15
generalize the tasklet mechanism to an arbitrary number of worker threads

andlaus:
this has been quite a bit more complicated than I initially thought.

PR #299, opened 2018-03-12
ECL output cleanup and tasklets

andlaus:
this PR does a smallish cleanup of the ECL output code and introduces "tasklets", a mechanism to run jobs asynchronously. the tasklet code is based on the asynchronous ECL output code, but it has been substantially cleaned up:

 - it does not do no polling anymore
 - it supports barriers
 - the terminology has been improved and generalized
 - the code is now properly documented

so far, only a single worker thread is used, but changing this is trivial from the tasklet runner's point of view. (The tasklets themselfs are a different story.)

this PR needs a small mop-up patch in opm-simulators.

PR #303, opened 2018-03-21
disable asynchronous output for the ECL output test

andlaus:
this lead to non-deterministic crashes deep inside libecl. My cursory hypotheses are that this test makes the assumption that the output is written synchronously (it tries to read back the results from disk immediately) and/or that libecl is not threadsafe.

### DIFF
--- a/tests/test_ecl_output.cc
+++ b/tests/test_ecl_output.cc
@@ -79,6 +79,7 @@ namespace Ewoms {
 namespace Properties {
 NEW_TYPE_TAG(TestEclOutputTypeTag, INHERITS_FROM(BlackOilModel, EclBaseProblem));
 SET_BOOL_PROP(TestEclOutputTypeTag, EnableGravity, false);
+SET_BOOL_PROP(TestEclOutputTypeTag, EnableAsyncEclOutput, false);
 }}
 
 static const int day = 24 * 60 * 60;


andlaus:
jenkins build this please

joakim-hove:
> it tries to read back the results from disk immediately

Well - that is certainly the case yes.

> and/or that libecl is not threadsafe.

The library itself has no global state - so it is possible make thread safe applications based on it, but of course the filesystem is a form of global state.

atgeirr:
I think async output behaviour is quite important to test, and here we have been given a crashing case for free...! I would rather like to see the test deal properly with async output, I guess it should not be too hard given the new Tasklet framework? Or would it require adding something there?

andlaus:
well, the test did not fail *deterministically*. depending on the system you ran it on, it failed maybe once out of 5 times. I suppose that's part of the fun with threads.

the problem is that the code that checks that the written results are correct is in the main thread immediately after the writing tasklet has been dispatched. I am thus pretty sure that it is not worthwhile to jump through hoops for this test (e.g. a call to `barrier()` can be added which makes it quasi-synchronous again, or the comparison can be done in a tasklet), but I'm open if you'll refactor it.

atgeirr:
> well, the test did not fail deterministically. depending on the system you ran it on, it failed maybe once out of 5 times. I suppose that's part of the fun with threads.

If you add a one-second delay I guess you could get it to fail every time?

> I am thus pretty sure that it is not worthwhile to jump through hoops for this test

On the contrary, this is the best test you could get to prove that the tasklets are up to the ... uh... task (let).

andlaus:
> If you add a one-second delay I guess you could get it to fail every time?

there is no guarantee for that. also, I'm pretty sure that there would be quite a few complaints if writing output would be delayed by 1 second for each timestep.

> On the contrary, this is the best test you could get to prove that the tasklets are up to the ... uh... task (let).

a dedicated test for tasklets would be a better idea as it could e.g. test the case with more than one worker thread. do you want do implement one? This would probably also be a good reason to get familiar with the code.

andlaus:
anyway, I think this PR should be merged even if you just consider it a band-aid. Mind to press green?

joakim-hove:
> If you add a one-second delay I guess you could get it to fail every time?

I have not seen the test in question - but I would assume the correct &trade; approach was to have some join() like call before reading and verifying the generated result file?

andlaus:
> I have not seen the test in question - but I would assume the correct ™ approach was to have some join() like call before reading and verifying the generated result file?

either that (this would imply deleting the simulator object) or triggering a barrier on the tasklet runner...

andlaus:
>> I'm pretty sure that there would be quite a few complaints if writing output would be delayed by 1 second for each timestep.
>
> Sure, it was just to get high failure rates while improving to code. I did not intend it as a permanent feature!

I forgot to mention that I did exactly what you proposed before opening the tasklet PRs (#299 and #301). it was not a permanent feature, though ;)

PR #304, opened 2018-03-29
possibly fix a race condition in the tasklet mechanism

andlaus:
if `barrierCondition_.notify_all()` happened while another thread was preparing to call `barrierCondition_.wait()`, the preparing thread would miss the notification and `barrierCondition_.wait()` would lock up. after this patch, this should not happen anymore because the whole wait() method of the barrier tasklet is operated with the barrier mutex locked. This has the added benefit of making the code a bit simpler.

this race only occured if condition variables do not check the predicate before waiting. My references are not very explicit about this.

### DIFF
--- a/ewoms/parallel/tasklets.hh
+++ b/ewoms/parallel/tasklets.hh
@@ -85,18 +85,18 @@ class TaskletRunner
 
         void wait()
         {
-            numWaitingMutex_.lock();
-            numWaiting_ += 1;
-            if (numWaiting_ >= numWorkers_ + 1)
-                barrierCondition_.notify_all();
-            numWaitingMutex_.unlock();
+            std::unique_lock<std::mutex> lock(barrierMutex_);
 
-            if (numWaiting_ < numWorkers_ + 1) {
+            numWaiting_ += 1;
+            if (numWaiting_ >= numWorkers_ + 1) {
+                lock.unlock();
+                barrierCondition_.notify_all();
+            }
+            else {
                 const auto& areAllWaiting =
                     [this]() -> bool
                     { return this->numWaiting_ >= this->numWorkers_ + 1; };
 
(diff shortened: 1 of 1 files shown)


andlaus:
jenkins build this please

andlaus:
any comments?

atgeirr:
> any comments?

I will, but need a little bit of time.

atgeirr:
I am reading this class quite carefully now. My preliminary conclusion is that it *definitely* needs unit testing! While not extremely long, the behavioural complexity is large. We have already seen two errors (including the one addressed here) that should become tests. Please add such testing to this PR or in a separate PR.

andlaus:
> However, I think the current code is not resistant to spurious wakeups, see http://en.cppreference.com/w/cpp/thread/condition_variable for a reference on wait().

Based on the same document, I'm pretty sure that it is because that's the whole purpose of the predicate. also, be aware that the mutex is locked after `wait()`.

atgeirr:
> Based on the same document, I'm pretty sure that it is because that's the whole purpose of the predicate. also, be aware that the mutex is locked after wait().

The wait() method *unlocks* the mutex and suspends the thread until the predicate is true, but it is possible to wake up with the predicate still false, which is why we should handle the situation.

andlaus:
> The wait() method unlocks the mutex and suspends the thread until the predicate is true

almost: it checks the predicate and if it is false it atomically unlocks the mutex and goes to sleep until notified or spuriously woken up. if woken up, it locks the mutex atomically again, and checks the predicate and repeats the whole excercise if it is still false. Welcome to threaded programming.

atgeirr:
Upon rereading the docs I agree that the predicate version of wait() does not need to be wrapped in a `while` loop.

atgeirr:
jenkins build this please

atgeirr:
I'll merge this now, I made an issue of the testing requirement.

PR #307, opened 2018-04-05
use Opm::isfinite() instead of std::isfinite(Opm::value())

andlaus:
besides being briefer, this also ensures that derivatives are finite.

### DIFF
--- a/ewoms/disc/common/fvbaselocalresidual.hh
+++ b/ewoms/disc/common/fvbaselocalresidual.hh
@@ -425,7 +425,7 @@ protected:
         size_t numDof = elemCtx.numDof(/*timeIdx=*/0);
         for (unsigned i=0; i < numDof; i++) {
             for (unsigned j = 0; j < numEq; ++ j) {
-                assert(std::isfinite(Toolbox::value(residual[i][j])));
+                assert(Opm::isfinite(residual[i][j]));
                 Opm::Valgrind::CheckDefined(residual[i][j]);
             }
         }
--- a/ewoms/models/blackoil/blackoilintensivequantities.hh
+++ b/ewoms/models/blackoil/blackoilintensivequantities.hh
@@ -339,14 +339,14 @@ public:
             if (!FluidSystem::phaseIsActive(phaseIdx))
                 continue;
 
-            assert(std::isfinite(Toolbox::value(fluidState_.density(phaseIdx))));
-            assert(std::isfinite(Toolbox::value(fluidState_.saturation(phaseIdx))));
-            assert(std::isfinite(Toolbox::value(fluidState_.temperature(phaseIdx))));
-            assert(std::isfinite(Toolbox::value(fluidState_.pressure(phaseIdx))));
-            assert(std::isfinite(Toolbox::value(fluidState_.invB(phaseIdx))));
+            assert(Opm::isfinite(fluidState_.density(phaseIdx)));
+            assert(Opm::isfinite(fluidState_.saturation(phaseIdx)));
+            assert(Opm::isfinite(fluidState_.temperature(phaseIdx)));
+            assert(Opm::isfinite(fluidState_.pressure(phaseIdx)));
+            assert(Opm::isfinite(fluidState_.invB(phaseIdx)));
         }
-        assert(std::isfinite(Toolbox::value(fluidState_.Rs())));
-        assert(std::isfinite(Toolbox::value(fluidState_.Rv())));
+        assert(Opm::isfinite(fluidState_.Rs()));
+        assert(Opm::isfinite(fluidState_.Rv()));
 #endif
     }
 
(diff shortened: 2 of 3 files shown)


andlaus:
jenkins build this please

atgeirr:
Isn't the correct way to do this:
```
using std::isfinite; // So the next line works fine for doubles
assert(isfinite(x)); // If x is something in namespace Opm, Opm::isfinite() will be called
```

andlaus:
No it isn't, it just a different way to use `std::isfinite()`: it will still not work with Evaluations, because eWoms is not using the `Opm` namespace and even if it would, I would consider the result an ugly and confusing hack.

Finally, be aware that `Opm::isfinite()` works with POD types just fine...

atgeirr:
For Evaluations, `isfinite()` should be implemented in the same namespace as the Evaluation class (I just assumed that was in Opm), then my suggestion will work. Furthermore, any later users who would like to use their own `FooNamespace::Number` class will have things just work, as long as they also define `FooNamespace::isfinite()`. This is why argument-dependent name lookup was invented in the first place! There is no `Opm::isfinite(FooNamespace::Number)` and there should not have to be one either.

We did at some point struggle with this in Dune, some Dune classes used `Dune::abs()` (I think, could have been another function), which meant we had to do some tricks to make it work, including in specific order etc. I don't want us to make the same mistake.

andlaus:
> For Evaluations, isfinite() should be implemented in the same namespace as the Evaluation class (I just assumed that was in Opm), then my suggestion will work.

it would still be very confusing, at least to dumbasses like me: `isfinite()` is a convenience function which works regardless of the type of the argument and is thus in the same namespace as the remaining convenience functions. let me think to remember who requested the convenience function feature ;).

If you want a version that only works for `Evaluation<>`, `Opm::DenseAd::isfinite()` is your friend.

>  Furthermore, any later users who would like to use their own FooNamespace::Number class will have things just work,

that's exactly the reason why you should use `Opm::isfinite()` over the other alternatives. (replace `FooNamespace` by `Ewoms` and you have the current situation.)

> which meant we had to do some tricks to make it work, including in specific order etc. I don't want us to make the same mistake.

the `using` statement coupled with ADL is just such a trick. IMO it is even a quite confusing one.

atgeirr:
> a convenience function which works regardless of the type of the argument

Not for new types.

> let me think to remember who requested the convenience function feature ;).

When you write it like that you seem to indicate I asked for them? I cannot recall that, but if so, I guess I have learned since then...!

> that's exactly the reason why you should use Opm::isfinite() over the other alternatives. (replace FooNamespace by Ewoms and you have the current situation.)

I don't think so, just to be clear: where is Opm::isfinite() defined? Just so I don't overlook something.

> the using statement coupled with ADL is just such a trick. IMO it is even a quite confusing one.

It's not a trick because:  a) it is idiomatic C++ (for example see the Note about the standard way to write swaps on http://en.cppreference.com/w/cpp/language/adl) b) it does not require headers to be included in a particular order

andlaus:
>> a convenience function which works regardless of the type of the argument
>
> Not for new types.

maybe. feel free to change most of opm-material, eWoms and large parts of opm-simulators.

Also the `using std` statement is very confusing because you suddenly can use classes from `std::` without prefixes. it is a price that is defintely not worth paying for a minor problem that has not been observed in practice, IMO.

>> the using statement coupled with ADL is just such a trick. IMO it is even a quite confusing one.
>
> It's not a trick because: a) it is idiomatic C++ (for example see the Note about the standard way to write swaps on http://en.cppreference.com/w/cpp/language/adl) b) it does not require headers to be included in a particular order

if you don't want to call it a "trick", how about "anti pattern"? at least it is very effective in disguising in what is actually used.

atgeirr:
> Also the using std statement is very confusing because you suddenly can use classes from std:: without prefixes.

Not at all. It's `using std::isfinite;`, not `using namespace std;`.

> maybe. feel free to change most of opm-material, eWoms and large parts of opm-simulators.

Why would doing it as I suggest require rewriting at all? To fully support new types, perhaps yes, but there is no reason to add *more* potential problems?

andlaus:
>>   Also the using std statement is very confusing because you suddenly can use classes from std:: without prefixes.
>
> Not at all. It's using std::isfinite;, not using namespace std;.

still, it is confusing: you're saying that you are using `std::isfinite()` but -- depending on the argument type -- you don't. have fun finding what's actually called without stepping through it with a debugger. I stick to the original point: `Opm::isfinite()` is the better alternative.

> Why would doing it as I suggest require rewriting at all?

because for opm-material and eWoms I insist on keeping the  code consistent: either do it *only* one way or *only* the other, but don't go for a worst-of-both-worlds mixed approach.

> To fully support new types, perhaps yes, but there is no reason to add more potential problems?

non-POD types are already fully supported, provided that all necessary functions are in the `Opm` namespace. given that -- besides `Opm::DenseAd::Evaluation` -- this a *very* exotic case and considering that much of the old OPM code does not even support anything different than `double`, I do not think that this is not  a matter which we should waste more time on discussing. Either you merge it, or I'll just close the PR. I'll do, as you prefer.

atgeirr:
> because for opm-material and eWoms I insist on keeping the code consistent: either do it only one way or only the other, but don't go for a worst-of-both-worlds mixed approach.

Well, I think it should be done the *other* way... However, I won't let the best stand in the way of the good, so if you *insist on this way or no way* I'll merge (I want the feature added -- testing the derivatives).

PR #309, opened 2018-04-10
Dissolve aux directory

### LINKED ISSUE #308: Cannot clone into Windows docker container: name "aux" is reserved in Windows
I am building OPM inside a Docker container running on my Windows laptop.  When I clone ewoms, it throws an error when trying to create the ewoms/aux directory.  It turns out that "aux" is a reserved string in Windows.

I fixed it by changing aux to aux_, but probably that is not the best solution.

andlaus:
this PR moves the contents of the `ewoms/aux` directory to more appropriate places. It also makes ebos compile with dune-fem on Dune 2.6. (on my machine older versions of dune-fem and dune do not build anymore due to cmake issues in dune-fem.)

this PR should fix #308. it requires a trivial mop-up for opm-simulators.


andlaus:
jenkins build this opm-simulators=1452 please

totto82:
Since Jenkins and @atgeirr is happy and we want this in the release. I will merge this and the opm-simulator patch now.

PR #311, opened 2018-04-12
ebos fix and compiler workaround

andlaus:
this PR contains two relative straight-forward changes: the first is a work around for a likely `constexpr` related compiler bug of older GCC versions, the second fixes https://github.com/OPM/opm-simulators/issues/1453. I think both should be part of the 2018.04 release.

thanks goes to @bska for finding the issue with initial temperature.

### DIFF
--- a/ewoms/models/blackoil/blackoilmodel.hh
+++ b/ewoms/models/blackoil/blackoilmodel.hh
@@ -145,8 +145,12 @@ private:
     static constexpr Scalar alpha = GET_PROP_VALUE(TypeTag, BlackoilConserveSurfaceVolume) ? 1000.0 : 1.0;
 
 public:
-    static constexpr Scalar value = 1.0/(30*4184.0*alpha);
+    typedef Scalar type;
+    static const Scalar value;
 };
+
+PROP_STATIC_CONST_MEMBER_DEFINITION_PREFIX_(BlackOilModel, BlackOilEnergyScalingFactor)
+    ::value = 1.0/(30*4184.0*alpha);
 
 // by default, ebos formulates the conservation equations in terms of mass not surface
 // volumes
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -659,8 +659,21 @@ public:
         }
 
         // we no longer need the initial soluiton
-        if (this->simulator().episodeIndex() == 0)
+        if (this->simulator().episodeIndex() == 0 && !initialFluidStates_.empty())  {
+            // we always need to provide a temperature and if energy is not conserved, we
+            // use the initial one. This means we have to "salvage" the temperature from
+            // the initial fluid states before deleting the array.
+            if (!enableEnergy) {
+                initialTemperature_.resize(initialFluidStates_.size());
+                for (unsigned i = 0; i < initialFluidStates_.size(); ++i) {
+                    const auto& fs = initialFluidStates_[i];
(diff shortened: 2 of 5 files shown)


andlaus:
jenkins build this please

ewoms/models/blackoil/blackoilmodel.hh
```diff
@@ -145,9 +145,13 @@ private:
     static constexpr Scalar alpha = GET_PROP_VALUE(TypeTag, BlackoilConserveSurfaceVolume) ? 1000.0 : 1.0;
 
 public:
-    static constexpr Scalar value = 1.0/(30*4184.0*alpha);
+    typedef Scalar type;
+    static const Scalar value;
 };
 
+PROP_STATIC_CONST_MEMBER_DEFINITION_PREFIX_(BlackOilModel, BlackOilEnergyScalingFactor)
+    ::value = 1.0/(30*4184.0*alpha);
```
atgeirr: While I can see that this number is not new code as such, it is quite cryptic! Could you write this using units? If not, at least make a comment that explains the constant?

andlaus:
I suppose this should go into the release as well?

atgeirr:
Yes, you can merge from my POV. However I'd still like to know where the cryptic constant comes from...

ewoms/models/blackoil/blackoilmodel.hh
andlaus: the unit system from opm-common is not available here, because the eWoms black-oil model is generic, i.e. is independent of the ECL I/O code.

the comment which you suggest is already here:

https://github.com/OPM/ewoms/blob/master/ewoms/models/blackoil/blackoilmodel.hh#L135

(it just documents the property, not the line itself)

atgeirr: Ok, I missed the comment since github cut off the diff display. Comment is fine.

I am a bit worried that you don't want to use units from opm-common here though, even in the black oil model (if not the ecl model).

andlaus: I'm open to use stuff from opm-common in a more general context, but only if it does not force people to deal with ECL file I/O by default and if no additional library is imposed upon people who are not interested in ECL simulations. (to see this from a different perspective. I'm pretty sure that you would object having to go through the pain of e.g. setting comsol or HDF I/O libs or python bindings only to not use them in the code which you're actually interested  in.) This is not going to happen in the near future, if ever...

andlaus:
@totto82: I think it's Green time.

ewoms/models/blackoil/blackoilmodel.hh
atgeirr: The basic Units.hpp is header only, and always available. It's only tie to ECL I/O is that it happens to reside in opm/parser/eclipse/Units/ (that could easily be changed). So I think it should be used here.

andlaus: > The basic Units.hpp is header only, and always available.

*exactly* this is the point: even though the needed functionality does not depend on the ECL I/O stuff at all, one is forced to jump through substantial hoops to use it because that stuff is located within opm-common. splitting opm-common into a "header only" module which features only self-contained functionality and into another which does ECL-I/O would do the trick, but that would kind-of defeat the goals of the recent opm-parser/opm-output/opm-common merger...

totto82:
I will merge this now, since we want this in the release. Maybe we should continue this discussion in oslo.

PR #312, opened 2018-04-19
Fix undefined mutex unlocking behaviour, add unit test for tasklets

andlaus:
this addresses some issues with undefined unlocking behavior in tasklets that somehow only seem to appear with MacOS.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -312,4 +312,5 @@ opm_add_test(obstacle_pvs_restart
 opm_add_test(tutorial1
              SOURCES tutorial/tutorial1.cc)
 
-opm_add_test(test_tasklets)
+opm_add_test(test_tasklets
+             DRIVER_ARGS --plain)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -311,3 +311,5 @@ opm_add_test(obstacle_pvs_restart
 
 opm_add_test(tutorial1
              SOURCES tutorial/tutorial1.cc)
+
+opm_add_test(test_tasklets)
--- a/ewoms/parallel/tasklets.hh
+++ b/ewoms/parallel/tasklets.hh
@@ -133,9 +133,6 @@ public:
      */
     TaskletRunner(unsigned numWorkers)
     {
-        // make sure that the worker threads block when the tasklet queue is empty
-        runnerMutex_.lock();
-
         threads_.resize(numWorkers);
         for (unsigned i = 0; i < numWorkers; ++i)
             // create a worker thread
@@ -182,10 +179,10 @@ public:
 
             // add the tasklet to the queue
             taskletQueue_.push(tasklet);
-            // fire up a worker thread
-            runnerMutex_.unlock();
 
             taskletQueueMutex_.unlock();
+
+            workAvailableCondition_.notify_all();
         }
     }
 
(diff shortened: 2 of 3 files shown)


andlaus:
the jenkins failure seems to be related to libecl!?

atgeirr:
> the jenkins failure seems to be related to libecl!?

I agree. Something fishy with the EclKW class, perhaps missing `inline` for non-template function in header? (That is usually the cause for such failures at least.)

andlaus:
still the same error. seems to be that annoying GitHub lag...

andlaus:
jenkins build this please

andlaus:
jenkins build this please

andlaus:
that was the github lag on jenkins again. I am pretty confident that it works now.

@akva2: I just checked: the problem does not occur with my private webserver: a force push to github from my workstation was immediately visible on the webserver. are you sure that (a) the problem is not caused by the jenkins setup and (b) that the current hosting provider does not put a proxy between the machine and github? finally (c) it could be that the physical location from where the request was made matters for github. if (b) or (c) are the case, I think it is time to move the CI server closer to home...

akva2:
how did you check if it was immediately available? by checking out the branch the pr tracks or by using the merge refs ? i think you did the former and not the latter.

andlaus:
> how did you check if it was immediately available?

on workstation:

```
git push andlaus TEST_BRANCH
```

on webserver:

```
git fetch andlaus
git log andlaus/TEST_BRANCH
```

on workstation:

```
touch CHANGELOG
git commit -a --amend --no-edit
git push -f TEST_BRANCH
```

on webserver:

```
git fetch andlaus
git log andlaus/TEST_BRANCH
```

in both cases, the commit IDs of the heads of the test branch were identical. AFAICS that is also how the jenkins server does it? (also note that it also worked with @atgeirr's workstation on Thursday, but back then we were sitting in the same office.)

akva2:
yes, so you are checking out the branch. that's not the problem. we cannot build the branch directly, we want to build the branch merged into the current master branch. this is the merge refs i refer to.

atgeirr:
A new test failure, this time in the added test... I don't get it, though?!?

andlaus:
yes, but you can easily to

```
git merge $REMOTE/$BRANCH
```

instead of


```
git merge upstream/pull//$PR_ID/head
```

right?

andlaus:
> A new test failure, this time in the added test... I don't get it, though?!?

the test driver thinks it is a simulation and tries to compare with reference results. fixed.

andlaus:
jenkins build this please

akva2:
well yes. but then i have to reimplement the entire error handling now done in the java code of the github plugin for jenkins in bash.

atgeirr:
> to me it seems that this is some hours of work at most.

Reimplementing the jenkins plugin looks like a lot more than some hours to me. Does it make sense to report this as an issue, perhaps even propose a patch to those who developed the plugin?

akva2:
problem is github, not the plugin.

andlaus:
> problem is github, not the plugin.

yes and no: github does not update the upstream/pull/$PR_ID/head reference in time after a force push, but the plugin could use the source branch from the outset. I have the feeling that there will be some mutual finger pointing between github and the jenkins devs if this is brought up to either of them. 

BTW: Jenkins is green now. mind to merge @atgeirr?

atgeirr:
> BTW: Jenkins is green now. mind to merge @atgeirr?

Not at all!

atgeirr:
With that merged: should it be picked to the release branch?

atgeirr:
Actually, I cannot figure out how to turn it on anymore. The async_output parameter is gone, and I cannot find any hooks into the EnableAsyncEclOutput property.

I do not think that it is a "macOS problem" since it invoked undefined behaviour, even if that behaviour turned out to be ok on current Linux implementations. It could easily break in the future.

andlaus:
> Actually, I cannot figure out how to turn it on anymore. The async_output parameter is gone, and I cannot find any hooks into the EnableAsyncEclOutput property.

run the tasklets test with a single worker thread.

> I do not think that it is a "macOS problem" since it invoked undefined behaviour

I agree, but "undefined" equalled "fine" on linux, so it is a bug that only affected non-Linux systems.

PR #313, opened 2018-04-23
fixing the polymer running for cells without water

GitPaean:
it was removed accidently in https://github.com/OPM/ewoms/pull/298.

### DIFF
--- a/ewoms/models/blackoil/blackoilpolymermodules.hh
+++ b/ewoms/models/blackoil/blackoilpolymermodules.hh
@@ -443,6 +443,9 @@ public:
                 * Toolbox::template decay<LhsEval>(fs.invB(waterPhaseIdx))
                 * Toolbox::template decay<LhsEval>(intQuants.porosity());
 
+        // avoid singular matrix if no water is present.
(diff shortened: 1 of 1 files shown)


GitPaean:
jenkins build this please

GitPaean:
It needs to go in the release.

andlaus:
seems good. Will merge once jenkins has okay'ed it...

andlaus:
I'm pretty sure that the problem is caused by the fact that the polymer-to-water ratio -- a.k.a. polymer concentration -- is picked as the primary variable. using a primary variable that corresponds directly to the amount of polymer would fix this in a better way. That's certainly post release stuff, though.

andlaus:
In an earlier, more academic life I had this before in the context of some compositional models: using concentrations as primary variables leads to a division by zero if the denominator becomes zero...

totto82:
Thanks for digging into this @GitPaean. Lets merge when Jenkins is green.

atgeirr:
Thanks for looking for, and finding, the bug! Merging.

PR #314, opened 2018-04-23
Improve tasklets

andlaus:
this PR adds a few minor improvements to the tasklet mechanism. in particular, the there now is the concept of a "worker thread index", which should make it relatively straight-forward to replace OpenMP by tasklets.

### DIFF
--- a/ewoms/parallel/tasklets.hh
+++ b/ewoms/parallel/tasklets.hh
@@ -61,6 +61,25 @@ private:
     int referenceCount_;
 };
 
+class TaskletRunner;
+
+// this class stores the thread local static attributes for the TaskletRunner class. we
+// cannot put them directly into TaskletRunner because defining static members for
+// non-template classes in headers leads the linker to choke in case multiple compile
+// units are used.
+template <class Dummy = void>
+struct TaskletRunnerHelper_
+{
+    static thread_local TaskletRunner* taskletRunner_;
+    static thread_local int workerThreadIndex_;
+};
+
+template <class Dummy>
+thread_local TaskletRunner* TaskletRunnerHelper_<Dummy>::taskletRunner_ = nullptr;
+
+template <class Dummy>
+thread_local int TaskletRunnerHelper_<Dummy>::workerThreadIndex_ = -1;
+
 /*!
  * \brief Handles where a given tasklet is run.
  *
@@ -136,7 +155,7 @@ public:
         threads_.resize(numWorkers);
         for (unsigned i = 0; i < numWorkers; ++i)
             // create a worker thread
-            threads_[i].reset(new std::thread(startWorkerThread_, this));
+            threads_[i].reset(new std::thread(startWorkerThread_, this, i));
     }
 
     /*!
@@ -157,6 +176,24 @@ public:
                 thread->join();
         }
     }
+
+    /*!
+     * \brief Returns the index of the current worker thread.
+     *
+     * If the current thread is not a worker thread, -1 is returned.
+     */
+    int workerThreadIndex() const
+    {
+        if (TaskletRunnerHelper_<void>::taskletRunner_ != this)
+            return -1;
+        return TaskletRunnerHelper_<void>::workerThreadIndex_;
+    }
+
+    /*!
+     * \brief Returns the number of worker threads for the tasklet runner.
+     */
+    int numWorkerThreads() const
+    { return threads_.size(); }
 
     /*!
      * \brief Add a new tasklet.
@@ -205,8 +242,13 @@ public:
 
 protected:
     // main function of the worker thread
-    static void startWorkerThread_(TaskletRunner* taskletRunner)
-    { taskletRunner->run_(); }
+    static void startWorkerThread_(TaskletRunner* taskletRunner, int workerThreadIndex)
+    {
+        TaskletRunnerHelper_<void>::taskletRunner_ = taskletRunner;
+        TaskletRunnerHelper_<void>::workerThreadIndex_ = workerThreadIndex;
+
+        taskletRunner->run_();
+    }
 
     //! do the work until the queue received an end tasklet
     void run_()
--- a/ewoms/parallel/tasklets.hh
+++ b/ewoms/parallel/tasklets.hh
@@ -220,7 +220,8 @@ protected:
                 [this]() -> bool
                 { return !taskletQueue_.empty(); };
 
-            workAvailableCondition_.wait(lock, /*predicate=*/workIsAvailable);
+            if (!workIsAvailable())
+                workAvailableCondition_.wait(lock, /*predicate=*/workIsAvailable);
 
             // remove tasklet from queue
             std::shared_ptr<TaskletInterface> tasklet = taskletQueue_.front();
@@ -239,10 +240,7 @@ protected:
                 // remove tasklets from the queue as soon as their reference count
                 // reaches zero, i.e. the tasklet has been run often enough.
                 taskletQueue_.pop();
-
             lock.unlock();
-            if (workIsAvailable())
-                workAvailableCondition_.notify_one();
 
             // execute tasklet
             tasklet->run();
--- a/tests/test_tasklets.cc
+++ b/tests/test_tasklets.cc
@@ -33,6 +33,8 @@
 #include <chrono>
 #include <iostream>
 
+Ewoms::TaskletRunner *runner;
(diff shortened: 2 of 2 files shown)


ewoms/parallel/tasklets.hh
```diff
@@ -61,6 +61,25 @@ private:
...
+
+// this class stores the thread local static attributes for the TaskletRunner class. we
+// cannot put them directly into TaskletRunner because defining static members for
+// non-template classes in headers leads the linker to choke in case multiple compile
+// units are used.
+template <class Dummy = void>
+struct TaskletRunnerHelper_
+{
+    static thread_local TaskletRunner* taskletRunner_;
+    static thread_local int workerThreadIndex_;
```
atgeirr: What exactly is `thread_local` here? Is it a C++11 concept?

akva2: http://en.cppreference.com/w/cpp/language/storage_duration it's c++11 to put a variable in the tls.

andlaus: note that the template class is needed to make the linker happy. any idea how to do this in a better way? static variables cannot be declared inline, right?

andlaus:
jenkins build this please

atgeirr:
@andlaus, I'll review this properly tomorrow.

joakim-hove:
> which should make it relatively straight-forward to replace OpenMP by tasklets.

NIH - a disease which is easily contracted  - I know :-)

andlaus:
> NIH - a disease which is easily contracted - I know :-)

Just like everyone else involved in *any* programming project (to some extend). ;)

but seriously, have you actually ever tried to compile that stuff with OpenMP? replacing OpenMP with tasklets would make everything work fine with a plain c++-11 compliant compile whereas OpenMP support is spotty for everything that is not GCC or clang...

atgeirr:
> replacing OpenMP with tasklets would make everything work fine

Assuming that the Tasklet code has the level of maturity to achieve this is a little bit optimistic... I am no huge fan of OpenMP though, for this kind of parallelism (which is why I am happy C++ provides it).

joakim-hove:
> but seriously, [...]

My experience with OpenMP is quite limited, so I can not really contribute to a discussion. But I must admit that my initial reaction to the "Easily replace OpenMP" claim was: "Wow - that was some hubris!"

andlaus:
> Easily replace OpenMP" claim was: "Wow - that was some hubris!"

it is *relatively* easy. I even have some prototype code for this already, so brace yourself for PRs...

atgeirr:
> it is relatively easy. I even have some prototype code for this already, so brace yourself for PRs...

If you are going all out with this, perhaps a separate, experimental repo would be a good idea? Of course I do not know where you intend to go with this.

andlaus:
a new repo is quite a bit of overkill: first, the git mechanism to deal with experimental features is called "branches"; second, this change affects only a few non-legacy locations (grep for "pragma *omp" in eWoms) and finally, by default OpenMP is disabled even if it is available and I have thus no intention whatsoever to enable multi-threaded linearization by default.

atgeirr:
> the git mechanism to deal with experimental features is called "branches"

Yes, but this is experimenting with a feature used in Flow, and you propose to merge it to master, not keeping it in a separate branch. Anyway, I'll review now!

bska:
Just out of curiosity, what is the reasoning behind using `std::thread` directly and unadorned `lock()`/`unlock()` calls as opposed to building the feature on top of `std::packaged_task`, `std::future` and/or `std::async()`?

andlaus:
> Just out of curiosity, what is the reasoning behind using std::thread directly and unadorned lock()/unlock() calls as opposed to building the feature on top of std::packaged_task, std::future and/or std::async()?

honestly: lack of knowledge, I've never seriously looked into any of these things. If you want to propose better solutions: patches are welcome, I promise to review them benevolently.

andlaus:
> Yes,  but this is experimenting with a feature used in Flow, and you propose  to merge it to master, not keeping it in a separate branch. Anyway, I'll  review now!

the oldest open opm-simulators PR is OPM/opm-simulators#1039 which is older than two years. So I guess branches are the way that experimental features are usually dealt with in OPM. 

Anyway, the present PR does not change anything which is used by `flow`, so it is *extremely* low risk...

atgeirr:
> Anyway, the present PR does not change anything which is used by flow, so it is extremely low risk...

That is false, since Flow uses the tasklets!

In any case, this PR does two things I think are unrelated:
 - Adding a worker thread index (implementation looks correct to me, if a little cludgy -- a cpp file would have been cleaner...).
 - Changing the way the condition_variable is used. I have tried to understand how this works, and I think the change is correct (i.e. the notify is not needed). Local testing revealed no failures from test_tasklets.

In the end, I think this can be merged. However I require one particular test first: please exercise the async output by running the Norne case with asynchronous output and verifying that you get no change. If that is successful, press green.

tests/test_tasklets.cc
```diff
@@ -59,23 +66,30 @@ int SleepTasklet::numInstantiated_ = 0;
 
 int main()
 {
-    Ewoms::TaskletRunner tr(2);
+    int numWorkers = 2;
+    runner = new Ewoms::TaskletRunner(numWorkers);
```
atgeirr: As an aside: why not use a smart pointer? It's been so long since I actually saw a delete statement that they are a bit exotic...

andlaus: because the pointer needs to be visible from global scope and we want to delete the object before the end of main().

andlaus:
> That is false, since Flow uses the tasklets!

yes, it uses tasklets. but the new functionality of this branch is not yet used and it also only uses a single worker thread. IOW, tasklets are currently just used as a straight replacement for `ThreadHandle`. (BTW: I'm pretty sure that with more than one worker thread, `ThreadHandle` has a race condition when modifying the work queue.)

>  Adding a worker thread index (implementation looks correct to me, if a little cludgy -- a cpp file would have been cleaner...).

I will not make people jump through the significant hoops associated with a library just for this. also, if `TaskletRunner` becomes a template class in the future, this would be mood.

> Changing the way the condition_variable is used. I have tried to understand how this works, and I think the change is correct (i.e. the notify is not needed). Local testing revealed no failures from test_tasklets.

I share your belief, but multi-threaded code has proven to be tricky more than once. I also think that the `if()` before the `wait()` is unnecessary, but it won't hurt I guess.

> please exercise the async output by running the Norne case with asynchronous output and verifying that you get no change. If that is successful, press green.

done that: no issues on my machine.

blattms:
> Anyway, the present PR does not change anything which is used by `flow`, so it is *extremely* low risk...

If this means that you claim that tasklets are not yet used by flow, then I highly doubt this.

Yesterday I started wondering why flow now spawns 34 threads per mpi process (used to be 2-4).
Your introduction of threads might be an explanation for this. Also I saw that there were some
bugs claimed to be from the parallel implementation with mpi that went away by fixed to tasklets.

I am also wondering whether we still use MPI in a thread safe manner. Do we even initialize it correctly, now?

andlaus:
> If this means that you claim that tasklets are not yet used by flow, then I highly doubt this.

I did not claim that. I said that none of the new functionality introduced by this PR is used by flow.

> Yesterday I started wondering why flow now spawns 34 threads per mpi process (used to be 2-4). Your introduction of threads might be an explanation for this. Also I saw that there were some bugs claimed to be from the parallel implementation with mpi that went away by fixed to tasklets.

bugs are possible, but I just checked: on my machine flow has is only 2 threads per process plus one overall, which is reasonable: one thread for ECL I/O, one for the main thread and I assume the last one is created internally by the pthreads or the MPI library. Also, the tasklet runner object is only created once at initialization, so I don't think the thread proliferation you saw is because of the tasklet stuff. Can you check?

> Also I saw that there were some bugs claimed to be from the parallel implementation with mpi that went away by fixed to tasklets. 

which ones where that?

> I am also wondering whether we still use MPI in a thread safe manner. Do we even initialize it correctly, now?

the ECL I/O code and the linearization procedure do not do any MPI calls, at least none that require communication. That said, this is not the case for e.g. `Dune::VTKWriter`, so parallel asynchronous VTK output is disabled in eWoms for now. (As far as I can see, the `MPI_Barrier()` call in `Dune::VTKWriter` is superfluous as well, so maybe it should be removed for the next Dune version?)

PR #316, opened 2018-04-28
Use RestartValue when calling Ecl writeTimeStep

joakim-hove:
See: https://github.com/OPM/opm-common/pull/391

### DIFF
--- a/ebos/eclthresholdpressure.hh
+++ b/ebos/eclthresholdpressure.hh
@@ -157,6 +157,11 @@ public:
         return thpres_[equilRegion1Idx*numEquilRegions_ + equilRegion2Idx];
     }
 
+    const std::vector<Scalar>& data() const {
+        return thpres_;
+    }
+
+
 private:
     // compute the defaults of the threshold pressures using the initial condition
     void computeDefaultThresholdPressures_()
--- a/ebos/eclwriter.hh
+++ b/ebos/eclwriter.hh
@@ -37,7 +37,10 @@
 
 #if HAVE_ECL_OUTPUT
 #include <opm/output/eclipse/EclipseIO.hpp>
+#include <opm/output/eclipse/RestartValue.hpp>
 #endif
+
+#include <opm/parser/eclipse/Units/UnitSystem.hpp>
 
 #include <opm/grid/GridHelpers.hpp>
 
@@ -146,9 +149,10 @@ public:
      */
     void writeOutput(Opm::data::Wells& localWellData, Scalar curTime, bool isSubStep, Scalar totalSolverTime, Scalar nextStepSize)
     {
-#if !HAVE_ECL_OUTPUT
-        throw std::runtime_error("Eclipse output support not available in opm-common, unable to write ECL output!");
-#else
+
+#if !HAVE_ECL_INPUT
+        throw std::runtime_error("Unit support not available in opm-common.");
+#endif
 
         int episodeIdx = simulator_.episodeIndex() + 1;
         const auto& gridView = simulator_.vanguard().gridView();
@@ -185,11 +189,8 @@ public:
 
         // write output on I/O rank
         if (collectToIORank_.isIORank()) {
-            std::map<std::string, std::vector<double>> extraRestartData;
-
-            // Add suggested next timestep to extra data.
-            if (!isSubStep)
-                extraRestartData["OPMEXTRA"] = std::vector<double>(1, nextStepSize);
+            const auto& eclState = simulator_.vanguard().eclState();
+            const auto& simConfig = eclState.getSimulationConfig();
 
             // Add TCPU
             if (totalSolverTime != 0.0)
@@ -198,23 +199,29 @@ public:
             bool enableDoublePrecisionOutput = EWOMS_GET_PARAM(TypeTag, bool, EclOutputDoublePrecision);
             const Opm::data::Solution& cellData = collectToIORank_.isParallel() ? collectToIORank_.globalCellData() : localCellData;
             const Opm::data::Wells& wellData = collectToIORank_.isParallel() ? collectToIORank_.globalWellData() : localWellData;
+            Opm::RestartValue restartValue(cellData, wellData);
 
             const std::map<std::pair<std::string, int>, double>& blockData
                 = collectToIORank_.isParallel()
                 ? collectToIORank_.globalBlockData()
                 : eclOutputModule_.getBlockData();
 
+            // Add suggested next timestep to extra data.
+            if (!isSubStep)
+                restartValue.addExtra("OPMEXTRA", std::vector<double>(1, nextStepSize));
+
+            if (simConfig.hasThresholdPressure())
+                restartValue.addExtra("THPRES", Opm::UnitSystem::measure::pressure, simulator_.problem().thresholdPressure().data());
+
             // first, create a tasklet to write the data for the current time step to disk
             auto eclWriteTasklet = std::make_shared<EclWriteTasklet>(*eclIO_,
                                                                      episodeIdx,
                                                                      isSubStep,
                                                                      curTime,
-                                                                     cellData,
-                                                                     wellData,
+                                                                     restartValue,
                                                                      miscSummaryData,
                                                                      regionData,
                                                                      blockData,
-                                                                     extraRestartData,
                                                                      enableDoublePrecisionOutput);
 
             // then, make sure that the previous I/O request has been completed and the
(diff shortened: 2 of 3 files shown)


joakim-hove:
jenkins build this opm-common=391 please

ebos/eclwriter.hh
```diff
@@ -180,41 +181,48 @@ public:
...
                 = collectToIORank_.isParallel()
                 ? collectToIORank_.globalBlockData()
                 : eclOutputModule_.getBlockData();
 
+            // Add suggested next timestep to extra data.
+            if (!isSubStep)
+                restartValue.add_extra("OPMEXTRA", std::vector<double>(1, nextStepSize));
+
+            if (simConfig.hasThresholdPressure())
+                restartValue.add_extra("THPRES", Opm::UnitSystem::pressure, simulator_.XXXX.data());
```
joakim-hove: @andlaus : A hint of how to get the `thpres` values here would be good!

andlaus: if you have the simulator object it is

`simulator.problem().thresholdPressure(cellIdx1, cellIdx2)`

for the feature at hand, it is probably be a good idea to amend `EclProblem` by a `thresholdPressure()` method which returns the `EclThresholdPressure` object and extend the `EclThresholdPressure` to get the list of threshold pressures between EQUIL regions instead of just being able to query element pairs.

joakim-hove: Ok thanks, I intended to use the new method I have added to get the full vector. Will be on Monday though

akva2:
jenkins build this opm-common=391 opm-simulators=1470 please

joakim-hove:
jenkins build this opm-common=391 opm-simulators=1470 please

ebos/eclthresholdpressure.hh
```diff
@@ -157,6 +157,11 @@ public:
         return thpres_[equilRegion1Idx*numEquilRegions_ + equilRegion2Idx];
     }
 
+    const std::vector<double>& data() const {
```
andlaus: return `std::vector<Scalar>` please.

ebos/eclwriter.hh
```diff
@@ -37,6 +37,11 @@
 
 #if HAVE_ECL_OUTPUT
 #include <opm/output/eclipse/EclipseIO.hpp>
+#include <opm/output/eclipse/RestartValue.hpp>
+#endif
+
+#if HAVE_ECL_INPUT
```
andlaus: that `#if` is superfluous: ebos won't compile anyway if it is false.

ebos/eclwriter.hh
```diff
@@ -150,6 +155,10 @@ public:
         throw std::runtime_error("Eclipse output support not available in opm-common, unable to write ECL output!");
 #else
 
+#if !HAVE_ECL_INPUT
```
andlaus: same as above. (I will refrain from mentioning this from here on.)

ebos/eclwriter.hh
```diff
@@ -198,23 +204,29 @@ public:
...
+            Opm::RestartValue restartValue(cellData, wellData);
 
             const std::map<std::pair<std::string, int>, double>& blockData
                 = collectToIORank_.isParallel()
                 ? collectToIORank_.globalBlockData()
                 : eclOutputModule_.getBlockData();
 
+            // Add suggested next timestep to extra data.
+            if (!isSubStep)
+                restartValue.add_extra("OPMEXTRA", std::vector<double>(1, nextStepSize));
```
andlaus: this will cause report steps and "substeps" contain a different set of fields. is this intentional. (Also, I think that only "report steps" are written anyway, but I'm not sure.)

ebos/eclwriter.hh
```diff
@@ -230,21 +242,19 @@ public:
...
-                                                         {"SOMAX", {Opm::UnitSystem::measure::identity, simulator_.problem().vapparsActive()}},
-                                                         {"PCSWM_OW", {Opm::UnitSystem::measure::identity, enableHysteresis}},
-                                                         {"KRNSW_OW", {Opm::UnitSystem::measure::identity, enableHysteresis}},
-                                                         {"PCSWM_GO", {Opm::UnitSystem::measure::identity, enableHysteresis}},
-                                                         {"KRNSW_GO", {Opm::UnitSystem::measure::identity, enableHysteresis}}};
-
-        std::map<std::string, bool> extra_keys {
-            {"OPMEXTRA" , false}
-        };
+        std::vector<Opm::RestartKey> solution_keys {{"PRESSURE" , Opm::UnitSystem::measure::pressure},
```
andlaus: it is kind of hard to find out what changed here. any hints?

ebos/eclwriter.hh
```diff
@@ -413,35 +423,29 @@ private:
...
-        std::map<std::string, std::vector<double>> extraRestartData_;
         bool writeDoublePrecision_;
 
         explicit EclWriteTasklet(Opm::EclipseIO& eclIO,
                                  int episodeIdx,
                                  bool isSubStep,
                                  double secondsElapsed,
-                                 Opm::data::Solution cellData,
-                                 Opm::data::Wells wellData,
+                                 Opm::RestartValue restartValue,
```
andlaus: This is not really my shot to call here, but I'd prefer a more descriptive name than `RestartValue` for this class, e.g. it seems like it does neither represent a single field, not do the fields contain a single value. maybe something like `RestartOutputFields`?

ebos/eclwriter.hh
joakim-hove: Yes - that was the way it was before as well; simple to change though.

ebos/eclwriter.hh
joakim-hove: The point of the class was to provide *value semantics* to everything passed back and forth between the simulator and the restart file.

ebos/eclwriter.hh
joakim-hove: > it is kind of hard to find out what changed here. any hints?

What has happened here is a quite limited refactoring, there is a small class `RestartKey`in opm-common which is used to denote which keywords are expected to be found in the restart file.  Previosuly it looked like this:
```C++
struct RestartKey {
    UnitSystem::measure dim;
    bool required;
};
```
and the datastructure in ewoms was a `std::map<std::string, RestartKey>`. Now the structure has been augmented with `std::string key` and the datastructure in ewoms is a `std::vector<RestartKey>`.

joakim-hove:
Thank you for the review - I must admit my dune setup is broken at the moment, so I have tried to compile using Jenkins; that has not been very efficient ....

joakim-hove:
jenkins build this opm-common=391 opm-simulators=1470 please

joakim-hove:
jenkins build this opm-common=391 opm-simulators=1470 please

andlaus:
> my dune setup is broken at the moment, so I have tried to compile using Jenkins; that has not been very efficient ....

have you tried the alternative build system ;)

ebos/eclwriter.hh
andlaus: I see, thanks for the explaination. It sounds like there are no new fields added.

ebos/eclwriter.hh
andlaus: okay. IMO it is just a bit confusing outside of the context it was written in.

ebos/eclwriter.hh
```diff
@@ -224,27 +231,24 @@ public:
...
-                                                         {"KRNSW_OW", {Opm::UnitSystem::measure::identity, enableHysteresis}},
-                                                         {"PCSWM_GO", {Opm::UnitSystem::measure::identity, enableHysteresis}},
-                                                         {"KRNSW_GO", {Opm::UnitSystem::measure::identity, enableHysteresis}}};
-
-        std::map<std::string, bool> extra_keys {
-            {"OPMEXTRA" , false}
-        };
+        std::vector<Opm::RestartKey> solution_keys {{"PRESSURE" , Opm::UnitSystem::measure::pressure},
+                                                    {"SWAT" ,     Opm::UnitSystem::measure::identity, FluidSystem::phaseIsActive(FluidSystem::waterPhaseIdx)},
+                                                    {"SGAS" ,     Opm::UnitSystem::measure::identity, FluidSystem::phaseIsActive(FluidSystem::gasPhaseIdx)},
```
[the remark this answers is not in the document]
joakim-hove: OK - I have updated with a `static_cast<bool>( ... )` - but really I would say that a function called: `phaseIsActive( ... )` *should* return `bool` ...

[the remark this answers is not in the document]
atgeirr: > a function called: phaseIsActive( ... ) should return bool ...

On that we agree.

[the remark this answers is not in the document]
andlaus: it would be better if you changed the return type of the `isPhaseActive()` method in opm-material instead of hacking around externally. I guess that the fact that it returns an integer is a copy-and-pasto, I certainly cannot come with a reason for it.

joakim-hove:
@andlaus : This has changed quite a lot since you gave your initial approval - still OK to merge?

andlaus:
go ahead. Please also change the return value of the fluid system's `isPhaseActive()` method to `bool` in opm-material.

PR #319, opened 2018-05-09
change ecl output name from SSOL to SSOLVENT

totto82:
Reason: use the same output name as eclipse. 

Need an update of ref. data in opm-tests before merging!

### DIFF
--- a/ebos/ecloutputblackoilmodule.hh
+++ b/ebos/ecloutputblackoilmodule.hh
@@ -773,7 +773,7 @@ public:
             sol.insert ("SOMAX", Opm::UnitSystem::measure::identity, std::move(soMax_), Opm::data::TargetType::RESTART_SOLUTION);
 
         if (sSol_.size() > 0)
-            sol.insert ("SSOL", Opm::UnitSystem::measure::identity, std::move(sSol_), Opm::data::TargetType::RESTART_SOLUTION);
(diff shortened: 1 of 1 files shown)


totto82:

andlaus:
jenkins build this opm-tests=10 please

andlaus:
let's merge; Jenkins is happy and this is a pretty trivial change. thanks for the patch.

atgeirr:
Is it not strange that this did not require a data update? OPM/opm-tests#10 is empty. Or did this only change the name used internally in the code?

andlaus:
> Is it not strange that this did not require a data update? OPM/opm-tests#10 is empty. Or did this only change the name used internally in the code?

I also understand this. In particular because Jenkins failed for the update_data trigger, but subsequently succeeded with OPM/opm-data#10. I hope nothing broke because of this...

totto82:
The test should have failed since the name is changed. It correctly failed on my machine. Do we need to make a trivial PR and check the state of Jenkins.

atgeirr:
The post-builder for ewoms is running right now, and may produce a failure.

andlaus:
> The test should have failed since the name is changed. It correctly failed on my machine.

it also rightfully failed on jenkins when you triggered it the first time. there might be a problem with the update_data script if new reference solutions are added, but what I don't understand in this case is that the second time that Jenkins was triggered with the updated data, it succeeded. very strange.

PR #322, opened 2018-05-15
[bugfix][PrimaryVariables] Implement FieldTraits for PrimaryVariables

dr-robertk:
The specialization of FieldTraits for the PrimaryVariables was missing. This patch implements them for all PrimaryVariables in eWoms based on FieldVector.

### DIFF
--- a/ewoms/disc/common/fvbaseprimaryvariables.hh
+++ b/ewoms/disc/common/fvbaseprimaryvariables.hh
@@ -28,6 +28,8 @@
 #ifndef EWOMS_FV_BASE_PRIMARY_VARIABLES_HH
 #define EWOMS_FV_BASE_PRIMARY_VARIABLES_HH
 
+#include <type_traits>
+
 #include "fvbaseproperties.hh"
 
 #include <opm/material/common/Valgrind.hpp>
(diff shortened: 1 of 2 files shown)


dr-robertk:
jenkins build this please with downstreams

andlaus:
looks fine. it seems like you got victim of the "non-inheritance" problem of standard traits and fixed it using the property system?

before I merge, what was the concrete issue that motivated the patch?

andlaus:
jenkins build this please

ewoms/disc/common/fvbaseprimaryvariables.hh
```diff
@@ -121,4 +121,39 @@ public:
...
+      typedef field_type                               real_type;
+  };
+
+
+  /** Compatibility traits class for DenseVector and DenseMatrix.
+   */
+  template<class TypeTag, template <class> class EwomsPrimaryVariables>
+  struct FieldTraits< EwomsPrimaryVariables< TypeTag > >
+    : public FieldTraitsImpl< TypeTag,
+               std::is_convertible< EwomsPrimaryVariables< TypeTag >,
```
andlaus: isn't this always true at the moment?

andlaus:
hm, jenkins is unhappy about it because it cannot not build it.

ewoms/disc/common/fvbaseprimaryvariables.hh
dr-robertk: Most likely. Just in case I made a default implementation that is the same as the one in dune-common.

dr-robertk:
jenkins build this please

dr-robertk:
I redid the FieldTraits impl. It's much better now and hopefully covers all cases. One more thing to consider is to move the femcpgridcompat to ewoms/common since it needs to be used also in co2injection_flash_ni_*.

dr-robertk:
jenkins build this please

dr-robertk:
@andlaus Please look at the femcpgridcompat.hh It should be renamed to dunefemcompat.hh and moved to ewoms/common and the cpgrid things should be guarded with a HAVE_OPM_GRID.

andlaus:
> Please look at the femcpgridcompat.hh It should be renamed to dunefemcompat.hh and moved to ewoms/common and the cpgrid things should be guarded with a HAVE_OPM_GRID.

wouldn't it be better to fix `CpGrid` to confirm to this part of the Dune grid interface and get rid of that file?

tests/co2injection_flash_ecfv.cc
```diff
@@ -35,6 +35,9 @@
 #include <ewoms/common/start.hh>
 #include <ewoms/models/flash/flashmodel.hh>
 #include <ewoms/disc/ecfv/ecfvdiscretization.hh>
+
+#include <ebos/femcpgridcompat.hh>
```
andlaus: why is this actually needed? Jenkins does not use dune-fem, right?

andlaus:
Okay, I'll merge this once the remaining two questions are answered...

dr-robertk:
It would be. I can provide a patch. The lower part of the file for quad
precision is still needed.

tests/co2injection_flash_ecfv.cc
dr-robertk: When fem is used the tests won't compile.

andlaus:
OPM/opm-material#297 should take care of the spurious need to include `femcpgridcompat.hh` can you confirm?

dr-robertk:
jenkins build this please

dr-robertk:
@andlaus I addressed your comments.

andlaus:
I'm happy with the patch now, but for some reason your jenkins trigger did not work for some reason. IIRC jenkins is touchy about firing off a comment immediately after the trigger. let's try again and merge if it is happy.

andlaus:
jenkins build this please

PR #324, opened 2018-05-30
Load thpres from restart file

joakim-hove:
See opm/opm-common#410

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -399,7 +399,8 @@ public:
             // create the ECL writer
             eclWriter_.reset(new EclWriterType(simulator));
 
-        // Hack to compute the initial thpressure values for restarts
+        // Loading the solution from a restart file is done recursively, we need this
+        // bool variable to signal the stop condition.
         restartApplied = false;
     }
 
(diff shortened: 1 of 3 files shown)


ebos/eclproblem.hh
```diff
@@ -851,6 +849,10 @@ public:
         return thresholdPressures_;
     }
 
+    EclThresholdPressure<TypeTag>& thresholdPressure() {
```
andlaus: nit: the eWoms convention for oneline methods is:
```c++
EclThresholdPressure<TypeTag>& thresholdPressure()
{ return thresholdPressures_; }
```

(this repeats below.)

ebos/eclwriter.hh
```diff
@@ -260,6 +260,13 @@ public:
             unsigned globalIdx = collectToIORank_.localIdxToGlobalIdx(elemIdx);
             eclOutputModule_.setRestart(restart_values.solution, elemIdx, globalIdx);
         }
+        const auto& input_thpres = eclState().getSimulationConfig().getThresholdPressure();
```
andlaus: camelCase please. (same for `restartValues`, `mutableSimulator` and `thpresValues`)

joakim-hove:
OK - coding style updated.

andlaus:
cool, thanks a lot. merge whenever convenient. BTW: is this jenkinsed as part of OPM/opm-simulators#1488?

joakim-hove:
> is this jenkinsed as part of OPM/opm-simulators#1488?

Yes

joakim-hove:
OK - this has been updated due to failures in the integration tests. Observe the member vaiable `restartApplied` in the `eclproblem.hh` implementation: My guess is that that is introduced specifically for `THPRES` - and should therefore not be required any longer; however the current implementation still needs the `restartApplied`flag to avoid a recursive death.

totto82:
> My guess is that that is introduced specifically for THPRES - and should therefore not be required any longer; however the current implementation still needs the restartAppliedflag to avoid a recursive death.

The flag can be removed together with the recursive call. I can clean this up after this is merged if you don't want to do it yourself.

joakim-hove:
> The flag can be removed together with the recursive call. I can clean this up after this is merged if you don't want to do it yourself.

Cool if you do that - my initial attempt just led to a recursive meltdown.

PR #325, opened 2018-06-10
Rename Completions -> Connections

joakim-hove:
See: https://github.com/OPM/opm-common/pull/419


andlaus:
the patch is good and I'm indifferent about how this is called, so merge whenever you deem appropriate.

One thing which I noticed: If I understood this correctly, the well model calls these thingies "perforations" in most places. is there a difference? if no, I think that "perforations" is the best name.

joakim-hove:
> If I understood this correctly, the well model calls these thingies "perforations" in most places. is there a difference?

 I hoped no one would ask; I think `connection == perforation`

lars-petter-hauge:
Perforations in a practical drilling -> completion view are actual holes that perforation guns create (ref. link from @GitPaean). These perforations can be placed approximately ten per meter (so obviously much more than just one per grid cell). Also, not all wells have such perforations into the reservoir, but rather use "pre-perforated" tubing with protection screens around them. However in the reservoir model we consider a grid cell perforated if it has an open connection to the reservoir, indifferent if the well in reality have had perforation guns used. I agree that it is confusing when reading manuals and that the concepts are interchanged.

What could be considered however is which keywords/workflow identify differences between connection, completion and perforation.

`WFRICTN` describes the well as a friction well - mainly used for horizontal and multilateral wells where friction along the production line represents significant pressure loss. In this case it is necessary to describe the distance between the **the start of the perforations in the first grid cell** to the **end of the perforations in the last grid cell** (and as such it differs from connection, which is only concerned about the direction from which fluid may flow into the grid block.

`COMPSEGS` also requires the distance of how far perforations reach into the grid blocks (and the `COMPSEGS` keyword has already been implemented..)

So I believe using `connection` would be the most appropriate in this case.

PR #326, opened 2018-06-14
remove support for openMP

andlaus:
this is the alternative to merging #318.


andlaus:
jenkins build this please

atgeirr:
> this is the alternative to merging #318.

I think actively embracing OpenMP (and maintaining instead of removing this code) should be considered a third alternative. Arguments are essentially already presented in #318 (we should not repeat that discussion here).

PR #327, opened 2018-06-14
Some property- and parameter system improvements

andlaus:
The parameter system retrieves more changes: This PR adds support for positional command line parameters and overhauls the system how the help messages are created. As a guiney pig, ebos gets support for positional parameters, and a description of is added to the help messages of ebos and the lens_* simulators.

the property system only introduces `BEGIN_PROPERTIES` and `END_PROPERTIES` macros in order to be flexible on how properties are declared in the future. Note that while this is not a very significant change in itself, converting all code to use these macros implies plenty of boiler plate changes, i.e., it is best to review the individual patches of this PR instead of the accumulated changes. (if wanted, it can also be split because it is fully bisectable.)

### DIFF
--- a/ebos/ebos.cc
+++ b/ebos/ebos.cc
@@ -32,10 +32,11 @@
 
 #include "eclproblem.hh"
 
-namespace Ewoms {
-namespace Properties {
+BEGIN_PROPERTIES
+
 NEW_TYPE_TAG(EclProblem, INHERITS_FROM(BlackOilModel, EclBaseProblem));
-}}
+
+END_PROPERTIES
 
 int main(int argc, char **argv)
 {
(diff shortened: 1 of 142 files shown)


andlaus:
jenkins build this please

andlaus:
yes, but is trival enough not to warrant a separate PR?! (it does not contain any functional changes)

the usual procedure for such micro-cleanups so far has been to embedd in other PRs so far, usually not even in a separate patch. If you want to change this, I'll open a separate PR and point such stuff out in my reviews, too.

andlaus:
jenkins build this please

totto82:
Looks fine. I will merge it when the test is done.

PR #329, opened 2018-06-20
ebos: allow to specify the ECL case name

andlaus:
... instead of just the full file name of the deck. this mirrors flow's behavior.


totto82:
Thanks. Even ebos --option=foo SPE1CASE1 --option2=foo2 works. I will merge.

PR #332, opened 2018-06-21
use standard std::mutex instead of the openMP one

andlaus:
only having to rely on c++-2011 is a good thing, right? This PR also gets of the custom implementation of `ScopedLock` (where required, it ought to be replaced by `std::unique_lock`.)


blattms:
Isn't there a subtle difference between the two?

OmpMutex::lock and OmpMutex::unlock do nothing if OMP is not activated (No OMP -> No locks). With your change there will be at least one if statement executed when locking (that is if the STL implementors were clever).

andlaus:
> OmpMutex::lock and OmpMutex::unlock do nothing if OMP is not activated (No OMP -> No locks). 

true. *but*: this is not a general threading library but one tailored at the eWoms use case and eWoms currently does not use any threads without openMP. (besides asynchronous output and there we need to use openmp-independent locking anyway). Also, nobody has found any consistent performance impact for #318 (where this patch is a subset of), so the performance argument is purely academic?

andlaus:
I did some casual testing on performance: for some reason this branch seems to be *faster* than the current master, at least the jitter in the linearization time of the approximately ten runs of lens_immiscible_ecvf_ad with one global grid refinement was lower and the runtime also tended to be a few miliseconds better. this is a bit strange, but maybe it can be explained that the compiler does not need to create `ScopedLock` objects anymore?

totto82:
Looks fine. I will merge it tomorrow if @atgeirr or  @dr-robertk don't object.

ewoms/disc/common/fvbasediscretization.hh
```diff
@@ -840,13 +840,13 @@ public:
                 asImp_().localResidual(threadId).eval(residual, elemCtx);
 
                 size_t numPrimaryDof = elemCtx.numPrimaryDof(/*timeIdx=*/0);
-                ScopedLock addLock(mutex);
+                mutex.lock();
```
atgeirr: Why is the mutex required here? Don't we write to each item only once in line 847?

andlaus:
@atgeirr: To bring this discussion out of the realm of speculation, can you please benchmark this and see if you get any performance regression? If anything, I saw a small improvement, but you're using a relatively exotic system...

ewoms/disc/common/fvbasediscretization.hh
andlaus: yes, but adding two FieldMatrix objects is not threadsafe and this code does not assume that each entry is only accessed once (for example, this is not the case for the VCFV discretization)

atgeirr: > this code does not assume that each entry is only accessed once (for example, this is not the case for the VCFV discretization)

Ok; it is only accessed once for the cell-centered FV discretization, right?

> adding two FieldMatrix objects is not threadsafe

That seems strange to me! In what sense do you mean it is not threadsafe?

atgeirr:
> @atgeirr: To bring this discussion out of the realm of speculation, can you please benchmark this and see if you get any performance regression? If anything, I saw a small improvement, but you're using a relatively exotic system...

I can take the chance on the performance based on your results and what I have read. If you give me a little time I can also check for myself, that is of course better, but I will probably not be able to do that until tomorrow or so.

ewoms/disc/common/fvbasediscretization.hh
andlaus: > Ok; it is only accessed once for the cell-centered FV discretization, right?

yes, but that's not the only customer of this code. (and for `flow` this particular code is not even used, so it's not worthwhile to optimize for that special case like for the actual linearization routine.)

>> adding two FieldMatrix objects is not threadsafe
>
> That seems strange to me! In what sense do you mean it is not threadsafe?

consider two threads which both simultaneously execute

```c++
x += y;
```

In this scenario, the value of `x` it is undefined because the addition of `x` is done in a register. The other thread does not "see" that register and may overwrite the value of the first thread at any point. Did anyone ever say that multi-threaded programming was easy?

atgeirr: Your point is general, and not specific to FieldMatrix. I assume you mean that adding them *in this context* is not threadsafe because two threads could both try to incrementally add to the same matrix? (So any threads can add FieldMatrix objects as much as they like, as long as they do not write to the same location.)

andlaus: > (So any threads can add FieldMatrix objects as much as they like, as long as they do not write to the same location.)

as far as I'm aware the definition of a piece of code being "thread safe" is that it can by called by any thread in any context, right? ;)

andlaus: (if not, everything is threadsafe.)

atgeirr: So you mean simply that FieldMatrix itself makes no effort to ensure it is written to by only one thread at a time?

I initially got the impression that you meant something worse was going on: a static memory caching or something that would make it impossible to parallelize.

andlaus: > So you mean simply that FieldMatrix itself makes no effort to ensure it is written to by only one thread at a time?

yes. It IMO does exactly what it is supposed to do, but it is not threadsafe in the sense that if multiple threads manipulate the same object at once, the result is undefined; hence the need for the mutex.

andlaus:
@atgeirr: any performance news from your side?

atgeirr:
Running test, but had to interrupt because of a meeting. Resuming soon.

atgeirr:
I could not observe a meaningful difference: there was a 20 second difference in disfavor of this PR on Norne, but all of that was in the linear solver, which is not touched at all here. So I'll put that down to random variation.

andlaus:
that's congruent with my measurements. mind to press green?

PR #333, opened 2018-06-28
rename completion to connection

totto82:
This is a pure rename PR to make the connection vs completion naming consistent.


totto82:
jenkins build this please

andlaus:
looks good & not much can go here, merging.

PR #334, opened 2018-06-29
parameter system: add ability to hide parameters

andlaus:
i.e. they are still there but not mentioned in the usage message. only registered parameters can be hidden an only before the parameter registration is closed by EWOMS_END_PARAMETER_REGISTRATION.


andlaus:
jenkins build this please

totto82:
Thanks. This answers my main concern with OPM/opm-simulators#1512

PR #335, opened 2018-06-29
WIP Fix thpress restart

totto82:
THPRES is read from the restart file
We no longer need to compute it from the initial conditions if defaulted.

Currently the restart of Model 2 doesn't work due to issues with THPRES input. 
This PR fixes this, but there is a slight difference in results. I have compared the
computed thpressures with the ones read from the restart file and they are identical as far as I can see. 

I will need to do some more testing to figure out whether the mis-match is related to THPRES or to something els.

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -446,9 +446,6 @@ public:
             // create the ECL writer
             eclWriter_.reset(new EclWriterType(simulator));
 
-        // Loading the solution from a restart file is done recursively, we need this
-        // bool variable to signal the stop condition.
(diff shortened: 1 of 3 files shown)


andlaus:
I assume this is primarily a fix to make THPRES work for restart runs of model 2, not a performance improvement? Anyway, the patch seems good to me. Seems like we need an update of the reference data?

andlaus:
jenkins build this please

andlaus:
okay, this looks good, and whilie it is not yet perfect, it improves things. merging.

PR #338, opened 2018-07-03
fix linkage of binaries that use EWOMS_HIDE_PARAM()

andlaus:
this lead to undefined symbols for binaries that use EWOMS_HIDE_PARAM() when using debugging flags on GCC. They where caused by trying to take references to the values of some properties and these values are `static const int` class attributes. Since I'm pretty sure that the c++ standard mandates that symbols for such variables must be emitted, this might constitute a GCC bug. (also, when optimizations are enabled the linker errors went away. this was probably caused by the fact that this variable was unused and the compiler thus probably killed it completely.)


andlaus:
jenkins build this please

totto82:
Thanks for the fix.

blattms:
Let's help other people (using a web search) and provide the error message.

PR #341, opened 2018-07-09
complain if grid adaptation ought to be used at the same time as asynchronous VTK output

andlaus:
with grid adaptation, the modification of the grid by the adaptation routines might happen at the same time as `Dune::VTKWriter` tries to access it. this leads to some very enigmatic runtime errors.


andlaus:
jenkins build this please

PR #342, opened 2018-07-10
black oil model: use reasonable equation weights if surface volumes are conserved

andlaus:
`flow` is unaffected by this because the equation weight is only used to compute the errors in the non-linear and linear solvers and for both of these `flow` rolls its own code...


andlaus:
jenkins build this please

PR #340, opened 2018-07-09
Better well API

andlaus:
this PR improves the API for wells in ebos. This includes some usability improvements and some small changes to the code for auxiliary discretization modules which enable them to do Schur complements. the usability improvments for the auxiliary modules require some trivial downstream cleanup.

PR #343, opened 2018-07-10
simulator: provide a mutable version of Simulator::executionTimer()

andlaus:
this allows external users like `flow` to make the execution timer return accurate timings. In `flow`'s case this is useful to implement stuff like the `TCPU` output field.

### DIFF
--- a/ewoms/common/simulator.hh
+++ b/ewoms/common/simulator.hh
@@ -284,6 +284,8 @@ public:
      */
     const Ewoms::Timer& executionTimer() const
     { return executionTimer_; }
+    Ewoms::Timer& executionTimer()
+    { return executionTimer_; }
 
     /*!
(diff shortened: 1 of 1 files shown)


andlaus:
jenkins build this please

akva2:
i am curious why you need a mutable ref to obtain accurate timings?

andlaus:
depends on what you mean: it is not necessary to obtain accurate timings per-se but is is required to be able to get accurate timings *from the eWoms simulator object*, because `flow` needs to be able to start timer when it enters its main loop.

BTW: do you mind to press Green for #340, #341 and #342 (#340 requires OPM/opm-simulators#1531)

akva2:
i see. makes sense. was just curious to the wording was all (could look like wrong handling of constness..)

andlaus:
no problem, sorry if I was too ambiguous.

PR #344, opened 2018-07-10
linear solvers: reduce the singularity limit to 10^-30

andlaus:
as usual, `flow` is completely unaffected by this because it does not use this code.


andlaus:
jenkins build this please

PR #347, opened 2018-07-12
CollectToIORank: fix some of the const mess

andlaus:
as noted by @blattms in his review of #346, some const qualifiers have been accidentally changed in CollectToIORank. this patch should revert them back to the state before #346.

on a more general note, while doing this patch I noticed that the `const`-ness situation in this class is (and always has been) quite messy and the code looks needlessly complicated in some places, i.e., it ought to be significantly refactored IMO.

### DIFF
--- a/ebos/collecttoiorank.hh
+++ b/ebos/collecttoiorank.hh
@@ -104,8 +104,8 @@ public:
         std::vector<int>& ranks_;
 
     public:
-        DistributeIndexMapping(std::vector<int>& globalIndex,
-                               std::vector<int>& distributedGlobalIndex,
+        DistributeIndexMapping(const std::vector<int>& globalIndex,
+                               const std::vector<int>& distributedGlobalIndex,
                                IndexMapType& localIndexMap,
                                IndexMapStorageType& indexMaps,
                                std::vector<int>& ranks)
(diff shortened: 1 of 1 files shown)


andlaus:
jenkins build this please

andlaus:
Jenkins is happy. @blattms, @akva2: merge whenever.

andlaus:
okay. merge or close the PR if you like...

andlaus:
@blattms: I don't want this to become another eternally open PR. if you don't take action, I'll close it on Thursday or Friday. Also, feel free to propose a better solution for your review comment of #346 if you do not like this patch.

blattms:
Please mind your tone.

ebos/collecttoiorank.hh
```diff
@@ -279,14 +279,14 @@ public:
         const Opm::data::Solution& localCellData_;
         Opm::data::Solution& globalCellData_;
 
-        const IndexMapType& localIndexMap_;
-        const IndexMapStorageType& indexMaps_;
+        IndexMapType& localIndexMap_;
+        IndexMapStorageType& indexMaps_;
```
blattms: What is the reason of removing const here (contrary to the commit message)? Neither `localIndexMap_` nor `indexMaps_` should have been modified, before. Otherwise we would have experienced compile errors, wouldn`t we?

andlaus:
commit message changed. still, the comment about the file needing some rework is valid. (it just does not belong into a commit message.)

andlaus:
jenkins build this please

ebos/collecttoiorank.hh
andlaus: this was done because adding `const` at the other place required it to be removed here (it did not build for whatever reason). did I ever mention that the `const`-ness situation is messy in that file? maybe I made a mistake, so go and test yourself...

andlaus:
okay, I'll close this for now. @blattms: feel free to reopen and merge or to propose something better.

PR #348, opened 2018-07-12
Improve ebos ECL I/O

andlaus:
this adds a `writeOutput()` method to `EclProblem` which extracts all relevant reservoir and well data from the appropriate eWoms objects. `flow` doesn't use this, so it is unaffected.

the second patch removes the `#if HAVE_ECL_{IN,OUT}PUT` guards because `ebos` will not be useful without them. (probably it does not even compile.)


andlaus:
jenkins build this please

andlaus:
Jenkins does not complain. @akva2: Green or any comments?

PR #349, opened 2018-07-18
newton convergence writer: provide a normalized version of the relative error

andlaus:
this makes it easier to inspect the convergence behavior with paraview. Since `flow` currently does not use this code, it is unaffected.


andlaus:
jenkins build this please

PR #352, opened 2018-07-26
ebos: add possibility to disable the FIP output even if requested in the deck

andlaus:
this significantly reduces the noise level of debug files if the FIP values are not of interest.

note that until OPM/opm-simulators#1512 gets merged this feature is not usable from `flow`.


andlaus:
jenkins build this please

atgeirr:
I have thought the same from time to time... Looks fine, will merge.

Somewhat related to this, there is some room for modifying the logging setup based on runtime parameters, controlling which log levels (debug, note, info, warning, error) go to which log stream. This possibility is currently unused. We could for example make the cout stream print all messages including debug by passing a runtime parameter.

PR #351, opened 2018-07-19
Black oil model: improve update

andlaus:
these changes bring the update routines of `ebos` closer to those of `flow`:
- limit pressure changes to 30% of the absolute value
- use a scaling factor for saturations to implement the Appleyard-chop. this is slightly more complicated than the previous code but it has the advantage that changes to the oil phase saturation also get limited.

`flow` currently uses completely separate code paths, so it is unaffected.

PR #353, opened 2018-07-26
fix up the more uncommon convergence criteria for the default linear solver

andlaus:
some header guard macros were not adapted after copying the respective source files and ´WeightedResidualReductionCriterion::update()` was not adapted after the signature of the method changed in the base class. the practical implications of this were limited, because the affected convergence criteria are not used by default. i.e., currently they are purely used for debugging purposes.

as usual, `flow` is unaffected  by these changes. Somewhat unusually, the other binaries in eWoms also are supposed to not care...

### DIFF
--- a/ewoms/linear/combinedcriterion.hh
+++ b/ewoms/linear/combinedcriterion.hh
@@ -22,10 +22,10 @@
 */
 /*!
  * \file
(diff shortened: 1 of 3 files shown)


andlaus:
jenkins build this please

andlaus:
@atgeirr: Do you mind to press Green for this and for #351?

PR #355, opened 2018-07-31
Restore const-ness accidentally lost.

atgeirr:
I see no compilation problems with this, let's see if Jenkins agrees.


andlaus:
thanks.

PR #356, opened 2018-08-01
linear: do not rescale the linear system of equations before solving

andlaus:
while the rescaling does not change anything mathematically, it confuses some preconditioners, in particular those which make assumptions on the semantics of a given equation.

the rescaling could either happen at model level or within a specialized linear solver backend.

`flow` is unaffected by this because it rolls its own stuff when it comes to the linear solver. the eWoms tests seem to exhibit a small performance increase on average. (most tests become faster, though some get slightly slower.)

### DIFF
--- a/ewoms/linear/parallelbasebackend.hh
+++ b/ewoms/linear/parallelbasebackend.hh
@@ -217,8 +217,6 @@ public:
         // the values of all processes (using the assignAdd() methods)
         overlappingMatrix_->assignFromNative(M);
 
-        asImp_().rescale_();
-
         // synchronize all entries from their master processes and add entries on the
         // process border
         overlappingMatrix_->syncAdd();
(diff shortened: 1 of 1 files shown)


andlaus:
jenkins build this please

atgeirr:
OK to merge when green.

andlaus:
has there recently been some changes in `flow` which cause master to fail? I'm not aware of what `BASE_MODEL_1` is. A reduced version of Norne?

akva2:
there has been changes, but master is fine. i do have to wonder if comparing TCPU makes much sense in the first place, however.

andlaus:
> i do have to wonder if comparing TCPU makes much sense in the first place, however.

I vote for "no" :). https://github.com/OPM/opm-tests/pull/40 fixes this.

akva2:
jenkins build this please

akva2:
green, approved.

PR #358, opened 2018-08-06
ebos: complain more explicitly about the input deck being unspecified

andlaus:
the previous error message was a bit confusing.

This should fix one of @atgeirr's itches!?

### DIFF
--- a/ebos/eclbasevanguard.hh
+++ b/ebos/eclbasevanguard.hh
@@ -65,7 +65,7 @@ NEW_PROP_TAG(EclDeckFileName);
 NEW_PROP_TAG(EclOutputDir);
 NEW_PROP_TAG(EclOutputInterval);
 
-SET_STRING_PROP(EclBaseVanguard, EclDeckFileName, "ECLDECK.DATA");
+SET_STRING_PROP(EclBaseVanguard, EclDeckFileName, "");
 SET_STRING_PROP(EclBaseVanguard, EclOutputDir, ".");
 SET_INT_PROP(EclBaseVanguard, EclOutputInterval, -1); // use the deck-provided value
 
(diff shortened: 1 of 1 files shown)


andlaus:
jenkins build this please

ebos/eclbasevanguard.hh
```diff
@@ -171,6 +171,11 @@ public:
 #endif
 
         std::string fileName = EWOMS_GET_PARAM(TypeTag, std::string, EclDeckFileName);
+
+        if (fileName == "")
+            throw std::runtime_error("No input deck file has been specified "
+                                     "via --ecl-deck-file-name='$FOO.DATA'");
```
atgeirr: No-one (probably) specifies the deck like that, I think the error message is most helpful if it shows the usual syntax. I suggest something like: "No input deck file has been specified, use 'flow CASE.DATA' to run the deck in CASE.DATA".

andlaus: > No-one (probably) specifies the deck like that.

we cannot mention the name `flow` here because this code is also used in wider contexts (currently only plain-`ebos`, but if somebody wants to experiment with ECL-decks/CpGrid while using a custom problem class, these would be affected as well). We also cannot rely on that the simulator in question uses positional parameters to specify the deck file (although `flow` and plain-`ebos` do).  I can offer remove the second part of that message outright, but I think it is better to keep it..

atgeirr: > we cannot mention the name flow here because this code is also used in wider contexts (currently only plain-ebos, but if somebody wants to experiment with ECL-decks/CpGrid while using a custom problem class, these would be affected as well).

Ok.

> We also cannot rely on that the simulator in question uses positional parameters to specify the deck file (although flow and plain-ebos do). I can offer remove the second part of that message outright, but I think it is better to keep it..

I think that is a *very* theoretical risk, that we should ignore in favour of having a good error message. But to be safer we can also include the command line option. What about changing the second line to: "as a command line argument, or via --ecl-deck-file-name=CASE.DATA"?

Also, should not the "ecl-" prefix go away, as there should be no conflict with other options?

andlaus:
> What about changing the second line to: "as a command line argument, or via --ecl-deck-file-name=CASE.DATA"?

good point. changed, thanks.

andlaus:
jenkins build this please

PR #359, opened 2018-08-08
Use a custom parameter file parser

andlaus:
this PR switches the parameter file parser to a custom implementation (instead of using `Dune::ParameterTreeParser`) and it also contains a few smaller cleanups of the parameter system which should not affect the developer experience. the switch to a custom parser allows the key names that are put into the parameter files to be transformed, i.e.,

```
enable-vtk=true
```

is now also possible in parameter files whereas before only CamelCase parameter names like

```
EnableVtk=true
```

were valid. besides this, any valid Windows INI file that does not use parameter groups should be readable, i.e., semicolons are now also treated as a comment indicator.

Despite C++'s file and string processing tools being quite blunt IMO, I tried to keep the prerequisites minimal, i.e., this code does not use boost or advanced c++-2011 features like `std::regex`.

### DIFF
--- a/ewoms/common/parametersystem.hh
+++ b/ewoms/common/parametersystem.hh
@@ -388,7 +388,8 @@ void printUsage(const std::string& helpPreamble,
 }
 
 /// \cond 0
-inline int noPositionalParameters_(std::string& errorMsg,
+inline int noPositionalParameters_(std::set<std::string>& seenParams OPM_UNUSED,
+                                   std::string& errorMsg,
                                    int argc OPM_UNUSED,
                                    const char** argv,
                                    int paramIdx,
@@ -549,7 +550,7 @@ std::string parseCommandLineOptions(int argc,
             || argv[i][1] != '-')
         {
             std::string errorMsg;
-            int numHandled = posArgCallback(errorMsg, argc, argv, i, numPositionalParams);
+            int numHandled = posArgCallback(seenKeys, errorMsg, argc, argv, i, numPositionalParams);
 
             if (numHandled < 1) {
                 std::ostringstream oss;
--- a/ewoms/common/parametersystem.hh
+++ b/ewoms/common/parametersystem.hh
@@ -41,6 +41,7 @@
 #include <dune/common/parametertree.hh>
 
 #include <map>
+#include <set>
 #include <list>
 #include <sstream>
 #include <string>
@@ -539,6 +540,7 @@ std::string parseCommandLineOptions(int argc,
         }
     }
 
+    std::set<std::string> seenKeys;
     int numPositionalParams = 0;
     for (int i = 1; i < argc; ++i) {
         // All non-positional command line options need to start with '-'
@@ -588,6 +590,17 @@ std::string parseCommandLineOptions(int argc,
 
         // parse argument
         paramName = transformKey_(parseKey_(s), /*capitalizeFirst=*/true);
+        if (seenKeys.count(paramName) > 0) {
+            std::string msg =
+                std::string("Parameter '")+paramName+"' specified multiple times as a "
+                "command line parameter";
+
+            if (!helpPreamble.empty())
+                printUsage<TypeTag>(helpPreamble, msg, std::cerr);
+            return msg;
+        }
+        seenKeys.insert(paramName);
+
         paramValue = s.substr(1);
 
         // Put the key=value pair into the parameter tree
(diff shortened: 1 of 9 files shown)


andlaus:
jenkins build this please

andlaus:
@atgeirr, @totto82: any objections to merging this? it should not lead to regressions for existing code. (Also note that technically OPM/opm-simulators#1512 is independent of it.)

atgeirr:
I think this is probably what I want, but I have not had the time to test it...

atgeirr:
> (Also note that technically OPM/opm-simulators#1512 is independent of it.)

Yes, I think this should be merged first (assuming it works as promised).

ewoms/common/parametersystem.hh
```diff
@@ -409,6 +399,111 @@ inline int noPositionalParameters_(std::string& errorMsg,
...
+inline void removeLeadingSpace_(std::string& s)
+{
+    unsigned i;
+    for (i = 0; i < s.size(); ++ i)
+        if (!std::isspace(s[i]))
+            break;
+    s = s.substr(i);
+}
+
+inline std::string transformKey_(const std::string& s,
```
atgeirr: I would assume this function already existed to handle the command line arguments, does that mean you refactored the old code as well?

ewoms/common/parametersystem.hh
```diff
@@ -471,107 +569,97 @@ std::string parseCommandLineOptions(int argc,
...
+
+        // TODO (?): support for parameter groups.
+
+        // find the "key" of the key=value pair
+        std::string key = transformKey_(parseKey_(curLine),
+                                        /*capitalizeFirst=*/true,
+                                        errorPrefix);
+
+        // deal with the equals sign
+        removeLeadingSpace_(curLine);
```
atgeirr: This allows spaces between the key and `=`, is that intentional?  Also after the `=` (line 640). Is it allowed in INI files?

ewoms/common/parametersystem.hh
```diff
@@ -471,107 +569,97 @@ std::string parseCommandLineOptions(int argc,
...
+            value = parseQuotedValue_(curLine, errorPrefix);
+        else
+            value = parseUnquotedValue_(curLine, errorPrefix);
+
+        // ignore trailing comments
+        removeLeadingSpace_(curLine);
+        if (!curLine.empty() && curLine[0] != '#' && curLine[0] != ';')
+            std::runtime_error(errorPrefix+"Syntax error, expecting 'key=value'");
+
+        // all went well, add the parameter to the database object
```
atgeirr: When down here, a lot of cutting from the front has happened to `curLine`, so it is probably not very effective. However it does not have to be, so no need to change (famous last words, I know). If at some point we need it to be effective, string_view may be our friend.

andlaus: effectiveness is not relevant here because modern computers are insanely fast, so IMO code simplicity is vastly more important in this context. Say 99% of all time: premature optimization is the root of all evil!

ewoms/common/parametersystem.hh
andlaus: yes, are no spaces allowed there? I was not sure and it does not hurt to support this anyways?

ewoms/common/parametersystem.hh
```diff
@@ -471,107 +569,97 @@ std::string parseCommandLineOptions(int argc,
...
+{
+    Dune::ParameterTree& paramTree = GET_PROP(TypeTag, ParameterMetaData)::tree();
+
+    const int tmpSize = 100*1000;
+    char tmp[tmpSize];
+    std::ifstream ifs(fileName);
+    unsigned curLineNum = 0;
+    while (ifs) {
+        // string and file processing in c++ is quite blunt!
+        ifs.getline(tmp, tmpSize); std::string curLine(tmp);
```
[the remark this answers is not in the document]
andlaus: thanks for the pointer. I did not know about `std::getline()`. fixed. (did I ever mention that I consider C++'s file handling capabilities to be blunt?)

ewoms/common/parametersystem.hh
andlaus: this was directly embedded into the command line parser. this PR factors it out into a separate function...

andlaus:
okay, I've also dealt with a few small things that came up during OPM/opm-simulators#1512.

andlaus:
jenkins build this please

PR #360, opened 2018-08-09
Fix issues in the newton update

totto82:
Add some more tricks from flow used in the newton update to make the solvent simulation fly again
Fix fallout for 2p cases 
I would like to test this a bit more before merging. 

- [x]  Run simulations on the big models together with OPM/opm-simulators#1542

### DIFF
--- a/ewoms/models/blackoil/blackoilnewtonmethod.hh
+++ b/ewoms/models/blackoil/blackoilnewtonmethod.hh
@@ -33,6 +33,14 @@
 #include <ewoms/common/signum.hh>
 
 #include <opm/material/common/Unused.hpp>
+BEGIN_PROPERTIES
+NEW_PROP_TAG(DpMaxRel);
+NEW_PROP_TAG(DsMax);
+NEW_PROP_TAG(RestrictOscilationInPrimaryVariableMeaning);
+SET_SCALAR_PROP(NewtonMethod, DpMaxRel, 0.3);
+SET_SCALAR_PROP(NewtonMethod, DsMax, 0.2);
+SET_BOOL_PROP(NewtonMethod, RestrictOscilationInPrimaryVariableMeaning, true);
+END_PROPERTIES
 
 namespace Ewoms {
 
@@ -58,7 +66,11 @@ class BlackOilNewtonMethod : public GET_PROP_TYPE(TypeTag, DiscNewtonMethod)
 
 public:
     BlackOilNewtonMethod(Simulator& simulator) : ParentType(simulator)
-    { }
+    {
+        unsigned numDof = simulator.vanguard().grid().size(0);
+        wasSwitched_.resize(numDof);
+        std::fill(wasSwitched_.begin(), wasSwitched_.end(), false);
+    }
 
     /*!
      * \brief Register all run-time parameters for the immiscible model.
@@ -66,6 +78,10 @@ public:
     static void registerParameters()
     {
         ParentType::registerParameters();
+        EWOMS_REGISTER_PARAM(TypeTag, Scalar, DpMaxRel, "Maximum relative change of pressure in a single iteration");
+        EWOMS_REGISTER_PARAM(TypeTag, Scalar, DsMax, "Maximum absolute change of any saturation in a single iteration");
+        EWOMS_REGISTER_PARAM(TypeTag, Scalar, RestrictOscilationInPrimaryVariableMeaning,
+                             "Adds a small factor to the primary variable switching conditions after its meaning has switched to hinder oscilations");
     }
 
     /*!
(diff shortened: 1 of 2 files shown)


ewoms/models/blackoil/blackoilnewtonmethod.hh
```diff
@@ -159,24 +163,30 @@ protected:
         Opm::Valgrind::CheckDefined(currentResidual);
 
         // saturation delta for each phase
-        Scalar deltaSw = update[Indices::waterSaturationIdx];
+        Scalar deltaSw = 0.0;
+        Scalar deltaSo = 0.0;
         Scalar deltaSg = 0.0;
-        Scalar deltaSo = -deltaSw;
-        if (currentValue.primaryVarsMeaning() == PrimaryVariables::Sw_po_Sg) {
+        Scalar deltaSs = 0.0;
```
andlaus: is this intentional? the change in solvent "saturation" was consciously lumped together with the gas saturation change because solvent is handled as a gas when it comes to the relperm and the relperms are the reason why we go through this trouble in the first place?

maybe there was a problem with this, though...

ewoms/models/blackoil/blackoilnewtonmethod.hh
```diff
@@ -220,17 +230,37 @@ protected:
 
             // do the actual update
             nextValue[pvIdx] = currentValue[pvIdx] - delta;
+
+            // keep the solvent saturation between 0 and 1
+            if (pvIdx == Indices::solventSaturationIdx)
+                nextValue[pvIdx] = std::min(std::max(nextValue[pvIdx], 0.0), 1.0);
```
andlaus: this should probably be done for all saturations, not just solvent? I suppose that the problem is the implicitly given saturation (So).

ewoms/models/blackoil/blackoilnewtonmethod.hh
```diff
@@ -220,17 +230,37 @@ protected:
...
+            if (pvIdx == Indices::polymerConcentrationIdx)
+                nextValue[pvIdx] = std::max(nextValue[pvIdx], 0.0);
+
         }
 
         // switch the new primary variables to something which is physically meaningful
-        if (nextValue.adaptPrimaryVariables(this->problem(), globalDofIdx))
+        // add an epsilon to make it harder to switch back immediately after the primary variable meaning is changed.
+        if (wasSwitched_[globalDofIdx]) {
+            wasSwitched_[globalDofIdx] = nextValue.adaptPrimaryVariables(this->problem(), globalDofIdx, 1e-5);
```
andlaus: I think having this switching "hysteresis" is okay and also its implementation is fine, but I would prefer not to having to go through that hoop. did you check if it makes a difference?

andlaus:
thanks. regardless of the question of which of these changes are desirable, the patch itself looks good already. One small question: would you mind adding the missing parameters for the thresholds which you've remarked upon in #351?

totto82:
> One small question: would you mind adding the missing parameters for the thresholds which you've 
remarked upon in #351?

No problem. 

I notices a failure in one of the Co2 cases when testing OPM/opm-simulators#1542, adding these tricks seemed to help. Doing the same as Flow seemed like a good starting point. 

Still, there still seems to be an issue with the CO2 case towards the end of the simulations time. I will need to look into that. Lets keep this PR open until all the convergence issues are settled.

andlaus:
>> would you mind adding the missing parameters for the thresholds which you've
remarked upon in #351?
>
> No problem.

cool, thanks.

> I notices a failure in one of the Co2 cases when testing OPM/opm-simulators#1542, adding these tricks seemed to help. Doing the same as Flow seemed like a good starting point.

It would probably a good thing to determine which of these changes did the trick. (If I had to bed, I would put my money on the switching hysteresis, but my the confidence is only ~40% here.)

> Lets keep this PR open until all the convergence issues are settled.

definite agreement! In any case: thanks a lot for all the work you put into this.

ewoms/models/blackoil/blackoilnewtonmethod.hh
totto82: I would like to keep it the "old" way for now.

ewoms/models/blackoil/blackoilnewtonmethod.hh
totto82: Testing shows that it is beneficial to allow the saturation to be negative and instead let the newton solver fix this. For some reason this is not the case for solvent. I guess it is due to the TL modeling.

andlaus: you're right: I forgot that negative hydrocarbon saturations were the PV switching condition.

ewoms/models/blackoil/blackoilnewtonmethod.hh
```diff
@@ -58,14 +66,22 @@ class BlackOilNewtonMethod : public GET_PROP_TYPE(TypeTag, DiscNewtonMethod)
...
+        std::fill(wasSwitched_.begin(), wasSwitched_.end(), false);
+    }
 
     /*!
      * \brief Register all run-time parameters for the immiscible model.
      */
     static void registerParameters()
     {
         ParentType::registerParameters();
+        EWOMS_REGISTER_PARAM(TypeTag, Scalar, DpMaxRel, "Maximum relative change of pressure in a single iteration");
```
andlaus: does this work with the master version of flow? the same parameter name is used there, and `EWOMS_REGISTER_PARAM` will complain if it detects the even smallest change in signature, e.g. if the description is different. Maybe these parameters should be prefixed by `BlackOil` or something.

totto82: I also expected this to fail. But it didn't for some reason. My plan was to remove them in OPM/opm-simulators#1542 and then merge these two PRs together.  If possible I would like to avoid change of parameter names. Do you want me be provide a patch to OPM/opm-simulators#1542 where these parameters are removed, or will you do it?

andlaus:
> My plan was to remove them in OPM/opm-simulators#1542 and then merge these two PRs together.

good idea!

> Do you want me be provide a patch to OPM/opm-simulators#1542 where these parameters are removed, or will you do it?

I can also do this myself, as you prefer. you can also close OPM/opm-simulators#1542 and open a new PR located your fork in order to make coordination easier...

totto82:
> I can also do this myself

Thanks, I would appreciate if you could do it!

ewoms/models/blackoil/blackoilnewtonmethod.hh
andlaus: maybe you used exactly the same description as in opm-simulators, then it should not complain. (this is a hack in the parameter system to avoid inconsistencies yet allowing the `registerParams()` functions to be called multiple times. maybe there is a better way to do this.)

totto82: I just copied it, so it is exactly the same.

akva2:
this broke master, ref https://ci.opm-project.org/job/ewoms/341/

andlaus:
this error is a bit strange: it looks like an illegal memory access, but I cannot see any reason who this PR should cause something like that. maybe it is the static variables inside a function (I doubt it), so maybe #368 fixes it.

andlaus:
ok, #369 hopefully fixes this

PR #361, opened 2018-08-13
Wellconnection updates

joakim-hove:
Downstream of: https://github.com/OPM/opm-common/pull/464

### DIFF
--- a/ebos/eclwellmanager.hh
+++ b/ebos/eclwellmanager.hh
@@ -788,31 +788,9 @@ protected:
                 const Opm::Connection* connection = connInfo.first;
                 std::shared_ptr<Well> eclWell = connInfo.second;
                 eclWell->addDof(elemCtx, dofIdx);
-
-                // the catch is a hack for a ideosyncrasy of opm-parser with regard to
-                // defaults handling: if the deck did not specify a radius for the
-                // completion, there seems to be no other way to detect this except for
-                // catching the exception
-                try {
-                    eclWell->setRadius(elemCtx, dofIdx, 0.5*connection->getDiameter());
-                }
-                catch (const std::logic_error&)
-                {}
-
-                // overwrite the automatically computed effective
-                // permeability by the one specified in the deck-> Note: this
-                // is not implemented by opm-parser yet...
-                /*
-                  Scalar Kh = connection->getEffectivePermeability();
-                  if (std::isfinite(Kh) && Kh > 0.0)
-                      eclWell->setEffectivePermeability(elemCtx, dofIdx, Kh);
-                */
-
-                // overwrite the automatically computed connection
-                // transmissibilty factor by the one specified in the deck->
-                const auto& ctf = connection->getConnectionTransmissibilityFactorAsValueObject();
-                if (ctf.hasValue() && ctf.getValue() > 0.0)
-                    eclWell->setConnectionTransmissibilityFactor(elemCtx, dofIdx, ctf.getValue());
+                eclWell->setConnectionTransmissibilityFactor(elemCtx, dofIdx, connection->CF());
+                eclWell->setRadius(elemCtx, dofIdx, connection->rw());
+                //eclWell->setEffectivePermeability(elemCtx, dofIdx, connection->Kh());
             }
         }
     }


ebos/eclwellmanager.hh
```diff
@@ -788,31 +788,9 @@ protected:
...
-                  if (std::isfinite(Kh) && Kh > 0.0)
-                      eclWell->setEffectivePermeability(elemCtx, dofIdx, Kh);
-                */
-
-                // overwrite the automatically computed connection
-                // transmissibilty factor by the one specified in the deck->
-                const auto& ctf = connection->getConnectionTransmissibilityFactorAsValueObject();
-                if (ctf.hasValue() && ctf.getValue() > 0.0)
-                    eclWell->setConnectionTransmissibilityFactor(elemCtx, dofIdx, ctf.getValue());
+                eclWell->setConnectionTransmissibilityFactor( elemCtx, dofIdx, connection->CF );
```
andlaus: if this was not clear from `setConnectionTransmissibilityFactor` of the downstream object what does `CF` stand for in `connection->CF`? (this is mainly an upstream issue.) also shouldn't this be a getter method? this is C++ after all...

finally: please watch your spaces (this is only a minor nit which won't block merging from my side)

joakim-hove: Thank you - I will address your comments; the upstream PR is still very much work in progress, and this PR was set up mainly to complete the chain.

andlaus:
merge whenever

ebos/eclwellmanager.hh
andlaus: > > also shouldn't this be a getter method? this is C++ after all...
> 
> Maybe; I have to some extent changed my mind on this. The values in a `Connection` object are all "immutable"[1] - and never assigned to after the constructor.

For of C++/OOP pundits the "correct" answer is almost certainly a resounding "definitely!":  even if the value is immutable, it might be calculated on the fly, or it might retrieved from another member object under some conditions and also it is opaque for mere mortals what data is supposed to be accessed via getter methods and what directly. getter methods make it possible to hide such implementation details; For example, in OPM this has become very handy for the `Evaluation` class where the storage layout was changed significantly in order to take advantage of SIMD instructions but at the same time the user code did not need to change at all.

> [1]: They are unfortunately not immutable in C++ sense, because the objects are stored in a `std::vector<>`.

another reason to use getters? In C++, people are allowed to shoot themselves into their feet, but it is IMO bad API design to make this easy.

joakim-hove: I do not disagree with your points - but I also feel that the simplification of the now refactored `Connection` class - which is essentially a plain `C` struct now (I think ...) is valuable; and the pattern:

```
public:
    T getFoo() const { return this->intrenal_foo; }

private:
    T internal_foo;
```

is kind of klunky; I guess an important question is whether you foresee the implementation of `foo` to be changed in the furure - e.g. calculated on the fly. In my case I see the `Connection` class to have reached it's low-entropy endpoint, and therefor do *not* see future changes as probable. If you can get two other developer to support you (preferably in the upstream PR - and: **No beer bribing**) I will change to use getter methods.

andlaus: > I do not disagree with your points - but I also feel that the simplification of the now refactored Connection class - which is essentially a plain C struct now (I think ...) is valuable;

do you mean "simplification" as "the implementation in opm-common is simpler" or in terms of "usage of this gets easier?". the former is probably correct, the latter isn't. (I always need to go to the corresponding header file to look up which class is supposed to be a C struct-with-methods and which is a "proper" class.)

> the pattern [...] is kind of klunky

I agree, but IMO it is only a little bit clunky, so the benefits of this pattern clearly outweigh its cost IMO. you've probably been hacking in C for too long to be fully converted to the church of OOP yet ;)

> I guess an important question is whether you foresee the implementation of foo to be changed in the furure

I wouldn't be so sure. I'd like to stick with Niels Bohr in this regard: "It's Difficult to Make Predictions, Especially About the Future."

but as I have said earlier: I don't consider this to be my shot to call. if you want to bear the grunt of having to change all downstreams when you change something there, I won't stop you.

PR #362, opened 2018-08-14
make the "OutputDir" parameter apply universally

andlaus:
all disc output, i.e. VTK, restart files, ECL and -- in the future -- logfiles, goes to that directory. before this, only the ECL output could be directed to a different than the current working directory and the parameter for this was called "EclOutputDir".

note that the Dune VTK writing infrastructure makes it harder than it needs to be: suddenly Dune::VTKWriter::write() does not work in parallel anymore, but Dune::VTKWriter::pwrite() must be called with the right arguments.

this PR requires a small mop-up PR in opm-simulators.


andlaus:
jenkins build this opm-simulators=1550  please

andlaus:
@atgeirr: merge?

PR #363, opened 2018-08-14
add a the infrastructure for tab-completion of parameters in bash

andlaus:
the file does create the functions which can be used for completion, but they must still be registered manually.

this superseeds OPM/opm-utilities#37. the idea is that downstream modules like opm-simulators can easily ship ready-to use completors for their products with the respective distribution packages. credits for the idea goes to @akva2.


andlaus:
@akva2: merge?

akva2:
we still need to handle the installation etc but i can look into that when doing the packaging.

andlaus:
yes. the idea is that downstream supply the boiler plate to register `_ewomsParameterCompletions()` for the products which they ship. for this to make sense for `opm-simulators` to make sense, OPM/opm-simulators#1512 needs to be merged first, though.

akva2:
sure, but ebos is a product as well.

andlaus:
okay, I've made the mechanism more convenient and robust and added a file for product completors that can to be installed by distribution packages. if nothing further crops up, I think this PR should be merged in its current state.

andlaus:
@akva2: mind to press Green? the idea is to deliver something analogous to `ewoms_product_completors.sh` with the packages for downstream modules. do you consider this approach to be workable?

PR #364, opened 2018-08-16
Parameter system: Improve usage message

andlaus:
This is a small epilogue to OPM/opm-simulators#1512. It prints the default values for all parameters in the usage message and on TTYs, it breaks lines nicely. (the latter is surprisingly hard to implement.)

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -445,9 +445,9 @@ public:
     {
         if (briefDescription_.empty())
             return
-                "The Ecl-deck Black-Oil reservoir Simulator (ebos); a hydrocarbon\n"
-                "reservoir simulation program that processes ECL-formatted input\n"
-                "files which is provided by the Open Porous Media project\n"
+                "The Ecl-deck Black-Oil reservoir Simulator (ebos); a hydrocarbon "
+                "reservoir simulation program that processes ECL-formatted input "
+                "files which is provided by the Open Porous Media project "
                 "(https://opm-project.org).";
         else
             return briefDescription_;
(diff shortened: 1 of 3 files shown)


andlaus:
jenkins build this please

andlaus:
> the message received was 'Parameter 'EclDeckFileName' specified multiple times as a command line parameter'.

I'll have a look. what's the full command that failed?

> i had the old output_dir=foo style parameters.

that explains it: if parameters don't start with `--` they are interpreted as positional parameters, and flow/ebos interprets them as the file name of the deck (`--ecl-deck-file-name`)

> the second thing i think should be support is that if i just invoke 'flow', i get a nasty exception,

I'll have a look. The usage message should be printed here...

akva2:
i know it's wrong and i understand what happened. just trying to improve user experience (and lessen support load..), so i think this case should be special-cased at least for one release.

andlaus:
OPM/opm-simulators#1552 should fix the first issue. I'm not sure about what to do about the second. maybe modifying the error message if multiple positional parameters are specified?

akva2:
```c++
for (int i = 1; i < argc; ++i)
if (!StartsWith(argv[i],--) && Contains(argv[i], "="))
  errorMessage;
```
is pseudo for what i would do. keep it for one release.

andlaus:
> is pseudo for what i would do. keep it for one release.

okay, seems to be workable. patch added.

andlaus:
jenkins build this please

akva2:
thank you

andlaus:
@akva2: merge?

PR #367, opened 2018-08-21
Fix reported time for initial timestep

totto82:


andlaus:
isn't `simulator.timeStepSize()` for the initial write zero? if not, should it be?

andlaus:
>  Anyway I found a solution that only touch opm-simulators. I is still a bit hacky but better than this!

is `problem_->writeOutput()` currently not called correctly in `flow`?

PR #368, opened 2018-08-21
black-oil: minor fixes to the update method

andlaus:
this primarily updates the comments in `BlackOilNewtonMethod::update()` and it prevents using `EWOMS_GET_PARAM()` in a fast path: `EWOMS_GET_PARAM()` needs to traverse a `std::map` of strings every time, so it should not be used in highly performance critical code. Given that, it is surprisingly fast, so the performance improvement of this is probably quite small.


andlaus:
jenkins build this please

andlaus:
jenkins build this please

andlaus:
I discovered a screwup of the type of the `RestrictOscilationInPrimaryVariableMeaning` property and the corresponding parameter: the property was a boolean while the parameter was a scalar. I renamed it to `PriVarOscilationThreshold` and made it a scalar...

andlaus:
jenkins build this please

andlaus:
jenkins build this please

andlaus:
I was wondering why this still fails. then I noticed that the parameters in master are [reverse](https://github.com/OPM/ewoms/blob/master/ewoms/models/blackoil/blackoilnewtonmethod.hh#L177-L178) relative to the variable names and how they are used.

there are two ways to fix this: either we update the reference data or we use 0.3 for the maximum saturation change as well as 0.2 as the maximum relative pressure change (instead of vice-versa). @totto82: what do you prefer?

totto82:
I suggest using what was in flow i.e. the following fix

```
--static const Scalar maxSaturationChange = EWOMS_GET_PARAM(TypeTag, Scalar, DpMaxRel);; // per iteration
static const Scalar maxSaturationChange = EWOMS_GET_PARAM(TypeTag, Scalar, DsMax);; // per iteration
--static const Scalar maxPressureChange = EWOMS_GET_PARAM(TypeTag, Scalar, DsMax); // relative, per iteration
static const Scalar maxPressureChange = EWOMS_GET_PARAM(TypeTag, Scalar, DpMaxRel); // relative, per iteration
```
and then update the reference data if necessary.  Interestingly this unintentional interchange had some positive impact on the convergence on some cases. Maybe we should re-investigate these tolerances one day.  

> I discovered a screwup of the type of the RestrictOscilationInPrimaryVariableMeaning property and the corresponding parameter: the property was a boolean while the parameter was a scalar. I renamed it to PriVarOscilationThreshold and made it a scalar...

I first made it a Scalar and then changed it to boolean to avoid users to tweak this parameter, but I am fine having it has a Scalar as well. It will make testing different thresholds easier. 

@andlaus Thanks fixing this!

andlaus:
alright, let's do a data update then. I also think this is the slightly better option.

andlaus:

totto82:
Thanks, I will merge this when Jenkins is happy.

totto82:
jenkins build this opm-tests=50 please

PR #369, opened 2018-08-21
black-oil model: make it generic again

andlaus:
The number of DOFs is only equal to the number of elements in the ECFV discretization. (note that this patch uses numDof() while numGridDof() would be sufficient. since it is numGridDof() + numAuxDof(), the proposed solution is thus slightly more defensive.)


andlaus:
jenkins build this please

andlaus:
jenkins build this please

PR #370, opened 2018-08-27
Default number of threads to auto

akva2:
In particular this restores multithreading in e.g. flow without having to use --threads-per-process.
It still honors threads per process, and it still honors OMP_NUM_THREADS. so everybody should be satisfied.

Note I had to disable multithreading in some ewoms tests, as they break with it enabled.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -166,10 +166,10 @@ opm_add_test(waterair_pvs_ni
              TEST_ARGS --grid-global-refinements=1)
 
 opm_add_test(lens_immiscible_vcfv_ad
-             TEST_ARGS --end-time=3000)
+             TEST_ARGS --end-time=3000 --threads-per-process=1)
 
 opm_add_test(lens_immiscible_vcfv_fd
-             TEST_ARGS --end-time=3000)
+             TEST_ARGS --end-time=3000 --threads-per-process=1)
 
 opm_add_test(lens_immiscible_ecfv_ad
              TEST_ARGS --end-time=3000)
@@ -235,10 +235,10 @@ foreach(tapp co2injection_flash_ni_vcfv
   opm_add_test(${tapp})
 endforeach()
 
-opm_add_test(reservoir_blackoil_vcfv TEST_ARGS --end-time=8750000)
+opm_add_test(reservoir_blackoil_vcfv TEST_ARGS --end-time=8750000 --threads-per-process=1)
 opm_add_test(reservoir_blackoil_ecfv TEST_ARGS --end-time=8750000)
-opm_add_test(reservoir_ncp_vcfv TEST_ARGS --end-time=8750000)
-opm_add_test(reservoir_ncp_ecfv TEST_ARGS --end-time=8750000)
+opm_add_test(reservoir_ncp_vcfv TEST_ARGS --end-time=8750000 --threads-per-process=1)
+opm_add_test(reservoir_ncp_ecfv TEST_ARGS --end-time=8750000 --threads-per-process=1)
 
 opm_add_test(fracture_discretefracture
              CONDITION ${DUNE_ALUGRID_FOUND}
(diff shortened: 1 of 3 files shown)


atgeirr:
Great! Making all ewoms tests run multi-threaded can be done afterwards.

atgeirr:
jenkins build this opm-common=473 run norne please

andlaus:
now *all* cores are used by default, i.e. for MPI runs on N cores, N^2 threads are spawned. not sure if this is intended. If this is just the behavior which is desired for `flow` it is better to set
```c++
SET_INT_PROP(EclFlowProblem, ThreadsPerProcess, -1);
```
in `opm-simulators/opm/autodiff/BlackoilModelEbos.hpp` instead of setting it globally.

blattms:
I guess the problem is that the default value needs to be set at compile time. Therefore the old behavior ( 2 cores, if there are at least two, 1 otherwise) can only be achieved by fiddling around with the command line parameters in FlowMainEbos.hpp

andlaus:
The number of threads can also be set at run time, the default is just specified at compile time. you can achieve the previous `flow` default behavior of two threads per process by simply setting the `ThreadsPerProcess` property to 2 as @totto82 mentioned in a recent comment for OPM/opm-simulators#1512.

As I understand this at this point, the issue is that having a second, completely independent, way to specify the same thing (the `OMP_NUM_THREADS` environment variable) screwed things up for some people.

atgeirr:
With the latest commit behaviour should be like this:
 - If the command line parameter is set, it is used.
 - Else, if OMP_NUM_THREADS is set, use that.
 - If none of those are set, set to 2 (or 1 if only one core is available).

I think that should be the sweet spot at this time.

atgeirr:
jenkins build this opm-common=473 run norne please

alfbr:
> As I understand this at this point, the issue is that having a second, completely independent, way to specify the same thing (the OMP_NUM_THREADS environment variable) screwed things up for some people.

I am afraid your understanding is wrong. As you should already know, the first breakage was that you broke compiling with openmp:
https://github.com/OPM/opm-simulators/issues/1556
as you also know, the second issue was that you broke setting default number of threads:
https://github.com/OPM/opm-simulators/pull/1515
Breakage of OpenMP environment variables has not bitten anybody yet afaik.

andlaus:
Ok, I'll merge once jenkins is happy.

> Note I had to disable multithreading in some ewoms tests, as they break with it enabled.

I just tested this: on my machine (12 threads) I did not get any failures for ewoms tests with `-1` for `ThreadsPerProcess`.

akva2:
well, i do.

________________________________
Fra: Andreas Lauser <notifications@github.com>
Sendt: mandag 27. august 2018 13.09.14
Til: OPM/ewoms
Kopi: Arne Morten Kvarving; Author
Emne: Re: [OPM/ewoms] Default number of threads to auto (#370)


Ok, I'll merge once jenkins is happy.

Note I had to disable multithreading in some ewoms tests, as they break with it enabled.

I just tested this: on my machine (12 threads) I did not get any failures for ewoms tests with -1 for ThreadsPerProcess.

andlaus:
weird...

akva2:
well, 64 threads are more likely to show races than 2.

andlaus:
> the second issue was that you broke setting default number of threads: OPM/opm-simulators#1515. Breakage of OpenMP environment variables has not bitten anybody yet afaik.

specifying `--threads-per-process=$N` works for master `flow` after OPM/opm-simulators#1557 has been merged. the only thing which was different to the state before was setting default of this to 2. (this got lost in transition because OPM/opm-simulators#1512 was open for *way* too long)

andlaus:
> yes. i did not have time to debug, but they crash consistently.

Ok. This should be fixed before the next release, buts let's take chances for now and merge this. thanks for the patch.

akva2:
https://ci.opm-project.org/job/ewoms/344/ <- additional multithreading related test failure.

PR #371, opened 2018-08-27
Add adjoint, rebased

atgeirr:
This replaces #287. A companion PR in opm-simulators will follow.

While not work in progress in the sense that the feature is done, its implementation must be reviewed and discussed.

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -237,6 +237,10 @@ SET_SCALAR_PROP(EclBaseProblem, NewtonMaxError, 0.1);
 // step succeeds at more than 14 Newton iteration is rather small
 SET_INT_PROP(EclBaseProblem, NewtonMaxIterations, 14);
 
+// set no adjoint variables ad default
+NEW_PROP_TAG(numAdjoint);
+SET_INT_PROP(EclBaseProblem, numAdjoint, 0);
+
 // also, reduce the target for the "optimum" number of Newton iterations to 6. Note that
 // this is only relevant if the time step is reduced from the report step size for some
 // reason. (because ebos first tries to do a report step using a single time step.)
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -262,7 +262,7 @@ SET_STRING_PROP(EclBaseProblem, OutputDir, ".");
 SET_BOOL_PROP(EclBaseProblem, EnableIntensiveQuantityCache, true);
 
 // the cache for the storage term can also be used and also yields a decent speedup
-SET_BOOL_PROP(EclBaseProblem, EnableStorageCache, true);
+SET_BOOL_PROP(EclBaseProblem, EnableStorageCache, false);
 
 // Use the "velocity module" which uses the Eclipse "NEWTRAN" transmissibilities
 SET_TYPE_PROP(EclBaseProblem, FluxModule, Ewoms::EclTransFluxModule<TypeTag>);
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -262,7 +262,7 @@ SET_STRING_PROP(EclBaseProblem, OutputDir, ".");
 SET_BOOL_PROP(EclBaseProblem, EnableIntensiveQuantityCache, true);
 
 // the cache for the storage term can also be used and also yields a decent speedup
-SET_BOOL_PROP(EclBaseProblem, EnableStorageCache, false);
+SET_BOOL_PROP(EclBaseProblem, EnableStorageCache, true);
 
 // Use the "velocity module" which uses the Eclipse "NEWTRAN" transmissibilities
 SET_TYPE_PROP(EclBaseProblem, FluxModule, Ewoms::EclTransFluxModule<TypeTag>);
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -622,11 +622,14 @@ public:
      * \param res The deserializer object
      */
     template <class Restarter>
-    void deserialize(Restarter& res)
+    void deserialize(Restarter& res, bool isOnRestart)
     {
         // reload the current episode/report step from the deck
-        beginEpisode(/*isOnRestart=*/true);
-
+        //beginEpisode(/*isOnRestart=*/true);
+        // should probably be removed
+        if(isOnRestart){
+            beginEpisode(/*isOnRestart=*/true);
+        }
         // deserialize the wells
         wellModel_.deserialize(res);
     }
--- a/ewoms/common/simulator.hh
+++ b/ewoms/common/simulator.hh
@@ -350,11 +350,15 @@ public:
      */
     Scalar timeStepSize() const
     {
+        return timeStepSize_;
+
+        /* This should be handeled by ghe time stepper
         Scalar maximumTimeStepSize =
             std::min(episodeMaxTimeStepSize(),
                      std::max( Scalar(0), endTime() - this->time()));
 
         return std::min(timeStepSize_, maximumTimeStepSize);
+        */
     }
 
     /*!
@@ -835,6 +839,7 @@ public:
             << episodeLength_ << " "
             << startTime_ << " "
             << time_ << " "
+            << timeStepSize_ << " "
             << timeStepIdx_ << " ";
         restarter.serializeSectionEnd();
     }
@@ -856,6 +861,7 @@ public:
             >> episodeLength_
             >> startTime_
             >> time_
+            >> timeStepSize_
             >> timeStepIdx_;
         restarter.deserializeSectionEnd();
     }
--- a/ewoms/common/simulator.hh
+++ b/ewoms/common/simulator.hh
@@ -817,6 +817,10 @@ public:
         typedef Ewoms::Restart Restarter;
         Restarter res;
         res.deserializeBegin(*this, t);
+        if (gridView().comm().rank() == 0)
+            std::cout << "Deserialize file '" << res.fileName() << "'"
+                      << ", next time step size: " << timeStepSize()
+                      << "\n" << std::flush;
         this->deserialize(res);
         problem_->deserialize(res, false);
         model_->deserialize(res);
--- a/ewoms/common/simulator.hh
+++ b/ewoms/common/simulator.hh
@@ -550,7 +550,7 @@ public:
             if (verbose_)
                 std::cout << "Deserialize from file '" << res.fileName() << "'\n" << std::flush;
             this->deserialize(res);
-            problem_->deserialize(res);
+            problem_->deserialize(res, true);// to things for restart
             model_->deserialize(res);
             res.deserializeEnd();
             if (verbose_)
@@ -808,6 +808,16 @@ public:
         res.serializeEnd();
     }
 
+    void deserializeAll(Scalar t)
+    {
+        typedef Ewoms::Restart Restarter;
+        Restarter res;
+        res.deserializeBegin(*this, t);
+        this->deserialize(res);
+        problem_->deserialize(res, false);
+        model_->deserialize(res);
+        res.deserializeEnd();
+    }
     /*!
      * \brief Write the time manager's state to a restart file.
      *
--- a/ewoms/disc/common/fvbaseelementcontext.hh
+++ b/ewoms/disc/common/fvbaseelementcontext.hh
@@ -100,6 +100,7 @@ public:
         enableStorageCache_ = EWOMS_GET_PARAM(TypeTag, bool, EnableStorageCache);
         stashedDofIdx_ = -1;
         focusDofIdx_ = -1;
+        focusTimeIdx_ = 0;
     }
 
     static void *operator new(size_t size) {
@@ -261,7 +262,8 @@ public:
      */
     void setFocusDofIndex(unsigned dofIdx)
     { focusDofIdx_ = dofIdx; }
-
+    void setFocusTimeIndex(unsigned timeIdx)
+    { focusTimeIdx_ = timeIdx; }
     /*!
      * \brief Returns the degree of freedom on which the simulator is currently "focused" on
      *
@@ -269,6 +271,9 @@ public:
      */
     unsigned focusDofIndex() const
     { return focusDofIdx_; }
+
+    unsigned focusTimeIndex() const
+    { return focusTimeIdx_; }
 
     /*!
      * \brief Return a reference to the simulator.
@@ -583,7 +588,7 @@ protected:
 #endif
 
         dofVars_[dofIdx].priVars[timeIdx] = priVars;
-        dofVars_[dofIdx].intensiveQuantities[timeIdx].update(/*context=*/asImp_(), dofIdx, timeIdx);
+        dofVars_[dofIdx].intensiveQuantities[timeIdx].update(/*context=*/asImp_(), dofIdx, timeIdx, focusTimeIdx_);
     }
 
     IntensiveQuantities intensiveQuantitiesStashed_;
@@ -601,6 +606,7 @@ protected:
 
     int stashedDofIdx_;
     int focusDofIdx_;
+    int focusTimeIdx_;
     bool enableStorageCache_;
 };
 
--- a/ewoms/disc/common/fvbaselinearizer.hh
+++ b/ewoms/disc/common/fvbaselinearizer.hh
@@ -163,10 +163,10 @@ public:
      *
      * This means the spatial domain plus all auxiliary equations.
      */
-    void linearize()
+    void linearize(unsigned focustimeIndex)
     {
-        linearizeDomain();
-        linearizeAuxiliaryEquations();
+        linearizeDomain(focustimeIndex);
+        linearizeAuxiliaryEquations(focustimeIndex);
     }
 
     /*!
@@ -179,7 +179,7 @@ public:
      * The current state of affairs (esp. the previous and the current solutions) is
      * represented by the model object.
      */
-    void linearizeDomain()
+    void linearizeDomain(unsigned focustimeIndex)
     {
         // we defer the initialization of the Jacobian matrix until here because the
         // auxiliary modules usually assume the problem, model and grid to be fully
@@ -189,7 +189,7 @@ public:
 
         int succeeded;
         try {
-            linearize_();
+            linearize_(focustimeIndex);
             succeeded = 1;
         }
         catch (const std::exception& e)
@@ -225,7 +225,7 @@ public:
      * \brief Linearize the part of the non-linear system of equations that is associated
      *        with the spatial domain.
      */
-    void linearizeAuxiliaryEquations()
+    void linearizeAuxiliaryEquations(OPM_UNUSED unsigned focustimeIndex)
     {
         auto& model = model_();
         const auto& comm = simulator_().gridView().comm();
@@ -434,7 +434,7 @@ private:
     }
 
     // linearize the whole system
-    void linearize_()
+    void linearize_(unsigned focustimeindex)
     {
         resetSystem_();
 
@@ -474,7 +474,7 @@ private:
                 if (!linearizeNonLocalElements && elem.partitionType() != Dune::InteriorEntity)
                     continue;
 
-                linearizeElement_(elem);
+                linearizeElement_(elem, focustimeindex);
             }
         }
 
@@ -482,7 +482,7 @@ private:
     }
 
     // linearize an element in the interior of the process' grid partition
-    void linearizeElement_(const Element& elem)
+    void linearizeElement_(const Element& elem, unsigned focustimeindex)
     {
         unsigned threadId = ThreadManager::threadId();
 
@@ -490,6 +490,7 @@ private:
         auto& localLinearizer = model_().localLinearizer(threadId);
 
         // the actual work of linearization is done by the local linearizer class
+        elementCtx->setFocusTimeIndex(focustimeindex);
         localLinearizer.linearize(*elementCtx, elem);
 
         // update the right hand side and the Jacobian matrix
--- a/ewoms/disc/common/fvbaselocalresidual.hh
+++ b/ewoms/disc/common/fvbaselocalresidual.hh
@@ -491,8 +491,8 @@ protected:
                           const ElementContext& elemCtx) const
     {
         EvalVector tmp;
-        //EqVector tmp2;
-        EvalVector tmp2;
+        EqVector tmp2;
+        EvalVector tmp2_der;// this is added to support calculation of storage term derivative focusTime!=0 for need for adjoint: always used if cachedStororage term si false
         RateVector sourceRate;
 
         tmp = 0.0;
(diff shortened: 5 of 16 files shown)


andlaus:
as a prelude to the actual review: the series seems to be the development branch without any prior cleanups. This makes it hard to ingest them IMO, so please make the individual patches sensible and self-contained. (e.g., squash everything together if you don't want to go through the hassle of reworking a series. That said, my personal experience is that I usually discover quite a few issues while doing the cleanup.)

ebos/eclproblem.hh
```diff
@@ -237,6 +237,10 @@ SET_SCALAR_PROP(EclBaseProblem, NewtonMaxError, 0.1);
 // step succeeds at more than 14 Newton iteration is rather small
 SET_INT_PROP(EclBaseProblem, NewtonMaxIterations, 14);
 
+// set no adjoint variables ad default
+NEW_PROP_TAG(numAdjoint);
+SET_INT_PROP(EclBaseProblem, numAdjoint, 0);
```
andlaus: capital `n` for property names please: properties should follow the same naming convention as classes.

also: is the adjoint functionality limited to the EclProblem or does it also affect the blackoil model? if the latter, the property definition should move to `blackoilproperties.hh` and setting the default value should go to `blackoilmodel.hh`

ebos/eclproblem.hh
```diff
@@ -622,11 +626,14 @@ public:
      * \param res The deserializer object
      */
     template <class Restarter>
-    void deserialize(Restarter& res)
+    void deserialize(Restarter& res, bool isOnRestart)
```
andlaus: Can you explain what this parameter does? maybe it is better to invert the name (e.g., rename to `isAdjointMode`)?

ewoms/common/simulator.hh
```diff
@@ -350,11 +350,15 @@ public:
      */
     Scalar timeStepSize() const
     {
+        return timeStepSize_;
```
andlaus: we need to find a better solution here. if it is done like this, episode boundaries get crossed by time steps..

ewoms/common/simulator.hh
```diff
@@ -550,7 +554,7 @@ public:
             if (verbose_)
                 std::cout << "Deserialize from file '" << res.fileName() << "'\n" << std::flush;
             this->deserialize(res);
-            problem_->deserialize(res);
+            problem_->deserialize(res, true);// to things for restart
```
andlaus: add a comment what the boolean argument does, e.g. `problem_->deserialize(res, /*onRestart=*/true);`

ewoms/common/simulator.hh
```diff
@@ -808,6 +812,20 @@ public:
         res.serializeEnd();
     }
 
+    void deserializeAll(Scalar t)
```
andlaus: the `deserialize()` method is supposed to deserialize the whole object, i.e., if it is different from this, this is most likely a bug. maybe I misunderstood the purpose of this method, though...

ewoms/disc/common/fvbaseelementcontext.hh
```diff
@@ -261,7 +262,8 @@ public:
      */
     void setFocusDofIndex(unsigned dofIdx)
     { focusDofIdx_ = dofIdx; }
-
+    void setFocusTimeIndex(unsigned timeIdx)
```
andlaus: please add a doxygen comment.

ewoms/disc/common/fvbaseelementcontext.hh
```diff
@@ -270,6 +272,9 @@ public:
     unsigned focusDofIndex() const
     { return focusDofIdx_; }
 
+    unsigned focusTimeIndex() const
```
andlaus: doxygen comment is missing.

ewoms/disc/common/fvbaselinearizer.hh
```diff
@@ -163,10 +163,10 @@ public:
      *
      * This means the spatial domain plus all auxiliary equations.
      */
-    void linearize()
+    void linearize(unsigned focustimeIndex)
```
andlaus: please make the argument default to `0` and use the usual naming convention (`focusTimeIdx`)

ewoms/disc/common/fvbaselinearizer.hh
```diff
@@ -179,7 +179,7 @@ public:
      * The current state of affairs (esp. the previous and the current solutions) is
      * represented by the model object.
      */
-    void linearizeDomain()
+    void linearizeDomain(unsigned focustimeIndex)
```
andlaus: ditto

ewoms/disc/common/fvbaselinearizer.hh
```diff
@@ -225,7 +225,7 @@ public:
      * \brief Linearize the part of the non-linear system of equations that is associated
      *        with the spatial domain.
      */
-    void linearizeAuxiliaryEquations()
+    void linearizeAuxiliaryEquations(OPM_UNUSED unsigned focustimeIndex)
```
andlaus: ditto

ewoms/disc/common/fvbaselinearizer.hh
```diff
@@ -434,7 +434,7 @@ private:
     }
 
     // linearize the whole system
-    void linearize_()
+    void linearize_(unsigned focustimeindex)
```
andlaus: parameter naming. since this is an internal method, there's little point in defaulting it.

ewoms/disc/common/fvbaselocalresidual.hh
```diff
@@ -492,6 +492,7 @@ protected:
     {
         EvalVector tmp;
         EqVector tmp2;
+        EvalVector tmp2_der;// this is added to support calculation of storage term derivative focusTime!=0 for need for adjoint: always used if cachedStororage term si false
```
andlaus: naming convention: `tmp2Der`, or even better `adjointStorageDerivative`

ewoms/disc/common/fvbaselocalresidual.hh
```diff
@@ -559,17 +560,32 @@ protected:
...
-            for (unsigned eqIdx = 0; eqIdx < numEq; ++eqIdx) {
-                tmp[eqIdx] -= tmp2[eqIdx];
-                tmp[eqIdx] *= scvVolume / elemCtx.simulator().timeStepSize();
-
-                residual[dofIdx][eqIdx] += tmp[eqIdx];
+            if( !elemCtx.enableStorageCache() ){
+                // Use the implicit Euler time discretization
+                for (unsigned eqIdx = 0; eqIdx < numEq; ++eqIdx) {
+                    if(elemCtx.focusTimeIndex()==0){
+                        tmp2_der[eqIdx] .clearDerivatives();// remove derivatives for the time index
```
andlaus: this will not work is `Evaluation` is a POD type. the generic way to get rid of the derivatives is:

```c++
x = Opm::getValue(x);
```

but what is this needed for here?

ewoms/disc/common/fvbaseprimaryvariables.hh
```diff
@@ -87,9 +87,9 @@ public:
      * it represents the a constant f = x_i. (the difference is that in the first case,
      * the derivative w.r.t. x_i is 1, while it is 0 in the second case.
      */
-    Evaluation makeEvaluation(unsigned varIdx, unsigned timeIdx) const
+    Evaluation makeEvaluation(unsigned varIdx, unsigned timeIdx,unsigned focustimeidx) const
```
andlaus: parameter naming convention: `focusTimeIdx`

ewoms/disc/ecfv/ecfvdiscretization.hh
```diff
@@ -202,7 +202,7 @@ public:
     void deserialize(Restarter& res)
     {
         res.template deserializeEntities</*codim=*/0>(asImp_(), this->gridView_);
-        this->solution(/*timeIdx=*/1) = this->solution(/*timeIdx=*/0);
+        this->solution(/*timeIdx=*/1) = this->solution(/*timeIdx=*/0);// may wrong and only valid for restart
```
andlaus: I don't really understand that comment. maybe a `deserializeAdjoint()` method should be added which does not do this?

ewoms/io/restart.hh
```diff
@@ -31,6 +31,7 @@
 #include <fstream>
 #include <iostream>
 #include <sstream>
+# include <boost/filesystem.hpp>
```
andlaus: this is a place  where I strictly oppose using boost: it is okay for `ebos` because boost is required for it anyway, but not for generic infrastructure code that is used in non-ECL contexts.

if there is really no other practical way to achieve what you need, at least add `#if HAVE_BOOST_FILESYSTEM` guards...

ewoms/io/restart.hh
```diff
@@ -69,22 +70,33 @@ class Restart
...
+        std::string dir = simulator.problem().outputDir();
         if (dir == ".")
             dir = "";
         else if (!dir.empty() && dir.back() != '/')
             dir += "/";
-
-        int rank = gridView.comm().rank();
+        namespace fs = boost::filesystem;
+        fs::path output_dir(dir);
+        fs::path subdir("ebos_restart");
```
andlaus: is it really necessary to create a subdirectory here?

ewoms/io/restart.hh
```diff
@@ -101,9 +113,7 @@ public:
     void serializeBegin(Simulator& simulator)
     {
         const std::string magicCookie = magicRestartCookie_(simulator.gridView());
-        fileName_ = restartFileName_(simulator.gridView(),
-                                     simulator.problem().outputDir(),
-                                     simulator.problem().name(),
+        fileName_ = restartFileName_(simulator,
```
andlaus: nice change!

ewoms/models/blackoil/blackoilenergymodules.hh
```diff
@@ -351,13 +351,14 @@ public:
...
-                            unsigned timeIdx)
+                            unsigned timeIdx,
+                            unsigned focusTimeIdx)
     {
         auto& fs = asImp_().fluidState_;
         const auto& priVars = elemCtx.primaryVars(dofIdx, timeIdx);
 
         // set temperature
-        fs.setTemperature(priVars.makeEvaluation(temperatureIdx, timeIdx));
+        fs.setTemperature(priVars.makeEvaluation(temperatureIdx, timeIdx, focusTimeIdx));
```
andlaus: you can use `elemCtx.focusTimeIndex()`. since this is an implementation detail of the blackoil model, I don't consider this too important, though.

ewoms/models/blackoil/blackoilindices.hh
```diff
@@ -81,10 +81,10 @@ public:
     ////////
 
     //! The index of the water saturation
-    static const int waterSaturationIdx = PVOffset + 0;
```
andlaus: is there a reason to change indices? While the eWoms code should be agnostic of this, I'm not so sure about the stuff in `opm-simulators`.

ewoms/models/blackoil/blackoilintensivequantities.hh
```diff
@@ -107,14 +107,14 @@ public:
     /*!
      * \copydoc IntensiveQuantities::update
      */
-    void update(const ElementContext& elemCtx, unsigned dofIdx, unsigned timeIdx)
+    void update(const ElementContext& elemCtx, unsigned dofIdx, unsigned timeIdx, unsigned focustimeidx)
     {
-        ParentType::update(elemCtx, dofIdx, timeIdx);
+        ParentType::update(elemCtx, dofIdx, timeIdx);// this only gives extrusion factor
```
andlaus: the comment is probably correct in the particular case you were looking at, but it could be misleading in the general one.

ewoms/models/blackoil/blackoilintensivequantities.hh
```diff
@@ -107,14 +107,14 @@ public:
     /*!
      * \copydoc IntensiveQuantities::update
      */
-    void update(const ElementContext& elemCtx, unsigned dofIdx, unsigned timeIdx)
+    void update(const ElementContext& elemCtx, unsigned dofIdx, unsigned timeIdx, unsigned focustimeidx)
```
andlaus: as usual: parameter naming convention. also: this information is already provided by the element context object.

ewoms/models/blackoil/blackoilintensivequantities.hh
```diff
@@ -327,12 +327,13 @@ public:
             porosity_ *= 1.0 + x + 0.5*x*x;
         }
 
-        asImp_().solventPvtUpdate_(elemCtx, dofIdx, timeIdx);
-        asImp_().polymerPropertiesUpdate_(elemCtx, dofIdx, timeIdx);
+        asImp_().solventPvtUpdate_(elemCtx, dofIdx, timeIdx, focustimeidx);// strictly do not need focus time since no makeEvaluation is done
```
andlaus: that comment is a bit too terse for my taste...

(and as usual, the additional parameter is redundant.)

ewoms/models/blackoil/blackoilintensivequantities.hh
```diff
@@ -327,12 +327,13 @@ public:
...
+        asImp_().polymerPropertiesUpdate_(elemCtx, dofIdx, timeIdx, focustimeidx);
         asImp_().updateEnergyQuantities_(elemCtx, dofIdx, timeIdx, paramCache);
 
         // update the quantities which are required by the chosen
         // velocity model
+        // TODO: use the focus time here
```
andlaus: it is already provided by the element context object. remove comment?

ewoms/models/blackoil/blackoilmodel.hh
```diff
@@ -503,6 +503,7 @@ public:
             }
         }
 
+        // this is restart spesific behavoir and to avoid this one have to hack outside for adjiont
```
andlaus: several typos. I also did not really get what you want to say with this.

ewoms/models/blackoil/blackoilpolymermodules.hh
```diff
@@ -861,10 +861,11 @@ public:
      */
     void polymerPropertiesUpdate_(const ElementContext& elemCtx,
                                   unsigned dofIdx,
-                                  unsigned timeIdx)
+                                  unsigned timeIdx,
+                                  unsigned focustime)
```
andlaus: the familiar comment: parameter naming convention, parameter is redundant. in addition here: parameter name is inconsistent.

ewoms/models/blackoil/blackoilpolymermodules.hh
```diff
@@ -952,7 +953,9 @@ class BlackOilPolymerIntensiveQuantities<TypeTag, false>
 public:
     void polymerPropertiesUpdate_(const ElementContext& elemCtx OPM_UNUSED,
                                   unsigned scvIdx OPM_UNUSED,
-                                  unsigned timeIdx OPM_UNUSED)
+                                  unsigned timeIdx OPM_UNUSED,
+                                  unsigned focustimeIdx OPM_UNUSED)
```
andlaus: parameter naming convention. (I'll skip this from here on.)

ewoms/models/blackoil/blackoiltwophaseindices.hh
```diff
@@ -76,10 +76,10 @@ public:
     //////////////////////////////
 
     //! The index of the water saturation. For two-phase oil gas models this is disabled.
-    static const int waterSaturationIdx  = waterEnabled ? PVOffset + 0 : -10000;
+    static const int waterSaturationIdx  = waterEnabled ? PVOffset + 1 : -10000;
```
andlaus: as for the threephase case: these PV-index changes could be potentially disruptive.

ebos/eclproblem.hh
atgeirr: > capital `n` for property names please: properties should follow the same naming convention as classes.

OK.



> also: is the adjoint functionality limited to the EclProblem or does it also affect the blackoil model? if the latter, the property definition should move to `blackoilproperties.hh` and setting the default value should go to `blackoilmodel.hh`

The property is only used from opm-simulators/Flow, so it should probably remain in the EclProblem.

ewoms/common/simulator.hh
atgeirr: This function also deserializes the model and problem, which is not done by deserialize() (why?).

ewoms/disc/common/fvbaseelementcontext.hh
```diff
@@ -583,7 +588,7 @@ protected:
 #endif
 
         dofVars_[dofIdx].priVars[timeIdx] = priVars;
-        dofVars_[dofIdx].intensiveQuantities[timeIdx].update(/*context=*/asImp_(), dofIdx, timeIdx);
+        dofVars_[dofIdx].intensiveQuantities[timeIdx].update(/*context=*/asImp_(), dofIdx, timeIdx, focusTimeIdx_);
```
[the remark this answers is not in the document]
atgeirr: I think we agree, we'll try to change this.

ewoms/disc/ecfv/ecfvdiscretization.hh
atgeirr: At the end of a step solution(0) and solution(1) are different, so at that point the code is wrong. The deserialize() method would probably be better if it wrote solution at both timeindices?

ebos/eclproblem.hh
andlaus: > The property is only used from opm-simulators/Flow, so it should probably remain in the EclProblem.

ok, everything is as it is supposed to be then. But is it actually used somewhere within eWoms? I can't remember seeing it.

ewoms/common/simulator.hh
andlaus: so far, the simulator is responsible for the time loop and by extension episode management in eWoms (it actually started out as a class called `TimeManager`). since `flow` bolts its own mechanisms on top, this change will help `flow` but it will break everything else. Note that this impedance mismatch has already been a major headache in the past.

ewoms/models/blackoil/blackoilindices.hh
atgeirr: There are two reasons:
 - having pressure as the first variable is most convenient for CPR
 - having pressure first is consistent with the two-phase models

ewoms/common/simulator.hh
andlaus: that was my point: if it is not done by `deserialize()`, then there's something wrong with `deserialize()` and it should be added to it. Possibly, the models and problems which this was tested for did not need it, so the bug was simply not discovered yet.

ewoms/models/blackoil/blackoiltwophaseindices.hh
atgeirr: The idea is to always have the pressure (which always is used) in the same spot, with the old code it varies between the (water, oil) and (oil, gas) cases.

ewoms/disc/ecfv/ecfvdiscretization.hh
andlaus: yes, I don't object to writing both. in the usage scenarios up to now the two solutions are just supposed to be identical. (for this, the restart file must maybe be written at the beginning of a timestep instead of at the end.)

atgeirr:
We'll go through the rest of the comments and propose a revision. There are a few somewhat fundamental questions that need clarification, long-term: what is the simulator class' responsibility (re: the timestep thing), and what is serialization supposed to do (re: the solution at timeindex 0 and 1 thing).

ewoms/models/blackoil/blackoilindices.hh
andlaus: > having pressure as the first variable is most convenient for CPR

Ok. I cannot really see why, though.

> having pressure first is consistent with the two-phase models.

yes, but does this matter? I'm a bit worried that the well model implicitly assumes an ordering of the primary variables.

andlaus:
thanks for the quick response. To answer your question: at the moment the Simulator basically the "global" object which holds everything and the time stepper. time step size control is delegated to the problem, though. (cf. `FvBaseProblem::nextTimeStepSize()`)

ewoms/common/simulator.hh
hnil: At the moment I disagree, but if you find a resonable solution it is fine. The current version at lest is very dependent on the forward state machin consept.

ewoms/common/simulator.hh
hnil: added If deserializing for adjoint calling reloading the current episode is not needed here. All relevant information should have been serilaized in the forward run.

ewoms/disc/common/fvbaseelementcontext.hh
hnil: done!

ewoms/disc/common/fvbaseelementcontext.hh
hnil: done!

ewoms/disc/common/fvbaseelementcontext.hh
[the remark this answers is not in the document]
hnil: Where should the knowledge of focusDofIdx and focusTimeIdx live. fvbaseelment, intensivequantities ??

ewoms/disc/common/fvbaselinearizer.hh
hnil: Ok. I would have preferred it to be obligatory, but ...
In general where is the responsibility of known what one linearize for?

ewoms/disc/common/fvbaselocalresidual.hh
hnil: Using tmp2Der for now. Removing tmp everywere with storage or some more descriptive would probably be better

ewoms/disc/common/fvbaselocalresidual.hh
hnil: as far as i remember tmp2 is calculated directly from the intensive quantites from prevois step. This have derivatives calculated with respect to the values with respect to something prevois step. However if one update everything this would not be needed i.e. all timeIdx.

hnil: What is POD time?

ewoms/disc/common/fvbaseprimaryvariables.hh
hnil: done

ewoms/disc/ecfv/ecfvdiscretization.hh
hnil: (de)serialization is preferably to write the state of an object and then restablish it. This is a bit different than restart.  The comment is to say that if you serialize and then deserialize again we are at at different state.

ewoms/io/restart.hh
hnil: It is not nesseary but else one easy make a mess of files. If you have a nice robust change it is fine for me, but i think boost::filesystem is best until it is in the standard..

ewoms/models/blackoil/blackoilindices.hh
hnil: It is convenient to have the pressure at the same index independent of models.

ewoms/models/blackoil/blackoilintensivequantities.hh
hnil: removed!

ewoms/models/blackoil/blackoilintensivequantities.hh
hnil: done!

ewoms/models/blackoil/blackoilintensivequantities.hh
hnil: removed comment

ewoms/models/blackoil/blackoilintensivequantities.hh
hnil: Let it stand due to the trouble with use of how intensive quantities are updated, old timeIdx is some time only shifted in place instead of updated true the updateIntensive quanties.

ewoms/disc/common/fvbaseelementcontext.hh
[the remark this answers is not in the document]
andlaus: neither: it should live in in the element context

ewoms/disc/common/fvbaselinearizer.hh
andlaus: > In general where is the responsibility of known what one linearize for?

that's the discretization's job. in forward mode we only need to linearize for the most recent solution, so I'd like to keep this the default.

ewoms/models/blackoil/blackoilmodel.hh
andlaus: > The point is anyway that serializing and restart is to different things

Yes, and no: to me restart = serialization -> write to disk -> read from disk -> deserialization. i.e., to me (de)serialization is proper subset of the operations required for restart. *partial* serialization is a different matter...

hnil:
This should be replaced by #447

PR #376, opened 2018-08-30
Fix eclipse output for reordered CpGrid and add command line parameter

blattms:
Companion award of OPM/opm-grid#340
The eclipse output expects the arrays in CellData to be ordered correctly, that
is local indices should be such that cartesian index i is running fastest, then
j and then k. In case the CpGrid is reordered that does not hold. Therefore we reorder
the array just before calling the eclipse IO methods.

In addition we added a command line parameter --enable-reorder-k-fastest to switch on
reordering. Default is false.

### DIFF
--- a/ebos/eclbasevanguard.hh
+++ b/ebos/eclbasevanguard.hh
@@ -68,6 +68,8 @@ NEW_PROP_TAG(EclOutputInterval);
 SET_STRING_PROP(EclBaseVanguard, EclDeckFileName, "");
 SET_INT_PROP(EclBaseVanguard, EclOutputInterval, -1); // use the deck-provided value
 
+NEW_PROP_TAG(EnableReorderKFastest);
+SET_BOOL_PROP(EclBaseVanguard, EnableReorderKFastest, false);
 END_PROPERTIES
 
 namespace Ewoms {
@@ -102,6 +104,8 @@ public:
                              "The name of the file which contains the ECL deck to be simulated");
         EWOMS_REGISTER_PARAM(TypeTag, int, EclOutputInterval,
                              "The number of report steps that ought to be skipped between two writes of ECL results");
+        EWOMS_REGISTER_PARAM(TypeTag, bool, EnableReorderKFastest,
+                             "If true then k will be the fastest running index of ijk for CpGrid");
     }
 
     /*!
--- a/ebos/eclwriter.hh
+++ b/ebos/eclwriter.hh
@@ -222,7 +222,8 @@ public:
                                                                      miscSummaryData,
                                                                      regionData,
                                                                      blockData,
-                                                                     enableDoublePrecisionOutput);
+                                                                     enableDoublePrecisionOutput,
+                                                                     globalGrid_.eclipseOuputIndexLookup());
 
             // then, make sure that the previous I/O request has been completed and the
             // number of incomplete tasklets does not increase between time steps
@@ -305,7 +306,8 @@ public:
                                                                      miscSummaryData,
                                                                      regionData,
                                                                      blockData,
-                                                                     enableDoublePrecisionOutput);
+                                                                     enableDoublePrecisionOutput,
+                                                                     globalGrid_.eclipseOuputIndexLookup());
 
             // then, make sure that the previous I/O request has been completed and the
             // number of incomplete tasklets does not increase between time steps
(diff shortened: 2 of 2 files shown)


blattms:
As the default has not changed, jenkins will not be able to test this PR.

ebos/eclbasevanguard.hh
```diff
@@ -102,6 +104,8 @@ public:
                              "The name of the file which contains the ECL deck to be simulated");
         EWOMS_REGISTER_PARAM(TypeTag, int, EclOutputInterval,
                              "The number of report steps that ought to be skipped between two writes of ECL results");
+        EWOMS_REGISTER_PARAM(TypeTag, bool, EnableReorderKFastest,
+                             "If true then k will be the fastest running index of ijk for CpGrid");
```
andlaus: I think you don't should not use a parameter for this because if the user gets this wrong output will break? instead, ask the grid...

ebos/eclwriter.hh
```diff
@@ -515,7 +517,17 @@ private:
...
-
+        /// \brief Indices for Reordering cell data for Eclipse IO.
+        ///
+        /// If not empty, a vector mapping the local Cpgrid index
+        /// (with index k running fastest) to the one needed by
+        /// Eclipse IO (i fastest, then j, and the k)
+        std::vector<int> eclipseReorderingIndices_;
+
+        ///
+        /// \param eclipseReorderingIndices If not empty, a vector mapping the local Cpgrid index
```
andlaus: only documenting a single parameter but nothing else here seems a bit weird. Since this is a non-public class, I am fine if you remove the documentation, but you may also leave it as-is.

ebos/eclwriter.hh
```diff
@@ -534,11 +547,31 @@ private:
...
         // callback to eclIO serial writeTimeStep method
         void run()
         {
+            if ( eclipseReorderingIndices_.size() )
+            {
+                // reorder each entry in the CellData to get the ordering expected
+                // by eclipse IO
+                for(auto&& entry: restartValue_.solution)
+                {
+                    decltype(entry.second.data) tmp(eclipseReorderingIndices_.size());
```
andlaus: can't `auto` be used here? copying the array does not matter too much performance-wise...

andlaus:
This just dropped into my mind: It is probably better to do the reordering where the ECL data is extracted, i.e. here:

https://github.com/OPM/ewoms/blob/master/ebos/ecloutputblackoilmodule.hh

this would avoid the hassle of having to copy and resort the data arrays. Also, the name of this functionality should probably be a bit more specific (e.g. "ECL element order"?) because "reordering" could mean anything and it is conceivable that other output formats require a different element order.

blattms:
> This just dropped into my mind: It is probably better to do the reordering where the ECL data is extracted, i.e. here:
>
> https://github.com/OPM/ewoms/blob/master/ebos/ecloutputblackoilmodule.hh

Well, that was my first thought, too. Unfortunately, that would not work in parallel. The reordering can only (easily) take place on the global grid after the data is collected.

blattms:
With the reordering now actually enabled when requested, I see deviations in the wells for e.g. Model1.

Not sure the ordering effects wells. Up to now I thought that the wells are using cartesian indices when they are ouput by ecl. Seems like I am wrong.

andlaus:
> Well, that was my first thought, too. Unfortunately, that would not work in parallel.

right. I forgot about parallel as usual.

> The reordering can only (easily) take place on the global grid after the data is collected.

I thought the `CollectToIORank` class does things like this? if this is more complicated to implement than the current solution, I'm okay with the PR in its current form, though.

> With the reordering now actually enabled when requested, I see deviations in the wells for e.g. Model1.

some numerical noise is to be expected, in particular because the linear solver is supposed to behave slightly differently?

ebos/eclwriter.hh
andlaus: BTW: parameters are not usually documented here because (a) eWoms strives to use self-evident parameter names and (b) IME the code churn for parameters is usually much higher than for semantics and adapting the parameter documentation is very frequently forgotten. Wrong documentation is IMHO much worse than no documetation.

blattms:
I am not talking about numerical noise.

andlaus:
this needs PR some manual conflict resolution now. I'm pretty sure it is a simple affair, though.

but I'm also curious: are there any news on this?

andlaus:
is there any intention of getting this and the opm-grid PR merge-ready soon? if no, I'll close this PR until there is an effort to finish it.

atgeirr:
Please give @blattms a little more time (until early next week) to respond before closing.

PR #377, opened 2018-08-31
Threading fixes

andlaus:
this fixes a race condition in the VTK output code if intrinsic permeabilitiy tensors ought to be written. the reason was that Dune::DynamicMatrix::resize() is not threadsafe. now, we simply rely on the fact that the `Tensor` objects are assumed to be dim \times dim matrices by the base class anyway.

Further this PR reverts most of #370 and #371 because they break builds without OpenMP enabled and the threading mumbo-jumbo for flow is better done within opm-simulators.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -166,10 +166,10 @@ opm_add_test(waterair_pvs_ni
              TEST_ARGS --grid-global-refinements=1)
 
 opm_add_test(lens_immiscible_vcfv_ad
-             TEST_ARGS --end-time=3000 --threads-per-process=1)
+             TEST_ARGS --end-time=3000)
 
 opm_add_test(lens_immiscible_vcfv_fd
-             TEST_ARGS --end-time=3000 --threads-per-process=1)
+             TEST_ARGS --end-time=3000)
 
 opm_add_test(lens_immiscible_ecfv_ad
              TEST_ARGS --end-time=3000)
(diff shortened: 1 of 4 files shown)


andlaus:
@akva2, @atgeirr: merge this and OPM/opm-simulators#1562?

andlaus:

andlaus:
to be clear: this PR does not change the results. I just want to hijack it to do the data update that was missing in #381.

andlaus:
@atgeirr: can this please be merged? there still is a performance regression in master compared to August 10 and I'm pretty sure that this PR fixes it in conjunction with OPM/opm-simulators#1562. (isn't this supposed to be one of your lines, @alfbr?)

atgeirr:
> @atgeirr: can this please be merged?

Probably yes, but I considered this more of @akva2's domain of expertise? Especially with all the back-and-forth on this I am not quite certain what is the best solution.

alfbr:
Sorry, I have not kept up on this one. Fixing performance regressions is always welcome of course, but I have no idea what we are dealing with here. I am more than happy if @akva2 has a look.

akva2:
seems fine to me, it moves the flow setup things to flow.

andlaus:
> I have no idea what we are dealing with here.

if you look at the curves of linuxbenchmarking.com, you'll notice that the current results for most decks are still worse than those of August 10 (but mostly better than August 25). I bet a beer that this is because each process uses numCores threads  if you  don't specify anything (and it seems like the linuxbenchmarking guy does not do anything). for e.g. "flow MPI extra with 8 threads" (which is a misnomer BTW, because this curve is for 8 *processes*), it ought to spawn 8*8=64 or 8*16=128 threads depending on the processor.

this two PRs restore the threading behaviour of August 10, i.e., 2 threads per process by default.

andlaus:
weird. @akva2: any idea why don't get a data update PR for opm-tests?

akva2:
because you commented before it had triggered. the pr builder see the last comment posted (it's a bug..).

akva2:

alfbr:
>  I bet a beer that this is because each process uses numCores threads if you don't specify anything

I'll take that beer on Wednesday :-) The performance hit is because of the tightened mass-balance tolerance (from 10^-5 to 10^-6).

andlaus:
> I'll take that beer on Wednesday :-) The performance hit is because of the tightened mass-balance tolerance (from 10^-5 to 10^-6).

maybe we need to buy each other a glass: the tightened tolerances don't really explain why the performance for tests with many processes is actually worse than before august 28.

joakim-hove:
jenkins build this opm-tests=57 please

andlaus:
jenkins build this opm-tests=57 please

andlaus:
> I'll buy you a beer anyway ;-)

let's see if the curves go down tomorrow, but your point seems to be solid, so I concede. no need for you to pay ;)

andlaus:
you earned yourself a beer: the curves did not change :).

PR #378, opened 2018-09-03
Change string used for THPRES in restartfile

joakim-hove:
Followup to: https://github.com/OPM/opm-common/pull/479


andlaus:
sorry to be a jerk about this, but you were waay too short on details: what was the reason for this change? (for the record: I'm okay with changing such "magic cookie" strings.)

PR #381, opened 2018-09-07
Return modified ntg's only when minpv mode is OPMFIL

totto82:


akva2:
this broke opm-simulators, e.g. https://ci.opm-project.org/job/opm-common-PR-builder/443/

alfbr:
That is to be expected, the reference solution needs to be updated.

andlaus:

andlaus:
oops, sorry.

PR #383, opened 2018-09-11
reset the TTY after the simulator was interrupted more reliably

andlaus:
this should help prevent the terminal from being deranged after the simulator has been aborted by user intervention in more cases.

Also, we now reset the TTY if an exception was received on the top-level.

### DIFF
--- a/ewoms/common/start.hh
+++ b/ewoms/common/start.hh
@@ -178,6 +178,32 @@ static inline int setupParameters_(int argc, const char **argv, bool registerPar
 }
 
 /*!
+ * \brief Resets the current TTY to a usable state if the program was aborted.
+ *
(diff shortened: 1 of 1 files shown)


andlaus:
Note: this does not have any effect on `flow` at the moment, because `flow` uses its own startup routines and time loop.

andlaus:
jenkins build this please

andlaus:
jenkins build this please

andlaus:
since nobody made any comments here during the last week, jenkins is happy, the changes are kind of simple, it is annoying to be stuck with a locked-up terminal, and flow is unaffected by this, I will self merge this. if it causes problems, I'll revert...

PR #384, opened 2018-09-12
Add --opm-rst-file parameter

joakim-hove:
See: https://github.com/OPM/opm-common/pull/493

This is a policy decision - please voice your opinion.

### DIFF
--- a/ebos/eclbasevanguard.hh
+++ b/ebos/eclbasevanguard.hh
@@ -63,10 +63,12 @@ NEW_PROP_TAG(EquilGrid);
 NEW_PROP_TAG(Scalar);
 NEW_PROP_TAG(EclDeckFileName);
 NEW_PROP_TAG(OutputDir);
+NEW_PROP_TAG(EnableOpmRstFile);
 NEW_PROP_TAG(EclOutputInterval);
 
 SET_STRING_PROP(EclBaseVanguard, EclDeckFileName, "");
 SET_INT_PROP(EclBaseVanguard, EclOutputInterval, -1); // use the deck-provided value
+SET_BOOL_PROP(EclBaseVanguard, EnableOpmRstFile, false);
 
 END_PROPERTIES
 
@@ -388,6 +390,9 @@ private:
         // specify the directory output. This is not a very nice mechanism because
         // the eclState is supposed to be immutable here, IMO.
         ioConfig.setOutputDir(outputDir);
+
+        bool opmRSTFile = EWOMS_GET_PARAM(TypeTag, bool, EnableOpmRstFile);
+        ioConfig.setEclCompatibleRST( !opmRSTFile );
     }
 
     Implementation& asImp_()
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -257,6 +257,9 @@ SET_BOOL_PROP(EclBaseProblem, EclOutputDoublePrecision, false);
 // The default location for the ECL output files
 SET_STRING_PROP(EclBaseProblem, OutputDir, ".");
 
+// Should we output OPM or ECL restart files
+SET_BOOL_PROP(EclBaseProblem, EnableOpmRstFile, false);
+
 // the cache for intensive quantities can be used for ECL problems and also yields a
 // decent speedup...
 SET_BOOL_PROP(EclBaseProblem, EnableIntensiveQuantityCache, true);
@@ -390,6 +393,8 @@ public:
                              "Tell the output writer to use double precision. Useful for 'perfect' restarts");
         EWOMS_REGISTER_PARAM(TypeTag, unsigned, RestartWritingInterval,
                              "The frequencies of which time steps are serialized to disk");
+        EWOMS_REGISTER_PARAM(TypeTag, bool, EnableOpmRstFile,
+                             "Should we include special OPM keywords to enable flow based restart");
     }
 
     /*!


joakim-hove:
@andlaus : This compiles - but the tests fail hard:
```
2: Test command: /private/joaho/work/OPM/ewoms/bin/runtest.sh "--plain" "test_equil"
2: Test timeout computed to be: 1500
2: ######################
2: # Running test 'test_equil'
2: ######################
2: executing "./bin/test_equil "
2: Instantiating the vanguard
2: Reading the deck file 'data/equil_base.DATA'
2: terminate called after throwing an instance of 'std::runtime_error'
2:   what():  Accessing parameter OpmRSTFile without prior registration is not allowed.
2: /private/joaho/work/OPM/ewoms/bin/runtest.sh: line 97:  6300 Aborted                 (core dumped) "$TEST_BINARY" $TEST_ARGS
1/1 Test #2: equil ............................***Failed    0.28 sec
```
some hints on how to proceed?

alfbr:
Depends on who the fool is I guess. Right now we do not have prediction in a working state, so in a typical usage scenario (for users, not developers) a restart will be more useful with Eclipse. Please note that this is intended as a stop-gap short-term solution. We expect to have both flow and Eclipse able to restart from the same files within weeks, then the switch will be removed again.

andlaus:
I'm sorry for not really getting what you mean: Is it better to write ECLIPSE compatible RST-files by default or not? All I wanted to say here is that if there is no significant harm in doing so, it is better to default to ECLIPSE compatible. That said, maybe there are some drawbacks to this; if so, what are they? (this is honest question, I do not have a clue here.)

ewoms/disc/common/fvbaseproperties.hh
```diff
@@ -185,7 +185,7 @@ NEW_PROP_TAG(EnableGridAdaptation);
  * \brief The directory to which simulation output ought to be written to.
  */
 NEW_PROP_TAG(OutputDir);
-
+NEW_PROP_TAG(OpmRSTFile);
```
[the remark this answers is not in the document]
andlaus: the name of this property is also quite confusing: I would assume that is a string with the name of the Opm restart file, but from the code it looks like it specifies if the restart files ought to be written in an ECLIPSE compatible way. Rename to `EnableEclipseCompatibleRestart`? or, if this is always a good idea, remove the parameter and set it unconditionally in the `EclipseState` object.

andlaus:
ok, I guess I made a sign-error in understanding the parameter. (it *disables* ECLIPSE compatible restart files instead of *enabling* them.)

ewoms/disc/common/fvbaseproperties.hh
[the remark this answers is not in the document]
andlaus: also, this property does not belong to the basic discretization infrastructure. move it to ebos, i.e., to `eclbasevanguard.hh`.

joakim-hove:
> (it disables ECLIPSE compatible restart files instead of enabling them.)

That is right

joakim-hove:
> you did not register the parameter before usage

Thank you!

ebos/eclproblem.hh
```diff
@@ -257,6 +257,9 @@ SET_BOOL_PROP(EclBaseProblem, EclOutputDoublePrecision, false);
 // The default location for the ECL output files
 SET_STRING_PROP(EclBaseProblem, OutputDir, ".");
 
+// Should we output OPM or ECL restart files
+SET_BOOL_PROP(EclBaseProblem, OpmRstFile, false);
```
andlaus: while this does no harm, it is not required because the property is already set for the `EclBaseVanguard` type tag (which is a parent of `EclBaseProblem`). if you like to leave it as a reminder, this is okay with me.

ebos/eclproblem.hh
```diff
@@ -390,6 +393,8 @@ public:
                              "Tell the output writer to use double precision. Useful for 'perfect' restarts");
         EWOMS_REGISTER_PARAM(TypeTag, unsigned, RestartWritingInterval,
                              "The frequencies of which time steps are serialized to disk");
+        EWOMS_REGISTER_PARAM(TypeTag, bool, OpmRstFile,
+                             "Should we include special OPM keywords to enable flow based restart");
```
andlaus: please move this to the the place where the parameter is used, i.e., to the `registerParameter()` function of `EclBaseVanguard`.

ebos/eclbasevanguard.hh
```diff
@@ -63,10 +63,12 @@ NEW_PROP_TAG(EquilGrid);
 NEW_PROP_TAG(Scalar);
 NEW_PROP_TAG(EclDeckFileName);
 NEW_PROP_TAG(OutputDir);
+NEW_PROP_TAG(OpmRstFile);
```
andlaus: please prefix with `Enable` (this is the eWoms standard convention for boolean parameters)

joakim-hove:
OK - parameter renamed to `EnableOpmRstFile`. Whether this should be merged at all I guess is up to @alfbr / @bska - I agree it is not particularly nice, and should eventually be removed again.

andlaus:
the parameter registration is still done in the wrong file (it logically belongs to the vanguard). the PR can be merged as-is, though. (In this case I'll open a small cleanup one which does the move.)

joakim-hove:
Ok - thanks, the best would be if you point to correct location, so I can test a correct binary locally.

andlaus:
simply cut-and-paste the `EWOMS_REGISTER_PARAMETER()` call to `eclbasevanguard.hh`...

andlaus:
(btw: your current solution is equally correct from the user's POV. for us developers it is IMO just confusing to do the parameter  registration within a class where it is not used at all.)

joakim-hove:
Thank you!

joakim-hove:
OK - it was not immediately clear to me where the fix should be; it works for me locally - so I am grateful if you make the final touch when/if this is merged.

andlaus:
agreed. as said: from a user's point of few there is nothing wrong with the PR in its present form...

PR #386, opened 2018-09-17
move the registration of the EnableOpmRstFile to the EclBaseVanguard

andlaus:
before this patch the parameter was registered by the problem but not used there. Since this is quite confusing, let's move registration to where the parameter is actually used, i.e., the vanguard.

@joakim-hove: please press Green if you do not object to this.


joakim-hove:
Thank you - merging.

PR #388, opened 2018-09-21
Handle more specific exceptions first

rolk:
The handler for DuneException will be hidden behind the std::exception
and never be invoked. It should be moved up so that the most specific
exception is handled first.

### DIFF
--- a/ewoms/common/start.hh
+++ b/ewoms/common/start.hh
@@ -347,16 +347,6 @@ static inline int start(int argc, char **argv)
         }
         return 0;
     }
-    catch (std::exception& e)
-    {
-        if (myRank == 0)
-            std::cout << e.what() << ". Abort!\n" << std::flush;
-
-        std::cout << "Trying to reset TTY.\n";
-        resetTerminal_();
-
-        return 1;
-    }
 #if ! DUNE_VERSION_NEWER(DUNE_COMMON, 2,5)
     catch (Dune::Exception& e)
     {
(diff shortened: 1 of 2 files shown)


andlaus:
No objections, I'll merge if jenkins is happy. That said, I thought that Dune < 2.5 does not derive its exception classes from `std::exception` so the impact of this is probably limited.

andlaus:
jenkins build this please

rolk:
@andlaus But isn't the test the opposite way, that the #ifdef kicks in if the version of DUNE is 2.5 or later. (At least I think that I have built DUNE 2.5, and gets the warning)

andlaus:
I think you did not see the exclamation mark before the `DUNE_VERSION_NEWER(2,5)`: these handlers are only enabled for 2.4 ;)

anyway jenkins is happy and it does not hurt to do it like this, so let's merge. thanks for the patch.

rolk:
@andlaus For what it's worth; if I do `grep "DUNE_COMMON_VERSION_" config.h` in the eWoms build directory, or in my module's build directory, I get this output:

    #define DUNE_COMMON_VERSION_MAJOR 2
    #define DUNE_COMMON_VERSION_MINOR 5
    #define DUNE_COMMON_VERSION_REVISION 3

and yet I got the warning.

rolk:
@andlaus Please disregard all the spam: I managed to trace down the source of all the woes; config.h was not included as the very first header. (Illustrates the perils of relying on HAVE macros outside of the DUNE headers in my opinion, but that's what we got!)

andlaus:
no worries: as said, catching the dune exceptions first is probably a good idea anyway.

(note that having to synchronize config.h with the dune build system is part of the "impedance mismatch" between the two build systems. Discussing whether this pain is worth the effort has already been done ad-nauseam, so let's skip this here.)

PR #389, opened 2018-09-24
Set default value of EnableOpmRstFile to true

joakim-hove:
As discussed on email the default will be to produce restart files which can be used in a flow -> flow restart.

Equinor will internally use a patched version:


andlaus:
you probably need a data update; let's see.

joakim-hove:
> you probably need a data update; let's see.

A :beer: on no ...

andlaus:
no deal ;). I'm not familiar enough with what this flag does.

PR #390, opened 2018-09-24
[Equinor internal]: Opm rst default false

joakim-hove:
**This should not be merged, but will be used internally in *Equinor*.**

The net effect of this pr is to add the option: `--enable-opm-rst-file=false` to the `flow` command line.

### DIFF
--- a/ebos/eclbasevanguard.hh
+++ b/ebos/eclbasevanguard.hh
@@ -72,7 +72,7 @@ NEW_PROP_TAG(IgnoreKeywords);
 SET_STRING_PROP(EclBaseVanguard, IgnoreKeywords, "");
 SET_STRING_PROP(EclBaseVanguard, EclDeckFileName, "");
 SET_INT_PROP(EclBaseVanguard, EclOutputInterval, -1); // use the deck-provided value
-SET_BOOL_PROP(EclBaseVanguard, EnableOpmRstFile, true);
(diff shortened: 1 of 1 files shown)


andlaus:
can this be closed? a branch on @joakim-hove's fork serves the same purpose and does not clutter the PR list for master: As far as I can see, the only reason to open a PR without the intend of merging it is to be able to give it the Jenkins treatment, but jenkins will fail here anyway.

atgeirr:
> if no objections are voiced, I'll close this in the afternoon or tomorrow.

Actually, this PR should be merged once we have full compatibility of restart files. If my understanding is correct this should be retained (and lose the [equinor internal] label soon).

andlaus:
ok, can we get rid of that knob completely then?

andlaus:
(I mean the `EnableOpmRstFile` parameter. There is little point to it once this is merged.)

andlaus:
what's the status on getting restart from plain ECL  files working?

andlaus:
jenkins build this please

andlaus:
this is superseeded by #476? closing.

totto82:

andlaus:
does this work as intended? I would expect that the reference data would change...

alfbr:
This has been tested and confirmed to work as intended on our end.

andlaus:
ok, I take your word for it.

PR #400, opened 2018-10-15
Support modifying trans from deck

totto82:
Depends on OPM/opm-common#462

This is an alternative to OPM/opm-common#522 and friends. 

Tested on simple case.


totto82:
jenkins build this opm-common=462 please

joakim-hove:
OK - this has been tested internally in Equinor. Merging

PR #401, opened 2018-10-15
Add support for region DR[SV]DTR

totto82:
Depends on OPM/opm-common#526


joakim-hove:
I have added some comments regarding this PR here: https://github.com/OPM/opm-common/pull/526#issuecomment-430032439

totto82:
jenkins build this opm-common=526 please

PR #403, opened 2018-10-17
Removed porosity from energy of rock

hnil:
Suggestion on remove the porosity from storage term in rock, else we will generate temperature since the rock density is not assumed to depend on porositymultipliers.

### DIFF
--- a/ewoms/models/blackoil/blackoilenergymodules.hh
+++ b/ewoms/models/blackoil/blackoilenergymodules.hh
@@ -168,7 +168,8 @@ public:
 
         // add the internal energy of the rock
         const auto& uRock = Opm::decay<LhsEval>(intQuants.rockInternalEnergy());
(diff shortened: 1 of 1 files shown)


andlaus:
are you sure that the values specified by SPECROCK already include porosity? I tried to look this up in the documentation but could not find anything explicit. the fact that SPECROCK only provides one value per saturation region seems to indicate that the values are for massive rock, though?

but maybe I misunderstood: if your point is that porosity multipliers ought not be be used here (seems reasonable), then the reference porosity should be used instead?

hnil:
I am not sure. But atleast it should not be poro since this is time dependent and with this definition we will not conserve mass of rock. This give temperature changes only due to porosity changes since the total mass of rock in this sence will change in the cell.

andlaus:
this is succeeded by #409. closing.

PR #404, opened 2018-10-18
Blackoil: Prepare for 2p conditional storage of fluidstate

totto82:
Depends on OPM/opm-material#310

Gives a small (5%) speed up in the linearizer for the Olympus case.


totto82:
jenkins build this opm-material=310 please

andlaus:
the increased complexity seems to be a worthwhile trade-off for the performance improvement. I'll merge this and the opm-material once the `flow` tests succeed on jenkins...

andlaus:
I think this is okay, but jenkins still has some -- probably minor -- complaints.

PR #405, opened 2018-10-23
ebos: add an API for aquifiers

andlaus:
this is similar to the mechanism for well models: the API is defined here, but can be overloaded by downstream modules. In contrast to the well model where the default class follows a simplistic approach, the default class for aquifiers does nothing, i.e., the actual implemenentation must be provided by downstream.


andlaus:
jenkins build this please

andlaus:
since jenkins is happy and this does not affect `flow` at present, I'll be frivolous enough to self-merge this tomorrow if no objections are raised.

andlaus:
seems like nobody objects. I'll merge because this blocks a aquifer cleanup PR for opm-simulators.

PR #407, opened 2018-10-29
blackoil model: fix the equation weights

andlaus:
The main change is that we now care quite a bit less about water than about the hydrocarbons. Also, the equation weight was inverted when conserving fluid surface volume instead of mass, oops. Note that the tolerances for oil, gas and water are now always expressed in terms of "invalid kg per second", i.e., the tolerances for the Newton-Raphson solver probably should be changed for problems that used the black-oil model formulated in terms of surface volumes.

this is only relevant for the eWoms Newton-Raphson solver, i.e. `flow` is completely unaffected at the moment.

### DIFF
--- a/ewoms/models/blackoil/blackoilmodel.hh
+++ b/ewoms/models/blackoil/blackoilmodel.hh
@@ -376,16 +376,22 @@ public:
         if (globalDofIdx >= this->numGridDof())
             return 1.0;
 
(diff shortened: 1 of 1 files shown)


andlaus:
jenkins build this please

totto82:
This fixes a bug in the ebos ewoms code (unaffected by flow). I will merge it.

PR #408, opened 2018-10-30
extend the aquifer model so that it can initialize itself

andlaus:
nothing is perfect on first try...


totto82:
This is trivial. I will merge.

PR #409, opened 2018-10-30
Use reference porosity for rock energy storage

andlaus:
this ought to succeed #403. the difference is that porosity needs to be considered for energy storage, but compressibility as it is implemented by the blackoil model must not. as an approximation for reality we thus use the reference porosity instead of the current one.

@hnil: please press green if you agree.


andlaus:
jenkins build this please

andlaus:
the thermal SPE1 test case shows a few very minor deviations as should be expected.

andlaus:

andlaus:
jenkins build this opm-tests=66 please

totto82:
Thanks to @hnil for noticing this bug and for @andlaus. I will merge this later today if no @hnil does not  object.

andlaus:
@totto82: seems like you forgot to press green ;)

PR #410, opened 2018-10-31
Prov. localGeometry meth. for elem.-centr. discret.

rolk:
The SubControlVolume class in vertex-centered discretization has a
localGeometry method that was used in the diffusion module for free
flow boundary conditions, but there was no counterpart to this is the
Dune::Entity class, which is used for element-centered discretization.

Admittedly not knowing whether this is absolutely the same concept,
there is a geometryInFather method in the Dune::Entity class which
provides seemingly similar information.

With this patch, problems using element-centered discretizations can
thus be compiled with diffusion enabled. Testing has been minimal,
however.

### DIFF
--- a/ewoms/disc/ecfv/ecfvstencil.hh
+++ b/ewoms/disc/ecfv/ecfvstencil.hh
@@ -133,6 +133,12 @@ public:
         const LocalGeometry geometry() const
         { return element_.geometry(); }
 
+        /*!
+         * \brief Geometry of the sub-control volume relative to parent.
+         */
+        const LocalGeometry localGeometry() const
+        { return element_.geometryInFather(); }
+
     private:
         GlobalPosition centerPos_;
         Scalar volume_;


rolk:
>  for ECFV localGeometry() should be the same as geometry(), though?

Maybe. My reasoning was this: in vertex-centered discretization [ScvGeometry](https://github.com/OPM/ewoms/blob/release/2018.10/ewoms/disc/vcfv/vcfvstencil.hh#L733) is grid-global and [ScvLocalGeometry](https://github.com/OPM/ewoms/blob/release/2018.10/ewoms/disc/vcfv/vcfvstencil.hh#L115) is something that is has local coordinates; and in Dune::Entity the [geometryInFather](https://gitlab.dune-project.org/core/dune-grid/blob/releases/2.5/dune/grid/common/entity.hh#L489) seems to fit the description. But I can't tell if this is the semantics that is required by the code in the [diffusion module](https://github.com/OPM/ewoms/blob/release/2018.10/ewoms/models/common/diffusionmodule.hh#L439).

rolk:
But what you are saying is that Dune::Entity does not use any local coordinates; everything is in the same coordinate system, so localGeometry() should be the same as geometry() in that case (and using geometryInFather will be detriment, since, if anything, it would return coordinates in a different space than what is expected)?

andlaus:
Ah, okay. The localGeometry() should return something like a reference element if I remember correctly. (it was introduced to specify the subdivision of an element in VCFV, but I'm not sure whether that was formulated in terms of global or local coordinates.) But I see your point and if my interpretation of the dune documentation is correct, I agree. the name `geometryInFather()` is awful then, though.

andlaus:
okay: I'm not completely sure if the solution here is correct, but I take your word that it works for you. merging.

PR #411, opened 2018-11-02
simplify API of Ewoms::GenericGuard

andlaus:
now it is possible to create a guard object like this:

```c++
auto guard = Ewoms::make_guard(callback);
```

where `callback` is any callable object which does not take arguments.


andlaus:
jenkins build this please

totto82:
All test seems fine, I will merge this trivial PR.

PR #412, opened 2018-11-02
parameter system: make it possible to determine if a parameter has been set at runtime

andlaus:
the macro is called `EWOMS_PARAM_IS_SET()` and is currently unused. Nevertheless it might be useful to have in some situations, but use with great care!


andlaus:
jenkins build this please

totto82:
All test seems fine and this seems to be a useful feature. I will merge.

PR #413, opened 2018-11-02
Save/Restore Temperature only in Thermal Runs

bska:
This commit makes the output and reload of temperature data contingent on the simulation run actually enabling thermal mode. This, in turn, increases the compatibility of Flow's restart mechanism with that of the commercial simulator's restart files.

### DIFF
--- a/ebos/ecloutputblackoilmodule.hh
+++ b/ebos/ecloutputblackoilmodule.hh
@@ -229,7 +229,9 @@ public:
         }
         // and oil pressure
         oilPressure_.resize(bufferSize, 0.0);
-        temperature_.resize(bufferSize, 0.0);
+
+        if (simulator_.vanguard().eclState().getSimulationConfig().isThermal())
+            temperature_.resize(bufferSize, 0.0);
 
         if (FluidSystem::enableDissolvedGas())
             rs_.resize(bufferSize, 0.0);
--- a/ebos/eclwriter.hh
+++ b/ebos/eclwriter.hh
@@ -318,13 +318,18 @@ public:
 
     void restartBegin()
     {
+        const auto& es = simulator_.vanguard().eclState();
+
         bool enableHysteresis = simulator_.problem().materialLawManager()->enableHysteresis();
-        bool enableSwatinit = simulator_.vanguard().eclState().get3DProperties().hasDeckDoubleGridProperty("SWATINIT");
+        bool enableSwatinit = es.get3DProperties().hasDeckDoubleGridProperty("SWATINIT");
+
+        bool enableTemperature = es.getSimulationConfig().isThermal();
+
         std::vector<Opm::RestartKey> solutionKeys{
             {"PRESSURE" , Opm::UnitSystem::measure::pressure},
             {"SWAT" ,     Opm::UnitSystem::measure::identity, static_cast<bool>(FluidSystem::phaseIsActive(FluidSystem::waterPhaseIdx))},
             {"SGAS" ,     Opm::UnitSystem::measure::identity, static_cast<bool>(FluidSystem::phaseIsActive(FluidSystem::gasPhaseIdx))},
-            {"TEMP" ,     Opm::UnitSystem::measure::temperature}, // always required for now
+            {"TEMP" ,     Opm::UnitSystem::measure::temperature, enableTemperature}, // restore if thermal run
             {"RS" ,       Opm::UnitSystem::measure::gas_oil_ratio, FluidSystem::enableDissolvedGas()},
             {"RV" ,       Opm::UnitSystem::measure::oil_gas_ratio, FluidSystem::enableVaporizedOil()},
             {"SOMAX",     Opm::UnitSystem::measure::identity, simulator_.problem().vapparsActive()},


bska:
Note: This pull request depends on OPM/opm-common#546 and must not be merged before that one is installed.

bska:
jenkins build this opm-common=546 please

ebos/ecloutputblackoilmodule.hh
```diff
@@ -229,7 +229,9 @@ public:
         }
         // and oil pressure
         oilPressure_.resize(bufferSize, 0.0);
-        temperature_.resize(bufferSize, 0.0);
+
+        if (simulator_.vanguard().eclState().getSimulationConfig().isThermal())
```
andlaus: ebos is more generic than eclipse here: even if energy is not conserved, the temperature does not necessarily need to be uniform throughout the domain. granted, not much is influenced by temperature, but anyway...

(also, it is probably better to use `if (enableEnergy)` on the simulator side.)

ebos/eclwriter.hh
```diff
@@ -318,13 +318,18 @@ public:
 
     void restartBegin()
     {
+        const auto& es = simulator_.vanguard().eclState();
```
andlaus: s/es/eclState/

ebos/eclwriter.hh
```diff
@@ -318,13 +318,18 @@ public:
...
+
         bool enableHysteresis = simulator_.problem().materialLawManager()->enableHysteresis();
-        bool enableSwatinit = simulator_.vanguard().eclState().get3DProperties().hasDeckDoubleGridProperty("SWATINIT");
+        bool enableSwatinit = es.get3DProperties().hasDeckDoubleGridProperty("SWATINIT");
+
+        bool enableTemperature = es.getSimulationConfig().isThermal();
```
andlaus: `enableEnergy` should already be present? (also be aware that non-uniform temperature does not necessarily imply energy conservation.)

ebos/eclwriter.hh
```diff
@@ -318,13 +318,18 @@ public:
...
+
         std::vector<Opm::RestartKey> solutionKeys{
             {"PRESSURE" , Opm::UnitSystem::measure::pressure},
             {"SWAT" ,     Opm::UnitSystem::measure::identity, static_cast<bool>(FluidSystem::phaseIsActive(FluidSystem::waterPhaseIdx))},
             {"SGAS" ,     Opm::UnitSystem::measure::identity, static_cast<bool>(FluidSystem::phaseIsActive(FluidSystem::gasPhaseIdx))},
-            {"TEMP" ,     Opm::UnitSystem::measure::temperature}, // always required for now
+            {"TEMP" ,     Opm::UnitSystem::measure::temperature, enableTemperature}, // restore if thermal run
```
andlaus: adding a third argument to `Opm::RestartKey` for enabling/disabling it is a quite strange design decision IMO. the key should simply not be added if it does not make sense. maybe there were deeper reasons?

andlaus:
@bska: are you going to fix this up, or shall it be closed?

bska:
> are you going to fix this up, or shall it be closed?

There's nothing really to fix.  This, or something very similar to this, is necessary to enable restarting Flow from an ECLIPSE-compatible restart file.  In particular, we must not require that `TEMP` is present in the result set from which we're restarting unless we're actually in a thermal run.  That also suggests that we should not bother outputting a temperature vector to the result set unless we're in a thermal run.

Since these changes are in the files that directly implement ECLIPSE-style input/output, EBOS's notion of energy conservation is not really relevant.

ebos/eclwriter.hh
bska: > adding a third argument to `Opm::RestartKey` for enabling/disabling it is a quite strange design decision IMO.

That's not really what the flag does.  Rather it indicates whether or not the associated vector must be present in the result set when restarting.  If a requisite vector is not present, then the restart procedure (and simulation run) terminates.

andlaus:
> There's nothing really to fix. This, or something very similar to this, is necessary to enable restarting Flow from an ECLIPSE-compatible restart file.

I did not mean anything fundamental, just responding to the review.

> That also suggests that we should not bother outputting a temperature vector to the result set unless we're in a thermal run.

I'm pretty sure all the `if` messiness that comes with this is not worth the benefits: storing temperature does not require a significant amount of memory and it also does not adversely affect performance (?, haven't checked but would be very surprised)

> Since these changes are in the files that directly implement ECLIPSE-style input/output, EBOS's notion of energy conservation is not really relevant.

since this is ebos core code, it is very well relevant for it ;). use `enableEnergy` instead of getting this directly from the parser to avoid inconsistencies within the simulator. if you really want to do it the Right Way (TM), you can change the flow main function to use the internalized parser flag to decide if energy ought to be enabled or not.

bska:
> > That also suggests that we should not bother outputting a temperature vector to the result set unless we're in a thermal run.
> 
> I'm pretty sure all the `if` messiness that comes with this is not worth the benefits: storing temperature does not require a significant amount of memory and it also does not adversely affect performance (?, haven't checked but would be very surprised)

I suppose there's one piece of information that's not clearly communicated here.  If we want to be strictly compatible with ECLIPSE, then we're not **allowed** to store temperature data in the restart file unless it's a thermal run (i.e., unless either of the `TEMP` or `THERMAL` keywords are specified in `RUNSPEC`).  Conversely, we can't expect there to be temperature data in the restart file when restarting a non-thermal run from a strictly ECLIPSE-compatible restart file.

In other words, the appropriate condition for saving/restoring temperature data is something along the lines of
```C++
bool saveRestoreTemp = (is_strictly_eclipse_compatible_restart)
    ? (runspec.has("TEMP") || runspec.has("THERMAL"))
    : GET_PROP_VALUE(TypeTag, EnableEnergy);
```
I'll update the PR accordingly.

andlaus:
> If we want to be strictly compatible with ECLIPSE, then we're not allowed to store temperature data in the restart file unless it's a thermal run 

I got this, and I agree that it shouldn't. my point is that we are guaranteed not to be in a thermal run if `enableEnergy` is false and we are guaranteed to be in one if it is true. IOW, the simulator does not currently care whether the parser flag is set or not.

bska:
> my point is that we are guaranteed not to be in a thermal run if `enableEnergy` is false and we are guaranteed to be in one if it is true.

Okay, so that suggests that it's possible to have `enableEnergy == true` when the run does not specify either of the `TEMP` or `THERMAL` keywords.  That's fine if we're not trying to create ECLIPSE-compatible restart files.  It is not fine if we are trying to create ECLIPSE-compatible restart files.  ECLIPSE-compatible restart files must not have temperature data unless the run specifies `TEMP` or `THERMAL`.

andlaus:
the simultor enables energy conservation if and only if the `EnableEnergy` property is set to `true`. it does not actually care about the parser here, but hopefully the property is set iff your new `EclipseState` flag is set to `true`.

(See it this way: starting an isothermal simulation for a deck that contains `TEMP` *might* make sense in some scenarios. At least I have much less trouble coming up with applications for this than for being able to set the `SummaryConfig` and `Schedule` objects externally.)

tl;dr: use `enableEnergy` within the simulator and make sure that it corresponds to the parser flag in `flow.cpp`!

andlaus:
I'd like to merge this, but you still need to do the `simulator_.vanguard().eclState().getSimulationConfig().isThermal())` to `enableEnergy` rename.

ebos/eclwriter.hh
andlaus: I suppose that's the problem with flag arguments: its often not clear what they to at the call-site.

bska:
> you still need to do the [`isThermal()` to `enableEnergy`] rename

Okay, I'll try.  That's a much bigger change though.

andlaus:

andlaus:
> That's a much bigger change though.

should be approximately two lines of code...

bska:
> > That's a much bigger change though.
> 
> should be approximately two lines of code...

In this repository, yes.  However, I will also need to touch at least opm-simulators, so there's more effort to coordinate and test everything.

andlaus:
(I wanted to say that the opm-simulator changes are not required from a technical point of view, bit are nice to have because they reduce code duplication between opm-simulators and opm-common.)

bska:
I'm closing this.  It is strictly superseded by #433.

PR #414, opened 2018-11-05
fix the exception-handling-during-linearization code

andlaus:
these exceptions must be handled within the loop over the grid, otherwise there will be a casualty amongst the threads.

### DIFF
--- a/ewoms/parallel/threadedentityiterator.hh
+++ b/ewoms/parallel/threadedentityiterator.hh
@@ -68,6 +68,14 @@ public:
     bool isFinished(const EntityIterator& it) const
     { return it == sequentialEnd_; }
 
+    // make sure that the loop over the grid is finished
+    void setFinished()
+    {
+        mutex_.lock();
+        sequentialIt_ = sequentialEnd_;
+        mutex_.unlock();
+    }
+
     // prefix increment: goes to the next element which is not yet worked on by any
     // thread
     EntityIterator increment()
--- a/ewoms/disc/common/fvbaselinearizer.hh
+++ b/ewoms/disc/common/fvbaselinearizer.hh
@@ -450,12 +450,10 @@ private:
 
         *matrix_ = 0.0;
 
-#ifdef _OPENMP
-        // to avoid a race condition if two threads handles an exception at
-        // the same time, we use an explicit lock to control access to the
-        // exception storage amongst thread-local handlers
+        // to avoid a race condition if two threads handle an exception at the same time,
+        // we use an explicit lock to control access to the exception storage object
+        // amongst thread-local handlers
         std::mutex exceptionLock;
-#endif
 
         // storage to any exception that needs to be bridged out of the
         // parallel block below. initialized to null to indicate no exception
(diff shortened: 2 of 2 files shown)


andlaus:
jenkins build this please

atgeirr:
I do not understand why moving the try/catch inside the loop changes anything? Is the threaded iterator somehow causing havoc otherwise? Also, is this a zero-overhead thing performance-wise?

atgeirr:
> otherwise that thread which experiences an exception bails out of the loop while the remaining ones still need to finish it. This has nothing to do with the threaded element iterator.

But why do the other threads care if of their kind bailed out? Cannot this be synchronized after the loop?

andlaus:
currently they don't. that's why the process can only proceed if all of them are finished instead of bailing out if one thread received an exception. In addition if there is an exception there's one less thread to do the work. It's pretty much a corner case and only leads to slightly worse performance, but still.

andlaus:
threaded programming is fun, but I guess that I repeat myself ;)

andlaus:
@atgeirr, @totto82: mind to merge?

atgeirr:
Despite your explanation, I still do not understand the rationale for the change. The exceptionPtr gets set correctly in any case (although if multiple threads fail, the exception stored there is that of the last-to-fail thread), and after the parallel block it is thrown. What is wrong with one of the threads bailing out of the loop? Will it not wait for the others to finish the parallel section before continuing?

atgeirr:
> in the multi-threaded the remaining threads need to finish the loop after an exception, for the sequential case, the loop gets aborted. although this does not lead to wrong semantics, it is very inelegant.

By "need", do you mean that it is a bug to let remaining threads abort the loop? Or just that this is the current behaviour? I think I need an example... Given two threads, and one of them throws after processing half its elements. What happens today, what happens after the PR, and what would happen in an ideal world?

andlaus:
I mean that the loop does not get aborted for the other threads, i.e., since all threads synchronize at the end of the parallel block, the threads that received an exception need to stay idle in the interim and the work of the other threads is thrown away.

in your example, the second thread has to process the remaining half of the elements alone while his buddy idles around at the end of the parallel section. after this PR, the buddy will terminate the loop as soon as it is finished linearizing its currently processed element.

ewoms/parallel/threadedentityiterator.hh
```diff
@@ -68,6 +68,14 @@ public:
...
     { return it == sequentialEnd_; }
 
+    // make sure that the loop over the grid is finished
+    void setFinished()
+    {
+        mutex_.lock();
+        sequentialIt_ = sequentialEnd_;
+        mutex_.unlock();
+    }
+
```
atgeirr: Just noticed a potential bug (or doc failure) in the function below: increment() behaves as postincrement, not preincrement.

atgeirr:
Thanks for the final explanation, it was successful: I now understand your point. So this does not fix a bug at all, and is quite unnecessary... On the other hand, it may break out of a failing loop faster. So ok, it can be merged (since it does not make the code any harder to understand than it was already).

andlaus:
jenkins build this please

andlaus:
any further objections to merging this?

atgeirr:
> any further objections to merging this?

Other than Jenkins not green, no.

atgeirr:
jenkins build this please

andlaus:
okay, I suppose that this is ready. merging on behalf of @atgeirr.

PR #415, opened 2018-11-05
DO-NOT-MERGE Don't scale the well rates

totto82:
The purpose of this patch is to show the reason for the test failure in OPM/opm-simulators#1555


andlaus:
Since OPM/opm-simulators#1555 has been merged it seems like this PR has served its purpose, so I'm closing this for now. Please feel free to reopen if there is any need.

PR #416, opened 2018-11-06
Avoid output of trivial fipnum regions

totto82:
This has annoyed me as well. 
Fix issue  OPM/opm-common#543


andlaus:
thanks, looks good. Will merge if jenkins does not object.

andlaus:
jenkins build this please

PR #418, opened 2018-11-07
ebos: clean up the mess that is the schedule and summaryConfig objects a bit

andlaus:
IMO the simulator should not be in the business of managing low-level parser objects in the first place, because that's what EclipseState is supposed to do?!

anyway, since these objects are not needed to decide which simulator to use, they are now always managed internally by the vanguard, i.e., setExternalDeck() does not take them anymore.

this needs a small mop-up in opm-simulators.

### DIFF
--- a/ebos/eclbasevanguard.hh
+++ b/ebos/eclbasevanguard.hh
@@ -153,12 +153,10 @@ public:
      * management of these two objects, i.e., they are not allowed to be deleted as long
      * as the simulator vanguard object is alive.
      */
-    static void setExternalDeck(Opm::Deck* deck, Opm::EclipseState* eclState, Opm::Schedule* schedule, Opm::SummaryConfig* summaryConfig)
+    static void setExternalDeck(Opm::Deck* deck, Opm::EclipseState* eclState)
     {
         externalDeck_ = deck;
         externalEclState_ = eclState;
-        externalSchedule_ = schedule;
-        externalSummaryConfig_ = summaryConfig;
     }
 
     /*!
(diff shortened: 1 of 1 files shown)


andlaus:
jenkins build this opm-simulators=1613 please

atgeirr:
> IMO the simulator should not be in the business of managing low-level parser objects in the first place, because that's what EclipseState is supposed to do?!

The Schedule is not a very low-level object, it is the highest-level abstraction we have for the simulation schedule, including well controls etc. Reducing the coupling between the EclipseState and Schedule was a necessary step towards allowing manipulation of schedules during simulation. Removing the possibility to set this externally means it can only come from the EclipseState directly (i.e. from the deck), with no chance to change it. So I think this should not be merged.

andlaus:
in the context of the simulator the `Schedule` *is* an implementation detail of the input deck parser, agreed? if yes it is  a low-level object in my opinion.

anyway, even if you disagree there's no need to clutter the `setExternalDeck()` calls with it because the only reason for `setExternalDeck()` is having to wait for the parser twice: once to decide which simulator needs to be instantiated and once for the simulator. since neither of these objects is needed to decide, there is also no need to make it possible to instantiate it externally from the simulator.

joakim-hove:
> there is also no need to make it possible to instantiate it externally from the simulator.

Well - in my opinion the possibility to instantiate, manage and most importantly *manipulate* the `Schedule`instance outside of the inner ewoms simulator is very valuable; although I agree we have not actually exploited that possibility.

andlaus:
I am happy to fold once you have a concrete example, but so far the externally instantiated low-level parser objects just create a lot of clutter. To repeat one your own idioms: let's burn that bridge once we come to it.

joakim-hove:
> but so far the externally instantiated low-level parser objects just create a lot of clutter.

You know the ewoms code orders of magnitude better than me; but as I see it there are two possible approaches to this:

1. One can say the ewoms code is the main application, and that objects like `EclipseState` and `Schedule` should be fully managed inside of ewoms data structures.

2. One could say that whatever `mainClass` in opm-simulators is the main application, and that ewoms provides a *simulation/mathematics service* to the simulator.

As I see it this PR moves the code base in direction 1), whereas I personally would rather prefer to move things in direction 2) - i.e. that the ewoms datastructures do *not* own the input/management structures like `Schedule` and `EclipseState`.  

> To repeat one your own idioms: let's burn that bridge once we come to it.

Well - I am currently working on this `ACTIONX` implementation; it is not ready tomorrow - but then I will need access to the `Schedule` instance.

atgeirr:
> in the context of the simulator the `Schedule` _is_ an implementation detail of the input deck parser, agreed? if yes it is a low-level object in my opinion.

No, we have even discussed several times that we want to use the `Well` objects (which are owned by the Schedule and exposed through it) in well code (part of the simulator) instead of the old `Wells` C-struct for example.

> anyway, even if you disagree there's no need to clutter the `setExternalDeck()` calls with it because the only reason for `setExternalDeck()` is to avoid having to wait for the parser twice: once to decide which simulator needs to be instantiated and once for the simulator. since neither of these objects is needed to decide, there is also no need to make it possible to instantiate it externally from the simulator.

Well, then there are two reasons: the one you describe, and the other to allow modification of the deck-given schedule. Perhaps splitting it in setExternalDeck() and setExternalSchedule() would be clearer?

andlaus:
> As I see it this PR moves the code base in direction 1), whereas I personally would rather prefer to move things in direction 2)

you're welcome to make proposals. As long as they are unambigously better than the current state in master they will be merged, if I disagree they might be merged as well,  just don't expect me to like or maintain this stuff in that case ;)

> No, we have even discussed several times that we want to use the Well objects (which are owned by the Schedule and exposed through it) in well code (part of the simulator) instead of the old Wells C-struct for example.

this is a flawed argument, because the well has full access to the simulator object and hence also to the schedule.

> [...]  in direction 2) - i.e. that the ewoms datastructures do not own the input/management structures like Schedule and EclipseState.

If you mean this seriously then the schedule should be completely managed by The Mighty Parser, i.e., `EclipseState`.

> Perhaps splitting it in setExternalDeck() and setExternalSchedule() would be clearer?

yes, if the second method was not static.

andlaus:
okay, I've implemented @atgeirr's suggestion, so the schedule and summaryConfig objects can now optionally be specified externally. any objections to merging this?

andlaus:
jenkins build this opm-simulators=1613 please

ebos/eclbasevanguard.hh
```diff
@@ -205,63 +203,60 @@ public:
...
             deck_ = externalDeck_;
             eclState_ = externalEclState_;
-            schedule_ = externalSchedule_;
-            summaryConfig_ = externalSummaryConfig_;
         }
 
+        // create the schedule and summary config objects. Note that if eclState is
+        // supposed to represent the internalized version of the deck, this constitutes a
+        // layering violation.
+        eclSchedule_.reset(new Opm::Schedule(*deck_, *eclState_, parseContext));
```
atgeirr: It seems to me that these objects are now always constructed, and then if we use for example setSchedule() later, this will be thrown away. That is not less, but more messy than what we have! If setSchedule() etc. also were static, this could be handled in the same way as with the deck. That way all of these objects would function in the same way, making it easier to understand what is going on. Given that the static method approach will have to remain for the deck, why do you not want to use it for the other objects?

andlaus: > It seems to me that these objects are now always constructed, and then if we use for example setSchedule() later, this will be thrown away.

yes, but if creating these objects is cheap, what's the problem. before jumping through huge hoops to avoid that, I'd like to see some numbers that it matters in practice.

> That is not less, but more messy than what we have!

this is in the eye of the beholder: `setExternalDeck()` can now be called without being forced to create schedule and summaryConfig objects. Also, I would also prefer not to have that patch. (you can manipulate the schedule even without it because there is a non-const variant of the `schedule()` method.)

> Given that the static method approach will have to remain for the deck, why do you not want to use it for the other objects?

because it avoids having to declare these static objects outside of the class which is -- frankly -- ugly.

atgeirr: > yes, but if creating these objects is cheap, what's the problem. before jumping through huge hoops to avoid that, I'd like to see some numbers that it matters in practice.

The problem is not one of performance, but one of messiness. I guess this is good enough for now, though.

andlaus:
okay, whatever: I've made the setExternal*() methods static again. I still firmly believe that the simulator should get out of the the business of dealing with low-level parser objects, but my feelings here are not strong enough to put up an extended argument. (BTW: anybody who objects direct usage of the `deck` object within the simulator should agree with me!)

@atgeirr: since all the technical objections should be fixed, would you mind pressing green?

joakim-hove:
> should get out of the business of dealing with low-level parser objects

Hehe - as written there we might agree; but I think interpretations are quite different. As I see it the simulator is located in the opm-simulators repository - and that should own and manage all of the input structures - ewoms should not hold on to these structures at all. But that is for another day.

andlaus:
what I -- and probably everybody else who is not interested in the parser itself -- want is something akin:

```
internaliizedDeck = Opm::parseEclDeck(fileName);
// use internaliizedDeck
```

all that scaffolding that became mandatory over the past years is a pretty annoying distraction, IMO.

joakim-hove:
> what I -- and probably everybody else who is not interested in the parser itself -- want is something akin

And what I want to give you is fully processed `EclipseState`, `Schedule` and `SummaryConfig` instances - ewoms code in the `flow`pipeline should not use neither the `Parser` class nor the `Deck` class. To take it one step further the ewoms code should not even hold on these classes - in particular not the `Schedule` class.

atgeirr:
> @atgeirr: since all the technical objections should be fixed, would you mind pressing green?

Sure!

andlaus:
> And what I want to give you is fully processed EclipseState, Schedule and SummaryConfig instances

agreed. but currently one has to instantiate all of these objects plus the deck manually. that's clunky and it is currently not even possible to avoid the `deck` object because not all required information is completely internalized! For the first issue a "one level higher than EclipseState" class which does all of this pesky object management would do the trick. (So far, I just assumed that `EclipseState` is supposed to represent the highest level of abstraction for ECL decks.)

@atgeirr: thanks for merging.

PR #420, opened 2018-11-08
Allow abs. tol. in linear solver to be set by prop

rolk:
The magic value -1 indicates that a value based on an order less of the
Newton tolerance should be used. The special value 0. cannot be used to
signal this, as that value is already used to indicate that the
criterion should be deactivated; however there were no way of setting
this to zero before this change.

Credits belong to Tor-Harald, who realized that an over-eager
convergence check was the reason that the simulations didn't update.


andlaus:
jenkins is picky about its name

andlaus:
jenkins build this please

andlaus:
okay, all green. merging, thanks for the fix!

PR #421, opened 2018-11-08
Avoid singul. for zero mole frac. in one-phase

rolk:
Having a multi-component, single-phase, gravity-driven system with parts
of the domain initialized to zero mole fraction of one of the
components, will result in a singular equation and zero component
fugacity.

This change introduce a small addition to the fugacity, not enough the
skew the results, but enough for the problem to be numerically
tractible.

Credits belong to Tor-Harald, who is the real mastermind behind tracking
down and fixing this problem.

### DIFF
--- a/ewoms/models/ncp/ncpintensivequantities.hh
+++ b/ewoms/models/ncp/ncpintensivequantities.hh
@@ -88,6 +88,27 @@ class NcpIntensiveQuantities
     typedef Ewoms::EnergyIntensiveQuantities<TypeTag, enableEnergy> EnergyIntensiveQuantities;
     typedef typename FluxModule::FluxIntensiveQuantities FluxIntensiveQuantities;
 
+    typedef Opm::MathToolbox<Evaluation> Toolbox;
+    /*!
+     * \brief Ensure that a quantity is above a minimum threshold, while keeping
+     * the derivative of a variable.
+     */
+    static void truncateToThreshold_(ComponentVector& vec,
+                                     Scalar threshold,
+                                     unsigned compIdx,
+                                     unsigned timeIdx)
+    {
+        if(vec[compIdx] < threshold) {
+            // don't include derivatives from the initial state
+            if (timeIdx == 0) {
+                vec[compIdx] = Toolbox::createVariable(threshold, compIdx);
+            }
+            else {
+                vec[compIdx] = Toolbox::createConstant(threshold);
+            }
+        }
+    }
+
 public:
     NcpIntensiveQuantities()
     {}
@@ -136,8 +157,16 @@ public:
 
         ComponentVector fug;
         // retrieve component fugacities
-        for (unsigned compIdx = 0; compIdx < numComponents; ++compIdx)
+        for (unsigned compIdx = 0; compIdx < numComponents; ++compIdx) {
             fug[compIdx] = priVars.makeEvaluation(fugacity0Idx + compIdx, timeIdx);
+
+            // as reservoirs have in the order of 10^2 bars pressure, the
+            // fugacities should be in the order of 10^7 Pascal; if we want a
+            // relative change of 10^-10 (see ewoms/models/blackoil/
+            // blackoilpolymermodules.hh:addStorage), we need a minimum of
+            // 10^{7-10} = 10^-3.
+            truncateToThreshold_(fug, 1.e-3, compIdx, timeIdx);
+        }
 
         // calculate phase compositions
         const auto *hint = elemCtx.thermodynamicHint(dofIdx, timeIdx);


andlaus:
jenkins build this please

andlaus:
looks good, but Jenkins complains and thus prevents me from pressing Green.

andlaus:
just assign a scalar value, i.e.,

```c++
fug[compIdx] = 1e-3;
```

andlaus:
(you will want to throw away the derivatives in that case anyway!)

andlaus:
(if you don't, you the results of FD and AD will differ. If you ask me, we had enough head scratching with this already ;))

totto82:
Could` fug[compIdx] = Toolbox::createVariable(1e-3, fugacity0Idx + compIdx) `work?

andlaus:
yes, but the difference between FD and AD won't go away.

Anyway, what you want to do is to limit the fugacity of a component to 10^-3 at least, and that limit does not depend on anything, so if it kicks in there should not be any non-zero derivatives.

andlaus:
that aside: what happens if you use 0 instead of 1e-3?

rolk:
> what happens if you use 0 instead of 1e-3

Then we're back to the initial condition of the code: That the fugacities were zero is what caused the problem to begin with.

rolk:
> you will want to throw away the derivatives in that case anyway!

Resetting the derivatives to zero gives us the 'non-finite update' condition; we need something that preserves the derivatives, but that gives a numerical result for the zero case.

> if you don't, the results of FD and AD will differ

Well, FD will have to get the derivatives some other way (than from the constant); how do we know that they would be zero?

andlaus:
>> what happens if you use 0 instead of 1e-3
>
>  Then we're back to the initial condition of the code: That the fugacities were zero is what caused the problem to begin with.

but why? the fugacities only appear in the numerator when calculating mole fractions:

x_\alpha^\kappa = f^\kappa / (\Phi_alpha^\kappa*p_\alpha)

andlaus:
> Well, FD will have to get the derivatives some other way (than from the constant); how do we know that they would be zero?

in FD, the derivative for the fugacities will be automatically zero if f and f+\eps are below the threshold.

andlaus:
hm. okay I thought about this: the problem is that when the threshold kicks in, the residual does not depend on one of the primary variables anymore. the "real" fix is to make sure that the fugacities are not negative after a newton-update, i.e, you need to modify [`NcpNewtonMethod::updatePrimaryVariables_()`](https://github.com/OPM/ewoms/blob/master/ewoms/models/ncp/ncpnewtonmethod.hh#L121)

rolk:
> make sure that the fugacities are not negative
 
Rebased to new fix. This seems to work, too. ParaView does not work from home, so I'll have to check how the result looks tomorrow.

jenkins build this please

andlaus:
okay, it does not hurt to wait a bit for better testing, I suppose. Jenkins is not only picky when it comes to its name, it is also pretty dumb: you need to address it explicitly, i.e., it won't react on commands embedded into other comments:

andlaus:
jenkins build this please

andlaus:
> Truncating negative fugacities, as in f75c3dd, results in "non-finite update". Even setting the fugacities to strictly positive (1.e-3) does that.

I wonder how negative fugacities could happen: this probably implies that the fugacity coefficient somehow becomes negative, because the newton update restricts changes of fugacity to between 0 and the minimum activity coefficient (i.e., fugacity coefficient multiplied by pressure).

rolk:
> probably implies that the fugacity coefficient somehow becomes negative

Not likely, in my opinion, as the "ideal gas" branch, which forced the fugacity coefficient to 1., also had this problem.

rolk:
jenkins build this please

rolk:
Reverted back to the previous version, with improvements that hopefully also will work (or at least compile) with FD.

andlaus:
ok,  this seems to simply avoid a deeper bug in the `CompositionFromFugacities` constraint solver: the derivatives that are spit out by the Newton method if fed with Evaluation objects seem to be incorrect. Unfortunately I've some trouble to wrap my head around that. Since the derivatives produced by this PR are also wrong -- just non-zero -- and it is unphysical because it prevents a component from disappearing completely, I'll close this for now. We should continue to think about how to do the Newton method with AD-objects, though. Also, thanks for the discussion!

andlaus:
> Isn't this just the same trick that is used by for instance the polymer model?

I don't think so because for this PR, the mass of the component is not conservered anymore if it is below the threshold whereas in the polymer blackoil extension, the result is just used to avoid a division by zero (?)

> If a less hacky fix exists, it would certainly be better. We would continue to use this patch for our simulations until we come up with a better solution.

we need to fix the Newton-Raphson method so that it also works correctly with the derivatives of Evaluation objects instead of just the values. For now, @rolk can use an ideal mixture with trivial fugacity coefficients because the fugacities are only relevant in multi-phase scenarios. (also, the result is significantly faster.)

rolk:
>  fugacities are only relevant in multi-phase scenarios

Won't it effect the level of the pressure?

PR #422, opened 2018-11-09
fix the auto-determined absolute tolerance for the linear solver

andlaus:
this was set to -1 by default because the computed value was not used anymore after ewoms#420. The automatically computed absolute tolerance now is also tighter than it was before which hopefully reduces issues.


andlaus:
jenkins build this please

andlaus:
this is also a clear bug. self-merging.

rolk:
Clearly brown-paper-bag-worthy insufficient testing from my side; wonder why the compiler doesn't give "Statement has no effect" warnings here?

andlaus:
mine (GCC 8.2) does, at least in release mode. that was the reason why I found this issue. anyway, not a big deal; the absolute tolerance was just always disabled before this...

PR #365, opened 2018-08-20
ISTLMatrixBackend

dr-robertk:
This PR adds a wrapper for ISTL::BCRSMatrix such that other matrix/solver backends could be used in the future.

PR #423, opened 2018-11-12
Sparse matrix abstraction

andlaus:
this add an abstraction layer for the matrix between the linearizer and linear solver. it allows to assemble the Jacobian matrices directly into the native format expected by linear solver. So far, only backends using Dune::BCRSMatrix are provided, but there are work-in-progress patches for dune-fem, vienna-CL and PETSc backends.

this superseeds #365 and also requires a cleanup PR in opm-simulators.


andlaus:
jenkins build this opm-simulators=1617

andlaus:
jenkins build this opm-simulators=1617

andlaus:
jenkins build this opm-simulators=1617 please

akva2:
jenkins build this opm-simulators=1617 please

andlaus:
jenkins build this opm-simulators=1617 please

andlaus:
jenkins build this opm-simulators=1617 please

dr-robertk:
Thanks for reviewing and improving. Looks good from my side. Contains all the features I needed. The methods naming on matrix adapter class I liked better before but I'm fine with the current naming also.

dr-robertk:
Should I merge this? It's ok from my side.

atgeirr:
I looked briefly at a previous incarnation of this, but would appreciate if I can take the time until tomorrow to test it (and its opm-simulators cousin).

dr-robertk:
Yes, of course. Take some time to test this please.

ewoms/linear/istlsparsematrixadapter.hh
```diff
@@ -0,0 +1,192 @@
...
+ */
+template <class MatrixBlockType, class AllocatorType=std::allocator<MatrixBlockType> >
+class IstlSparseMatrixAdapter
+{
+public:
+    //! \brief Implementation of matrix
+    typedef Dune::BCRSMatrix<MatrixBlockType, AllocatorType> IstlMatrix;
+
+    //! \brief block type forming the matrix entries
+    typedef typename IstlMatrix::block_type MatrixBlock;
```
atgeirr: We should static_assert that MatrixBlock is the same type as MatrixBlockType (if that is not always true, the code is too clever...).

ewoms/linear/istlsparsematrixadapter.hh
```diff
@@ -0,0 +1,192 @@
...
+     */
+    void addToBlock(const size_t rowIdx, const size_t colIdx, const MatrixBlock& value)
+    { (*istlMatrix_)[rowIdx][colIdx] += value; }
+
+    /*!
+     * \brief Commit matrix from local caches into matrix native structure.
+     *
+     * For the ISTL adapter this is unnecessary because there is no caching mechanism.
+     */
+    void commit()
```
atgeirr: The need for distinct commit() vs. finalize() escapes me, I assume other backends distinguish them? Still, why not just finalize(), and let finalize() also do the commit work if necessary for the backend?

ewoms/linear/istlsparsematrixadapter.hh
```diff
@@ -0,0 +1,192 @@
...
+    { return columns_; }
+
+    /*!
+     * \brief Set all matrix entries to zero.
+     */
+    void clear()
+    { (*istlMatrix_) = Scalar(0.0); }
+
+    /*!
+     * \brief Set given row to zero except for the diagonal entry which is set to one.
```
atgeirr: I think this is misleading (and not just because one can choose a different diagonal constant). The diagonal entry will be set only if the row already has a nonzero diagonal entry, otherwise line 141 will never be executed. While it is probably ok for our uses (we will probably always have a diagonal entry), as a general adapter this should be addressed either in the code or in the comment.

ewoms/linear/istlsparsematrixadapter.hh
andlaus: changed. Not so sure if it adds much value, though.

ewoms/linear/matrixblock.hh
```diff
@@ -0,0 +1,303 @@
...
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with OPM.  If not, see <http://www.gnu.org/licenses/>.
+
+  Consult the COPYING file in the top-level source directory of this
+  module for the precise wording of the license and the list of
+  copyright holders.
+*/
+#ifndef EWOMS_MATRIX_BLOCK_HH
```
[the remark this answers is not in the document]
andlaus: The version from opm-simulator cannot be used within eWoms, so it needs to be lowered. The reason why the opm-simulators version is not deleted in the compagnion PR yet is to avoid code changes.

BTW: I did not get any measureable performance regression for Norne if I replaced `Opm::MatrixBlock` by `Dune::FieldMatrix`, so maybe this hack can be dropped entirely? (I tested this using Dune 2.6.)

ewoms/linear/istlsparsematrixadapter.hh
andlaus: I'm not 100% sure either, but I think `finalize()` is supposed to create the sparsity pattern, while `commit()` deals with things like transferring the values of entries from main memory to the data structure used by the linear solver, e.g., on the GPU. Is that correct, @dr-robertk?

ewoms/linear/istlsparsematrixadapter.hh
andlaus: ah: you mean that there is no entry in the sparsity pattern for the main diagonal? this seems like quite a corner case, but I've adapted the comment.

ewoms/linear/matrixblock.hh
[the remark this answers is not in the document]
atgeirr: > BTW: I did not get any measureable performance regression for Norne if I replaced `Opm::MatrixBlock` by `Dune::FieldMatrix`, so maybe this hack can be dropped entirely? (I tested this using Dune 2.6.)

I think Dune versions are significant for this, so perhaps we can safely drop it once we require 2.6.

andlaus:
jenkins build this opm-simulators=1617 please

andlaus:
the jenkins error seems to be quite unrelated to this:

```
/var/lib/jenkins/workspace/ewoms-PR-builder/deps/opm-simulators/opm/autodiff/ISTLSolver.hpp:76:53: error: there are no arguments to 'isIORank' that depend on a template parameter, so a declaration of 'isIORank' must be available [-fpermissive]
           isIORank_(isIORank(parallelInformation_arg)),
```

(this PR should not affect the legacy ISTLSolver class?)

atgeirr:
> (this PR should not affect the legacy ISTLSolver class?)

It did, by removing an include that transitively included something needed. I described the necessary change in OPM/opm-simulators#1617.

andlaus:
jenkins build this opm-simulators=1617 please

andlaus:
*ARGH!*: MPIUtilities.hpp did not include all header files necessary either. fixed!

andlaus:
jenkins build this opm-simulators=1617 please

andlaus:
it seems like jenkins did not pick up the last trigger. maybe because the commit ID is unchanged (only the opm-simulators PR needed a fix), maybe because of something else. let's retry.

andlaus:
jenkins build this opm-simulators=1617 please

andlaus:
@akva2: jenkins does not pick this up, I'll close and reopen the PR, but this is not a good solution IMO.

andlaus:
jenkins build this opm-simulators=1617 please

atgeirr:
If there is no change in this repo, and that is the reason for the failure, perhaps you should trigger from the other PR instead?

atgeirr:
Thanks for the effort, merging.

PR #425, opened 2018-11-12
Apply multiplicators from EDITNNC from the Deck.

blattms:


blattms:
jenkins build this opm-common=560 please

blattms:
rebased onto master

blattms:
jenkins build this opm-common=560 please

blattms:
jenkins build this opm-common=560 please

blattms:
Seems like I am missing the jenkins foo for this module. Triggered the build from OPM/opm-common#560

akva2:
Access granted.

alfbr:
We have tested the implementation, and confirm correct results. From our side this is ready to be merged.

blattms:
Removing WIP as this is tested now.

blattms:
jenkins build this opm-common=560 please

alfbr:
It would be good to have this (and the companion PR in opm-common) merged :-)

joakim-hove:
Ok - we might be waiting in a deadlock, the one in common has been ready for merging some time. If nothing happens during the day I will merge them both at 1500 today.

blattms:
Great.

blattms:
Will do a shameless self merge tonight.

blattms:
but would prefer if somebody else took the time...

ebos/ecltransmissibility.hh
```diff
@@ -504,6 +505,59 @@ private:
...
+
+        for (; elemIt != elemEndIt; ++elemIt) {
+            int elemIdx = elementMapper.index(*elemIt);
+            int cartElemIdx = vanguard_.cartesianIndexMapper().cartesianIndex(elemIdx);
+            globalToLocal[cartElemIdx] = elemIdx;
+        }
+
+        // editNNC is supposed to only reference non-neighboring connections and not
+        // neighboring connections. Use all entries for scaling if there is an NNC.
+        for(auto nnc = editNNC.data().begin(), end = editNNC.data().end(); nnc != end;)
```
atgeirr: Formatting nitpicks: space after for, remove extra semicolon after `nnc != end`.

ebos/ecltransmissibility.hh
```diff
@@ -504,6 +505,59 @@ private:
...
+                }
+                else
+                {
+                    std::ostringstream sstr;
+                    sstr << "Cannot edit NNC from " << nnc->cell1 <<" to "<<nnc->cell2
+                         << " as it does not exist";
+                    Opm::OpmLog::warning(sstr.str());
+                }
+                ++nnc;
+            }while( nnc!= end && c1==nnc->cell1 && c2==nnc->cell2);
```
atgeirr: Nitpick: space before and after while.

ebos/ecltransmissibility.hh
```diff
@@ -504,6 +505,59 @@ private:
...
+            {
+                if ( candidate != trans_.end() )
+                {
+                    // NNC exists
+                    candidate->second *= nnc->trans;
+                }
+                else
+                {
+                    std::ostringstream sstr;
+                    sstr << "Cannot edit NNC from " << nnc->cell1 <<" to "<<nnc->cell2
```
atgeirr: Nitpick: space around `<<`.

ebos/ecltransmissibility.hh
```diff
@@ -504,6 +505,59 @@ private:
...
+            int elemIdx = elementMapper.index(*elemIt);
+            int cartElemIdx = vanguard_.cartesianIndexMapper().cartesianIndex(elemIdx);
+            globalToLocal[cartElemIdx] = elemIdx;
+        }
+
+        // editNNC is supposed to only reference non-neighboring connections and not
+        // neighboring connections. Use all entries for scaling if there is an NNC.
+        for(auto nnc = editNNC.data().begin(), end = editNNC.data().end(); nnc != end;)
+        {
+            auto c1 = nnc->cell1, c2 = nnc->cell2;
```
atgeirr: I am always nervous when I see auto used for two variables like this, will they be of the same type, or can they differ? In this case it is probably all fine, though.

ebos/ecltransmissibility.hh
```diff
@@ -504,6 +505,59 @@ private:
...
+                {
+                    // NNC exists
+                    candidate->second *= nnc->trans;
+                }
+                else
+                {
+                    std::ostringstream sstr;
+                    sstr << "Cannot edit NNC from " << nnc->cell1 <<" to "<<nnc->cell2
+                         << " as it does not exist";
+                    Opm::OpmLog::warning(sstr.str());
```
atgeirr: Please add a tag to this warning, or we can get flooded by warnings, such as `warning("EDITNNC unknown NNC", sstr.str())`.

ebos/ecltransmissibility.hh
```diff
@@ -351,6 +351,7 @@ public:
 
         // potentially overwrite and/or modify  transmissibilities based on input from deck
         updateFromEclState_();
+        applyEditNNC_(elemMapper);
```
atgeirr: It would make sense to call applyEditNNC_() from updateFromEclState_(), it is after all getting everything from the EclipseState so it is a little misleading to not do that?

ebos/ecltransmissibility.hh
atgeirr: I think the do...while logic here is unnecessary. I assume it is there to handle the case when there are multiple edits to the same transmissibility? If so, then after removing the do...while everything should still work just nice.

Keeping the do...while can make for a tiny performance improvement in cases with many edits to the same nnc, but that would be exotic I think. Also, then the do...while should be inside the `if (candidate != trans_.end())`-branch. All in all I prefer to remove it.

ebos/ecltransmissibility.hh
blattms: IMHO this would depend on how eclipse handles this. Is there a consensus on how this should be here?

ebos/ecltransmissibility.hh
blattms: In principal I agree but I am reluctant to do this as we might introduce errors. It would mean that others would need to retest their models which seems like a waste of resources. Maybe you could do this in a different PR?

ebos/ecltransmissibility.hh
atgeirr: Ok. Perhaps the old function should be named "setFromEclState_" to be more clear, but that is not for this PR.

ebos/ecltransmissibility.hh
atgeirr: I see, this is a very unusual for loop, and I failed to notice its unusualness the first time around (I see now that the increment in the do...while is what keeps it all rolling). If you *really* need this convoluted flow control, it must be commented clearly at least, I still fail to see that it is needed however?

blattms: Nevermind, I will do the loop change. But I insist that you do the testing.

ebos/ecltransmissibility.hh
```diff
@@ -504,6 +505,62 @@ private:
...
+                ++nnc;
+                std::ostringstream sstr;
+                sstr << "Cannot edit NNC from " << nnc->cell1 << " to " << nnc->cell2
+                     << " as it does not exist";
+                Opm::OpmLog::warning(sstr.str());
+            }
+            else
+            {
+                // NNC exists
+                while ( nnc!= end && c1==nnc->cell1 && c2==nnc->cell2 )
```
atgeirr: It is simpler now which is good, but I still do not understand the need for the inner while loop here?

atgeirr:
> Refactored loop. Seems like the results for opm-tests/edit-nnc match those before the refactoring in the eye norm. Would be good if @atgeirr would double check.

I am not able to do any testing (on which case btw?) until Friday at the earliest. If results are identical as before, and @alfbr verified the earlier version, we should be fine?

ebos/ecltransmissibility.hh
blattms: You mean something like this (code completely untested):
```
        auto oldc1=-1, oldc2=-1;
        auto candidate = trans_.find(isId_(low, high));
        for (auto nnc = editNNC.data().begin(), end = editNNC.data().end(); nnc != end; ++nnc)
        {
            auto c1 = nnc->cell1, c2 = nnc->cell2;
            auto low = globalToLocal[c1], high = globalToLocal[c2];
            if ( low > high)
            {
                std::swap(low, high);
            }
            if ( oldc1 != c1 || oldc2 != c2 )
            {
                candidate = trans_.find(isId_(low, high));

                if ( candidate == trans_.end() )
               {
                   std::ostringstream sstr;
                   sstr << "Cannot edit NNC from " << nnc->cell1 << " to " << nnc->cell2
                        << " as it does not exist";
                   Opm::OpmLog::warning(sstr.str());
                  continue;
               }
               // NNC exists
                oldc1 = c1;
                oldc2 = c2;
          }
           candidate->second *= nnc->trans;
        }
```
Is that really more elegant? Seems like more lines to save multiple (potentially expensive) finds.

blattms:
Your call.

alfbr:
I will not be able to conduct any testing until next week earliest, so this one is yours I believe @atgeirr

ebos/ecltransmissibility.hh
atgeirr: I think a single loop is at least more readable. I still do not see why it is so important to treat repeating (c1, c2) pairs in a special way, presumably for performance, when this is not likely to occur in practice (I may be wrong) and in any case is not likely to perform much worse (I may be wrong again). I guess I have flogged this horse long enough though, and will merge!

PR #427, opened 2018-11-14
Ewoms pure black oil

hnil:
Pull request for discussion: This changes make it posssible to to run with a black oil fluidsystem with minimal interface.

This avoid that one has to implement interface to function not needed if one only are running black oil. It should not change other tings.

A better implementation would probably be to have a compiler switch for some of the commented lines to be able to easy use the genneral vtk output functions in the future..

### DIFF
--- a/ebos/ecloutputblackoilmodule.hh
+++ b/ebos/ecloutputblackoilmodule.hh
@@ -397,12 +397,14 @@ public:
                 Opm::Valgrind::CheckDefined(gasFormationVolumeFactor_[globalDofIdx]);
 
             }
+            /*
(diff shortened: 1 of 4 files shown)


andlaus:
you probably mean fluid *state*? I think this PR is a bad idea because the whole purpose for fluid states is to provide a common way to query thermodynamic quantities while having flexibility in how the state is stored in memory. Also, the interface is not terribly large, so implementing it is not too much of a hassle!?

andlaus:
@hnil: this PR seems to be unnecessary and inactive. I'll close it for now, but if it is strictly required for something, feel free to reopen.

PR #429, opened 2018-11-14
[bugfix][MatrixBlock] fix fallout from SparseMatrixAdapter merge for DUNE 2.4.1.

dr-robertk:
This fixes compilation issues mentioned in opm-simulators#1635.

PR #430, opened 2018-11-15
MatrixBlock: make BaseType public again

andlaus:
having it private seems to cause trouble for older compilers...

this should fix OPM/opm-simulators#1635.

@GitPaean: please merge if it helps!


andlaus:
didn't see #429. sorry for the noise.

PR #434, opened 2018-11-16
linear solver backends: merge prepareMatrix() and prepareRhs() to prepare

andlaus:
I'm not 100% sure about the performance implications, though I'm reasonably confident that the results are still correct. test well before merge!

### DIFF
--- a/ewoms/linear/parallelbasebackend.hh
+++ b/ewoms/linear/parallelbasebackend.hh
@@ -213,29 +213,19 @@ public:
     void eraseMatrix()
     { cleanup_(); }
 
-    void prepareMatrix(const SparseMatrixAdapter& M)
+    void prepare(SparseMatrixAdapter& M, Vector& b)
     {
         // make sure that the overlapping matrix and block vectors
         // have been created
         prepare_(M);
 
         // copy the interior values of the non-overlapping linear system of
-        // equations to the overlapping one. On ther border, we add up
-        // the values of all processes (using the assignAdd() methods)
+        // equations to the overlapping one.
         overlappingMatrix_->assignFromNative(M.istlMatrix());
 
         // synchronize all entries from their master processes and add entries on the
         // process border
         overlappingMatrix_->syncAdd();
-        // the entries on the border have already been added in prepareRhs()
-        overlappingb_->sync();
-    }
-
-    void prepareRhs(const SparseMatrixAdapter& M, Vector& b)
-    {
-        // make sure that the overlapping matrix and block vectors
-        // have been created
-        prepare_(M);
 
         overlappingb_->assignAddBorder(b);
 
--- a/ewoms/linear/overlappingblockvector.hh
+++ b/ewoms/linear/overlappingblockvector.hh
@@ -119,9 +119,10 @@ public:
                 (*this)[domRowIdx] = nativeBlockVector[nativeRowIdx];
         }
 
-        // add up the contents of border rows, for the remaining rows,
-        // get the values from their respective master process.
-        syncAddBorder();
+        // add up the contents of the overlapping rows. Since non-native rows are set to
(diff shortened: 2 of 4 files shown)


andlaus:
jenkins build this please

andlaus:
any news here? I think it's time to close or merge this PR...

atgeirr:
> any news here? I think it's time to close or merge this PR...

Could you please say a little bit about what you are trying to do? I.e. what was the intended use for separate methods, why do you now combine them?

andlaus:
It's been a long time since I've written the original API, so the definitive answer is lost in the mist of time. IIRC the reason why this was split was related to slightly better performance, i.e. to be able to sum up the boundary DOFs of the residual before the convergence check, do the convergence check and deal with the Jacobian  only if another iteration needs to be done.

This has turned out not to be necessary because the Jacobian and the residual are computed at the same time and the wall time needed to synchronize the Jacobian across processes has turned out to be neglible. Since the split approach is not really fool proof (which of the two methods needs to be called first? what happens if you forget to call one?), everyone is better off with a simpler interface. (this emerged during the opm-simulators linear solver transition.)

atgeirr:
> This has turned out not to be necessary because the Jacobian and the residual are computed at the same time and the wall time needed to synchronize the Jacobian across processes has turned out to be neglible. Since the split approach is not really fool proof (which of the two methods needs to be called first? what happens if you forget to call one?), everyone is better off with a simpler interface. (this emerged during the opm-simulators linear solver transition.)

Thanks for your answer! It's enough for me. However as you state initially we should verify performance first, has anyone done anything apart from Norne with this? Also, it is not clear to me if this is a change where we should expect identical results as before, or not?

andlaus:
> Also, it is not clear to me if this is a change where we should expect identical results as before, or not?

I am willing to give some warranties here: Unless some serious bug crept in, the results are identical because the difference with this PR is that now the right-hand-side vector and its Jacobian are processed in a single step while it was two before. Also, this mainly affects the MPI parallelism.

ewoms/linear/parallelbasebackend.hh
```diff
@@ -213,29 +213,19 @@ public:
...
-        // equations to the overlapping one. On ther border, we add up
-        // the values of all processes (using the assignAdd() methods)
+        // equations to the overlapping one.
         overlappingMatrix_->assignFromNative(M.istlMatrix());
 
         // synchronize all entries from their master processes and add entries on the
         // process border
         overlappingMatrix_->syncAdd();
-        // the entries on the border have already been added in prepareRhs()
-        overlappingb_->sync();
```
atgeirr: Why is the sync() call no longer needed?

andlaus: because the cross-process synchronization is taken care of on line 240. IOW, the `sync()` call was always unnecessary. probably it was a leftover from debugging the arithmetic overlap classes.

atgeirr: To be clear, do you men line 230 (240 in the current code), the assignAddBorder() call?

andlaus: yes. That said, I took this as an opportunity to look at the `assignAddBorder()` method and discovered that it was buggy for "remote" borders. This was probably the reason for the `sync()` call; b476ac905b38dd31f2bd3b5b21cb4438df78b7f1 should fix that.

andlaus: to be clear: this does not affect anything which does not use the algebraic overlap code. in particular it has no influence on the linear solver stuff from opm-simulators.

andlaus:
jenkins build this please

andlaus:
seems like the technical issues have been resolved here. @atgeirr, @totto82: mind to press green?

PR #435, opened 2018-11-29
Tasklets: Provide Additional Context For 'std::exception'

bska:
This commit adds an information-only exception handler for types derived from `std::exception`.  The handler prints `what()` in addition to the general notification about an exception in the case of exceptions propagating from within a `TaskletInterface::run()`.

This, in turn, is useful in tracking down the origin of a problem without running the process in a debugger.

### DIFF
--- a/ewoms/parallel/tasklets.hh
+++ b/ewoms/parallel/tasklets.hh
@@ -230,8 +230,11 @@ public:
                 try {
                     tasklet->run();
                 }
+                catch (const std::exception& e) {
+                    std::cerr << "ERROR: Uncaught std::exception when running tasklet: " << e.what() << ". Trying to continue.\n";
+                }
                 catch (...) {
-                    std::cerr << "ERROR: Uncaught exception when running tasklet. Trying to continue.\n";
+                    std::cerr << "ERROR: Uncaught exception (general type) when running tasklet. Trying to continue.\n";
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


bska:
jenkins build this please

blattms:
I kind of lost track where tasklets are used (still output only?), but anyway.

I am not sure whether this change is enough. It seems like flow will continue running (as the exception is caught and just an error is printed). Is this what we want in case output is not working correctly?

bska:
> I kind of lost track where tasklets are used (still output only?)

It's not entirely clear to me either.  At this time the `tasklets.hh` file is directly included only in [`vtkmultiwriter.hh`](https://github.com/OPM/ewoms/blob/118ae279dc77604583740df96f223ae15ed8af71/ewoms/io/vtkmultiwriter.hh#L36) and [`eclwriter.hh`](https://github.com/OPM/ewoms/blob/118ae279dc77604583740df96f223ae15ed8af71/ebos/eclwriter.hh#L36), but I have not tried to determine the transitive inclusion set.

> I am not sure whether this change is enough. It seems like flow will continue running (as the exception is caught and just an error is printed). Is this what we want in case output is not working correctly?

To be honest, I **think** for the ECLIPSE-compatible output case an exception should imply process termination.  On the other hand, I don't immediately see all the implications in an MPI/multiprocess context with multithreaded I/O.

PR #436, opened 2018-11-30
remove transmissibilities less then 1e-6 in the given unit system

totto82:
This removes transmissibilities less than 1e-19 in SI units to improve convergence.

### DIFF
--- a/ebos/ecltransmissibility.hh
+++ b/ebos/ecltransmissibility.hh
@@ -34,6 +34,8 @@
 #include <opm/parser/eclipse/EclipseState/Grid/GridProperties.hpp>
 #include <opm/parser/eclipse/EclipseState/Grid/FaceDir.hpp>
 #include <opm/parser/eclipse/EclipseState/Grid/TransMult.hpp>
+#include <opm/parser/eclipse/Units/Units.hpp>
+
 
 #include <opm/grid/CpGrid.hpp>
 
@@ -90,7 +92,10 @@ class EclTransmissibility
 public:
     EclTransmissibility(const Vanguard& vanguard)
         : vanguard_(vanguard)
-    {}
+    {
+        const Opm::UnitSystem& unitSystem = vanguard_.deck().getActiveUnitSystem();
+        transmissibility_threshold_  = unitSystem.parse("Transmissibility").getSIScaling() * 1e-6;
+    }
 
     /*!
      * \brief Actually compute the transmissibilty over a face as a pre-compute step.
@@ -344,6 +349,9 @@ public:
                 trans *= transMult.getRegionMultiplier(insideCartElemIdx,
                                                        outsideCartElemIdx,
                                                        faceDir);
+
+                if (trans < transmissibility_threshold_)  //remove trans less than 1e-6 in given unit
+                    trans = 0.0;
 
                 trans_[isId_(elemIdx, outsideElemIdx)] = trans;
             }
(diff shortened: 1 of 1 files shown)


totto82:
@GitPaean This potentially impacts the prediction case. I would appreciate if you could test this patch for that particular case.

GitPaean:
Thanks for the notice. I will do some tests, probably next week.  

But what is the reason that this `transmissibility_threshold_` is there in the first place? for convergence reason?

totto82:
> But what is the reason that this transmissibility_threshold_ is there in the first place? for convergence reason?

It improves the convergence and seems to be necessary to get the new norne model running. Since the threshold is very small it doesn't effect most models, but I notice that it removes some transmissibilities in the prediction case.

atgeirr:
There is a keyword MINNNCT I believe that cuts NNC (only) connections with small transmissibilities. Do you know if there is such a keyword that applies a similar threshold to all connections? If so, the feature could be triggered using that keyword, and if not should we consider adding it?

GitPaean:
The new update of the ecl and opm-common intervene with some old decks which have higher priority at the moment. I will test this PR next  week.

OPMUSER:
What about MINVALUE keyword in the EDIT section using the TRANX, TRANY and TRANZ arrays?

I have not tested this, but in the commercial simulator it states "If the MINVALUE keyword is used before an array has been input, it will have no effect on that array when it is subsequently input. MINVALUE can only operate on arrays that are already input."

So this should work in the EDIT section.

GitPaean:
I did some testing running with this PR. It changed the behavoir of the prediction in some way, but I will not say it makes the result worse (better for some wells and worse for some other wells). For prediction, if something changes earlier, then the result afterwards will change. 

If you think  the changes in this PR is the correct thing to do, I think we should merge it.  I am not aware of the rational that this threshold was introduced for.

alfbr:
This improves convergence on some important cases, and is the default threshold level specified by the input file format. Hence, I believe this should go in.

totto82:
@atgeirr Can I merge this?

blattms:
I know that I am late to the game, but anyhow:
This PR will only effect the computed transmissibilities. These seem a bit preliminary as they might either get overwritten by the transmissibilities as specified in the deck or modified via the NNC or EDITNNC keyword later on.
Maybe it would make sense to move the check and modification to a new loop over the transmissibilities at the very end of upate [ecltransmissibility.hh#L363](https://github.com/OPM/ewoms/blob/ae974b032aab89dfa31d712c1322dc925f58879c/ebos/ecltransmissibility.hh#L363)
```
}
for ( auto&& trans: trans_ ){
  if (trans < transmissibility_threshold_)  //remove trans less than 1e-6 in given unit
    trans = 0.0;
}
```
Only then we modify the final transmissibility.

atgeirr:
> This PR will only effect the computed transmissibilities. These seem a bit preliminary as they might either get overwritten by the transmissibilities as specified in the deck or modified via the NNC or EDITNNC keyword later on.

On the other hand that may be considered a feature: if you actively specify a transmissibility using NNC or EDITNNC you might want it to be used even if tiny. I have no strong opinion either way.

alfbr:
Actually, another compilcating factor came up. This PR if I understand correctly removes all transmissibilities below 1e-6, but the file format specification states that this should only be done for NNCs. For neighbour connections, you will typically find transmissibilites below 1e-6 in a simulation.

atgeirr:
> the file format specification states that this should only be done for NNCs.

That is what I understood the MINNNCT to do, but as I said, I got confused...

totto82:
> On the other hand that may be considered a feature: if you actively specify a transmissibility using NNC or EDITNNC you might want it to be used even if tiny. I have no strong opinion either way.

@alfbr Any opinion on the "correct" behavior for transmissibility specified in the edit section?

alfbr:
I believe the changes in the EDIT section should be applied before low values are filtered out, but we can double check over here.

totto82:
I can make a PR that only remove the NNCs

alfbr:
Excellent!

PR #440, opened 2018-12-07
adapt to recent changes in dune istl solvers

totto82:
I haven't investigated why Jenkins didn't complain about this earlier, but hopefully this is sufficient to make him happy again.


totto82:
jenkins build this please

totto82:
I guess @blattms was faster. I will close this.

blattms:
Maybe you should add the first change. My fix uses Scalar instead of LinearSolverScalar. I have no idea when these my differ.

PR #441, opened 2018-12-07
Do not log messages during getWriteRestartFile query on all procs.

blattms:
Previously all processes reported
"Warning: Fast restart using SAVE is not supported. Standard restart file is written instead."
in the PRT file.
Now this is done only on the master process where logging is activated.


blattms:
jenkins build this opm-common=578 please

PR #443, opened 2018-12-10
Add parameter for (dis)enable the tracer model

totto82:
The tracer model slows down the simulation time and is disabled by
default, until the performance is improved.


totto82:
jenkins build this please

atgeirr:
Fine, we'll reenable it when performance is back where it should be.

PR #445, opened 2018-12-11
Remove only nnc trans

totto82:
This PR applies the cut-off after the user specified NNCs are added.

### DIFF
--- a/ebos/ecltransmissibility.hh
+++ b/ebos/ecltransmissibility.hh
@@ -89,6 +89,9 @@ class EclTransmissibility
     typedef Dune::FieldMatrix<Scalar, dimWorld, dimWorld> DimMatrix;
     typedef Dune::FieldVector<Scalar, dimWorld> DimVector;
 
+    static const unsigned elemIdxShift = 32; // bits
+
+
 public:
     EclTransmissibility(const Vanguard& vanguard)
         : vanguard_(vanguard)
@@ -350,9 +353,6 @@ public:
                                                        outsideCartElemIdx,
                                                        faceDir);
 
-                if (trans < transmissibility_threshold_)  //remove trans less than 1e-6 in given unit
-                    trans = 0.0;
-
                 trans_[isId_(elemIdx, outsideElemIdx)] = trans;
             }
         }
@@ -360,6 +360,9 @@ public:
         // potentially overwrite and/or modify  transmissibilities based on input from deck
         updateFromEclState_();
         applyEditNNC_(elemMapper);
+
+        //remove very small non-neighbouring transmissibilities
+        removeSmallNonCartesianTransmissibilities_();
     }
 
     /*!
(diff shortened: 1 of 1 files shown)


totto82:
jenkins build this please

ebos/ecltransmissibility.hh
```diff
@@ -350,16 +353,16 @@ public:
...
                 trans_[isId_(elemIdx, outsideElemIdx)] = trans;
             }
         }
 
         // potentially overwrite and/or modify  transmissibilities based on input from deck
         updateFromEclState_();
         applyEditNNC_(elemMapper);
+
+        //remove very small non-neighbouring transmissibilities
+        removeSmallNNCTransmissibilities_();
```
atgeirr: I think it should be named "removeSmallNonCartesianTransmissibilities_()" since it can remove transmissibilities between cells that are physical neighbours across faults. We should be careful distinguishing this from "true" NNCs between cells that are *not* physical neighbours.

ebos/ecltransmissibility.hh
```diff
@@ -402,6 +405,26 @@ public:
...
+                const auto& id = trans.first;
+                const auto& elements = isIdReverse_(id);
+                int gc1 = std::min(cartMapper.cartesianIndex(elements.first), cartMapper.cartesianIndex(elements.second));
+                int gc2 = std::max(cartMapper.cartesianIndex(elements.first), cartMapper.cartesianIndex(elements.second));
+
+                // only adjust the NNCs
+                if (gc2 - gc1 == 1 || gc2 - gc1 == cartDims[0] || gc2 - gc1 == cartDims[0]*cartDims[1] )
+                    continue;
+
+                //remove trans less than 1e-6 in given unit
```
atgeirr: 1e-6 -> "the threshold (by default 1e-6 in the deck's unit system)"

ebos/ecltransmissibility.hh
```diff
@@ -607,18 +630,22 @@ private:
...
         std::uint64_t elemBIdx = std::max(elemIdx1, elemIdx2);
 
         return (elemBIdx<<elemIdxShift) + elemAIdx;
     }
 
-    std::uint64_t directionalIsId_(unsigned elemIdx1, unsigned elemIdx2) const
+    std::pair<unsigned, unsigned> isIdReverse_(const std::uint64_t& id) const
     {
-        static const unsigned elemIdxShift = 32; // bits
+        unsigned elemAIdx = id;
```
atgeirr: I had to dig a bit to check that this is indeed correct. Please add a comment:
```
    // Assigning an unsigned integer to a narrower type discards the most significant bits.
    // See "The C programming language", section A.6.2.
```

Also, while here we should make things bullet proof by ensuring we are using 32 bit numbers when expected: please replace `unsigned` -> `std::uint32_t` for `elemIdx1`, `elemIdx2` and `elemAIdx` here, in `isId_()` and in `directionalIsId_()`, as well as in the return type of `isIdReverse_()`.

totto82:
jenkins build this please

atgeirr:
Can be merged when green.

PR #446, opened 2018-12-13
Do not invalidate intensive quantites cache when DRSDT is activated

totto82:
DRSDT changes the solution between timesteps and thus violates
the assumption in fvbaselocalresidual.hh that the beginning of the time
step is the same as the end of the last time step.

If the intesive quantities are not updated teh DRSDT effect will not
be effective before iteration 1. This may couse convergence issues.
A proper fix would be to compute and store the accumulation term when
the time-step is over.

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -713,7 +713,10 @@ public:
 
         aquiferModel_.beginEpisode();
 
-        if (doInvalidate)
+        // if drsdt and drvdt is active invalidating the intesive quantities
+        // results in wrong initial storage term. (timeIndex = 1)
+        // TODO: Explicitly store the last storage term in the endTimeStep method
+        if (doInvalidate && !drsdtActive_() && !drvdtActive_())
             this->model().invalidateIntensiveQuantitiesCache(/*timeIdx=*/0);
     }
 


totto82:
jenkins build this please

ebos/eclproblem.hh
```diff
@@ -713,7 +713,10 @@ public:
 
         aquiferModel_.beginEpisode();
 
-        if (doInvalidate)
+        // if drsdt and drvdt is active invalidating the intesive quantities
+        // results in wrong initial storage term. (timeIndex = 1)
+        // TODO: Explicitly store the last storage term in the endTimeStep method
+        if (doInvalidate && !drsdtActive_() && !drvdtActive_())
```
andlaus: it's probably slightly better to set the `doInvalidate` flag instead of making the condition more complicated:

```c++
doInvalidate = doInvalidate || drsdtActive_() || drvdtActive_();
```

also, this should be done in `beginTimeStep()` not at `beginEpisode()`? finally, on an unrelated note, the update method of the `BlackOilNewtonMethod` should also respect the maximum composition change. (maybe it already does, I just skimmed over that beast.)

andlaus: come think about it, the main part of the problem is probably that the Newton update does not consider the maximum composition change. At the beginning of a time step the IQs need to be invalidated if DR[SV]DT is active, though because the derivatives w.r.t. some primary variables are different for the next time step then.

totto82: > At the beginning of a time step the IQs still need to be invalidated if DR[SV]DT is active, 

I agree, but then we need to store the accumulation term for timeStepIdx=1 before we apply the drsdt update. The accumulation term for timeStepIdx = 1should be computed with the intensive quantities based on drsdt value from the last time step and the accumulation term for timeStepIdx=0 should be computed with intensive quantities based on drsdt value from the current time step.

andlaus:
> we need to store the accumulation term for timeStepIdx=1 before we apply the drsdt update

why? I think this is already done that way if the storage cache is enabled and if it is disabled, the storage term should be correct if the solution is consistent with the IQs (which it currently is likely not).

> The accumulation term for timeStepIdx = 1should be computed with the intensive quantities based on drsdt value from the last time step

you probably mean the storage term using the Rs and Rv values of the solution at the last time step? (DR[SV]DT does only make sense if a time delta is involved.)

andlaus:
BTW: we should probably also to check the behavior of the current implementation for "saturated" fluids, i.e.,  for cells where free gas is present and the Rs value changes due to pressure.

alfbr:
A merge is greatly appreciated. This has been tested and fixes a critical bug for us.

andlaus:
the problem is that this PR does not fix the full issue. It may work in some cases, but that's basically a side effect. Be a bit patient. (if this is really urgent, you can merge it to your internal branch in the meantime.)

totto82:
I agree with @andlaus that this should not be merged. I suggest that it should be kept open until we have a proper fix.

andlaus:
#452 should fix this The Right Way (TM)? closing this pr.

PR #447, opened 2018-12-17
Changes to facilitate adjoint.

hnil:
This is a new try on fasilitate adjoint in ewoms. The pull request is messy but I hope this will be fixable.
It again introduce focusTimeIdx. At the moment it only applies to the storage term (some errors apperears if not related to well sources). It also have some changes related to time stepping. Adjoint steps backward at steps of the forward simulator which require a bit other logic.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -5,25 +5,24 @@ Changes in eWoms 2018.10
   - Support for positional command line parameters.
   - Possibility to hide parameters in the usage message. This is
     useful if parameters are there but not actually used, as is the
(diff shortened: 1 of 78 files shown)


hnil:
jenkins build this please

andlaus:
What's the status of this? It currently has some merge conflicts with master and it should be rebased/squashed on top of the current master to make it easier to handle the imminent move of ebos to opm-simulators...

andlaus:
looks like the bulk (all?) of this PR needs to move to opm-simulators because the `ebos` changed its home module. closing for now.

PR #450, opened 2018-12-20
Add support for the Multz option ALL in PINCH

totto82:
This needs some more testing 

- [x] what happens when different means of multiplying the transmissibility in the z direction is considered. 
- [x] test for MULTZ-

### DIFF
--- a/ebos/ecltransmissibility.hh
+++ b/ebos/ecltransmissibility.hh
@@ -129,6 +129,7 @@ public:
         const auto& cartMapper = vanguard_.cartesianIndexMapper();
         const auto& eclState = vanguard_.eclState();
         const auto& eclGrid = eclState.getInputGrid();
+        const auto& cartDims = cartMapper.cartesianDimensions();
         auto& transMult = eclState.getTransMult();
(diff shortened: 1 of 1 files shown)


ebos/ecltransmissibility.hh
```diff
@@ -323,7 +324,32 @@ public:
 
                 // apply the full face transmissibility multipliers
                 // for the inside ...
-                applyMultipliers_(trans, insideFaceIdx, insideCartElemIdx, transMult);
+
+                // The MULTZ needs special case if the option is ALL
```
andlaus: could the new code moved to `applyMultipliers_()`

andlaus: yes, that would be better. this method is quite a bit too long as it is, IMO.

totto82:
jenkins build this please

totto82:
This is ready from my side. @andlaus Can this be merged?

alfbr:
This has been tested and confirmed to solve the problem on our side.

andlaus:
looks good. thanks for the patch.

PR #452, opened 2018-12-20
make it possible to explicitly compute the storage term of the previous time step

andlaus:
some weird hacks (hello, DR[SV]DT) cause a change of the storage term in the first Newton-Raphson iteration compared to the solution of the previous time level. In order to use the correct values, one thus must explicitly recompute the storage term for the previous time step instead of just reusing the result of the first Newton-Raphson iteration of the current time step.

This (hopefully) fixes the root issue which is papered over  by #446. I've tested this on Norne and did not notice any performance regression. decks which have DRSDT or DRVDT active at some point may yield slightly different results, so a data update might be required.

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -744,6 +744,16 @@ public:
     }
 
     /*!
+     * \brief Return if the storage term of the first iteration is identical to the storage
+     *        term for the solution of the previous time step.
+     *
+     * For quite technical reasons, the storage term cannot be recycled if either DRSDT
+     * or DRVDT are active in ebos.
+     */
+    bool recycleFirstIterationStorage() const
+    { return !drsdtActive_() && !drvdtActive_(); }
+
+    /*!
      * \brief Called by the simulator before each Newton-Raphson iteration.
      */
     void beginIteration()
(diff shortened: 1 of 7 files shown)


andlaus:
jenkins build this please

andlaus:
BTW: a better name for the new callback method to the problem would be greatly appreciated.

andlaus:
jenkins build this please

totto82:
Thanks. This should fix the DRSDT issue. I would not be able to test this before January. Could you test this on the open Norne model where you have disabled the VAPPARS keyword and compare it with the #446 hack. They should give the same answer.

totto82:
I have tested this and it behaves as expected. I will merge this tomorrow at noon if nobody objects.

andlaus:
okay; this is now a fait acompli, but I did check it with de-VAPPARized Norne as you suggested. most wells look like this: 
a
but there are others that look like this:
b

notice the purple kink for PR446 at about 1500 days? I'm not 100% sure, but for 98.723% that's because the simulator breaks down at report step 121 and when it tries again with a chopped the time step, the intensive quantities are **way** off.

(also, note that there is no difference between the curve for master and the others. that's because that deck  sets DRSDT to 0. if it is set to something else, the curves are different, but they still  match.)

@alfbr: I hope you agree that given the above, it was a good idea not to rush merging #446.

andlaus:
> #446 was a one-liner fix that could easily be reverted once a proper fix was available.

but it made violated the design of the linearization machinery considerably and created a subtle bug. if that PR was merged, I'm pretty sure that nobody would have invested the time for a proper fix, and when the bug  finally bit it's anyone's guess who would have bitten, and who wanted everyone to drop everything to fix it yesterday and who would spend the next week slamming his head against the nearest wall.

alfbr:
>  I'm pretty sure that nobody would have invested the time for a proper fix

I don't understand how you can reach that conclusion. It was you who invested the time.

andlaus:
> It was you who invested the time.

I wouldn't have if #446 was merged. At least not if I wasn't explicitly asked, and I doubt that I would have.

andlaus:
Also, I was told that you have internal branches were additional patches applied. Why don't you use these for your internal milestones? (the public master branches serve a different purpose. Consider if every stakeholder would do what you proposed if they have some internal milestones coming up.)

alfbr:
It is impossible for me to know what you have been told, and who told you. I can however ensure you that we have no internal branches and no internal patches. Please refrain form making such statements, as it undermines what we try to achieve. A completely open development. What is true, is that we currently cannot wait for the next release until we roll out github contributed patches to our users. This process should probably be improved, but I don't think this thread is the right place to discuss that.

PR #453, opened 2019-01-07
Add ErrorGuard argument when parsing

joakim-hove:
Downstream of: https://github.com/OPM/opm-common/pull/602

### DIFF
--- a/ebos/eclbasevanguard.hh
+++ b/ebos/eclbasevanguard.hh
@@ -33,6 +33,7 @@
 
 #include <opm/parser/eclipse/Parser/Parser.hpp>
 #include <opm/parser/eclipse/Parser/ParseContext.hpp>
+#include <opm/parser/eclipse/Parser/ErrorGuard.hpp>
 #include <opm/parser/eclipse/Deck/Deck.hpp>
 #include <opm/parser/eclipse/EclipseState/checkDeck.hpp>
 #include <opm/parser/eclipse/EclipseState/EclipseState.hpp>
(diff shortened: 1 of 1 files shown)


andlaus:
Parser errors should be handled before the simulator tries to set up any data structures using the faulty data.

With the approach itself I don't like the fact that the simulator is shot into the head by the parser by default if there was something which the latter considers to be a problem. the shooting-into-the-head business should be done by the simulator itself, i.e., it should commit suicide if instead of being placed front of a firing squad. 

That said, it you stubbornly insist on merging this PR in something close to its current form, I insist that the route which you mentioned in your long comment at the bottom of the patch must be taken with the "handle errors" part *after* clearing the error log and essentially coming down to throwing an exception.

andlaus:
I just noticed: Isn't the name `Opm::ErrorGuard` much too generic for a class that is very specific to the parsere? There already is am `Ewoms::GenericGuard` class which is really what it says.  (a guard class which you hand up a cleanup function that's called in the destructor.)

`Opm::ParserErrorLog` would be a better name, IMO.

joakim-hove:
> I insist that the route which you mentioned in your long comment at the bottom of the patch must be taken with the "handle errors" part after clearing the error log and essentially coming down to throwing an exception.

Done

joakim-hove:
> Parser errors should be handled before the simulator tries to set up any data structures using the faulty data.

I don't really understand what you mean here. There are four (large) datastructures whuch are assembled by code loosely defined as *the parser* - those are: `Deck, EclipseState, Schedule` and `SummaryConfig`. The `ErrorGuard` is used to assemble errors from all of these four - calling scope - i.e. the simulator could chose to chose use separate `ErrorGuard` arguments for the four objects, but the purpose of this has been to facilitate proceeding with the input treatment as *far as possible* for before the user is presented with a fatal error message and the whole thing stops, to achieve that it makes sense to use the same `ErrorGuard` instance for all four objects, before status is made up.

andlaus:
> I don't really understand what you mean here

simple: trying to set up simulator objects that are derived from the parser data  (Grid, transmissibilities, etc.) when the latter is potentially garbage does not make any sense and might even lead to exceptions thrown during the set-up of these objects. with this patch in its present form these exceptions would lead to a head-shot. I'm not 100% sure if this happens, but it looks like it: e.g., the call to `filterConnections_()` looks suspicious.

joakim-hove:
OK - I have moved the error handler block a couple of lines up, I agree that is a good change.

PR #454, opened 2019-01-07
Strict parsing

joakim-hove:
Add a command-line option `--strict-parsing`. If the `--strict-parsing` option is enabled we all `ParseContext` error modes are treated with the mode `DELAYED_EXIT1` - this implies that:

1. All recognized error modes are treated as fatal.
2. Parsing, and construction of `EclipseState++` objects continues all the way (unless some other failure pops up); if there are errors - all errors and warnings are dumped on `stderr` and the application stops.

The first commit comes from: #453 - which must be merged first.

**Update:** This PR is not really complete, because the `Deck` and `EclipseState` objects are assigned from external scope, so the `--strict-parsing` option must be parsed/interpreted also in opm-simulators.

### DIFF
--- a/ebos/eclbasevanguard.hh
+++ b/ebos/eclbasevanguard.hh
@@ -65,6 +65,7 @@ NEW_PROP_TAG(Scalar);
 NEW_PROP_TAG(EclDeckFileName);
 NEW_PROP_TAG(OutputDir);
 NEW_PROP_TAG(EnableOpmRstFile);
+NEW_PROP_TAG(EclStrictParsing);
 NEW_PROP_TAG(EclOutputInterval);
 NEW_PROP_TAG(IgnoreKeywords);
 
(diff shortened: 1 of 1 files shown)


ebos/eclbasevanguard.hh
```diff
@@ -64,13 +65,15 @@ NEW_PROP_TAG(Scalar);
...
 NEW_PROP_TAG(EnableOpmRstFile);
+NEW_PROP_TAG(StrictParsing);
 NEW_PROP_TAG(EclOutputInterval);
 NEW_PROP_TAG(IgnoreKeywords);
 
 SET_STRING_PROP(EclBaseVanguard, IgnoreKeywords, "");
 SET_STRING_PROP(EclBaseVanguard, EclDeckFileName, "");
 SET_INT_PROP(EclBaseVanguard, EclOutputInterval, -1); // use the deck-provided value
 SET_BOOL_PROP(EclBaseVanguard, EnableOpmRstFile, true);
+SET_BOOL_PROP(EclBaseVanguard, StrictParsing, false);
```
andlaus: this should better be called `EclEnableStrictParsing`

joakim-hove: That *is* a mothfull - how about: `EnableStrictParsing`or `EclStrictParsing`?

andlaus: IMO both are better than the "naked" `StrictParsing`. It's a close call, but I slightly prefer `EclStrictParsing`.

joakim-hove:
OK - I have updated the parameter name to `EclStrictParsing`; but I have also made some efforts to instantiate *all* the objects which use/require the `ErrorGuard` object from `flow.cpp` - and then this will not be very relevant anyways - for flow at least.

Will merge if Jenkins goes through.

joakim-hove:
jenkins build this please

joakim-hove:
jenkins build this please

PR #455, opened 2019-01-08
ebos: fix explicit initialization for twophase cases

andlaus:
since OPM/opm-material#310 has been merged, data for deactivated phases is not stored anymore and can thus no longer be accessed/set.

this fixes OPM/opm-simulators#1686


totto82:
Thanks. This work for the simplified test provided by @andrescodas. Please report back if it does not work for the full case

PR #456, opened 2019-01-09
fix some masochistic compiler warnings for the GCC 9 pre-release

andlaus:
the flags which I used are
```
-pedantic \
-Wall \
-Wextra \
-Wformat-nonliteral \
-Wcast-align
-Wpointer-arith \
-Wmissing-declarations \
-Wcast-qual \
-Wshadow
-Wwrite-strings \
-Wchar-subscripts \
-Wredundant-decls \
-fstrict-overflow \
-O3 \
-march=native \
-DNDEBUG=1
```

note that some heavy filtering is not the worst idea because DUNE is far from not emiting any warnings with these flags.

Also, there were some pesky warnings in test_ecl_output which I don't know how to fix:

```
tests/test_ecl_output.cc:218:73: warning: missing initializer for member ‘Opm::data::Connection::effective_Kh’ [-Wmissing-field-initializers]
```

### DIFF
--- a/ebos/equil/initstateequil.hh
+++ b/ebos/equil/initstateequil.hh
@@ -713,9 +713,9 @@ phaseSaturations(const Grid& grid,
                                          /*storeSaturation=*/true,
                                          /*storeDensity=*/false,
                                          /*storeViscosity=*/false,
-                                         /*storeEnthalpy=*/false> SatOnlyFluidState;
+                                         /*storeEnthalpy=*/false> MySatOnlyFluidState;
 
-    SatOnlyFluidState fluidState;
+    MySatOnlyFluidState fluidState;
     typedef typename MaterialLawManager::MaterialLaw MaterialLaw;
 
     const bool water = FluidSystem::phaseIsActive(FluidSystem::waterPhaseIdx);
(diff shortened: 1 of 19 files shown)


andlaus:
jenkins build this please

ebos/equil/initstateequil.hh
```diff
@@ -713,9 +713,9 @@ phaseSaturations(const Grid& grid,
                                          /*storeSaturation=*/true,
                                          /*storeDensity=*/false,
                                          /*storeViscosity=*/false,
-                                         /*storeEnthalpy=*/false> SatOnlyFluidState;
+                                         /*storeEnthalpy=*/false> MySatOnlyFluidState;
 
-    SatOnlyFluidState fluidState;
+    MySatOnlyFluidState fluidState;
     typedef typename MaterialLawManager::MaterialLaw MaterialLaw;
```
totto82: What is the reasoning here?

andlaus: naming conflict with https://github.com/OPM/ewoms/blob/master/ebos/equil/equilibrationhelpers.hh#L97-L108

until the EQUIL code gets seriously refactored, this is the best way to chicken out, IMO.

totto82: Ok

PR #457, opened 2019-01-11
Add Schedule and SummaryConfig arguments to setExternalDeck()

joakim-hove:
The parsing process creates four top level objects: `Deck, EclipseState, Schedule` and `SummaryConfig`. With the advent of the `ErrorGuard` class it makes sense to create all four objects with the same `ErrorGuard`instance - to ensure that the maximum amount of error messages are presented to the user, before we jump ship.

With this PR - and the accompanying in opm-simulators, all four objects are created in `flow.cpp` and then all are set in the ewoms model using `setExternalDeck()`.

Depends on: https://github.com/OPM/opm-simulators/pull/1704

~~Should merge: #454 first~~

### DIFF
--- a/ebos/eclbasevanguard.hh
+++ b/ebos/eclbasevanguard.hh
@@ -158,10 +158,12 @@ public:
      * management of these two objects, i.e., they are not allowed to be deleted as long
      * as the simulator vanguard object is alive.
      */
-    static void setExternalDeck(Opm::Deck* deck, Opm::EclipseState* eclState)
+    static void setExternalDeck(Opm::Deck* deck, Opm::EclipseState* eclState, Opm::Schedule * schedule, Opm::SummaryConfig * summaryConfig)
     {
         externalDeck_ = deck;
         externalEclState_ = eclState;
+        externalEclSchedule_ = schedule;
+        externalEclSummaryConfig_ = summaryConfig;
     }
 
     /*!


andlaus:
you should use a separate `setExternal` method for each of these objects. there already are `setExternalSchedule()` and `setExternalSummaryConfig()`...

joakim-hove:
> you should use a separate setExternal method for each of these objects. there already are setExternalSchedule() and setExternalSummaryConfig()...

OK .... but could I then alternative suggest:

1. `setExternalDeck()` is renamed to e.g. `setExternalInput()?` and takes *all four*
2. The `setExternalSchedule` and `setExternalSummaryConfig` are removed.

?

andlaus:
I strongly want to avoid parameter proliferation for this method. IMO, all of these objects should ideally be provided by the eclipse state object anyway (if they are not, well, does `EclipseState` it does not really represent the ECL state), of if it is not and has not been set explicitly, they should be created internally based on the objects which were provided externally.

joakim-hove:
> I strongly want to avoid parameter proliferation for this method.

Well; there are currently four parameters/members; this is PR does not imply any change of that. Two of them are currently passed in from flow, and two of them are instantiated internally in ewoms. The reason I would very much like to create them all in the same place is that that will facilitate the use of one common `ErrorGuard` argument for all four objects, and thereby the maximum amount of error information passed to the user before we take a dive - in the case of errors. 

> all of these objects should ideally be provided by the eclipse state object anyway

Well - that is just not the case. The `SummaryConfig` object is an ugly exception - that I agree; but the split between `EclipseState` and `Schedule` does in my opinion make sense - the `EclipseState` object contains all *static* information, and the `Schedule` object contains the dynamic properties. 

I feel your points are slightly on the side of my alternative suggestion:

> OK .... but could I then alternatively suggest:

    1. setExternalDeck() is renamed to e.g. setExternalInput()? and takes all four
    2. The setExternalSchedule and setExternalSummaryConfig are removed.

What do you think of that?

joakim-hove:
Ping?

andlaus:
> Well; there are currently four parameters/members

two: the deck and the eclState. Note that while I won't accept new parameters for `setExternalDeck()` if these objects are not unambiguously part of "the deck", being able to set more high-level objects using separate methods is fine. you can create these objects in the same place using the same `ErrorGuard` and then call all the methods for the vanguard separately; there's no need to  introduce the parameter proliferation disease in ebos for this. BTW: why does the deck not provide the ErrorGuard and ParseContext objects used to create it?

> Well - that is just not the case. [...]

I think from a user's point of view (in terms of "developer who uses this") it is not really clear what counts as a static and what as a dynamic parameter:  If I'm not mistaken, even the grid properties can in principle be modified in the SCHEDULE section?

but let's agree to disagree on this.

> What do you think of that?

I don't like anything which will potentially result in a method that takes 15 random parameters after a few iterations, so I don't like it better than adding these parameters to `setExternalDeck()` ;)

joakim-hove:
> two: the deck and the eclState. 

OK - I misunderstood; I though of the class which setter works on - yes the function increased from two to four arguments in my suggestion.

> you can create these objects in the same place using the same ErrorGuard and then call all the methods for the vanguard separately;

I will do that

PR #458, opened 2019-01-11
Make runtest.sh also work on BSD/macOS.

atgeirr:
In particular:
 - "wc -w" will have spaces in the output on BSD,
 - "python" defaults to python 3, so we must explicitly invoke python2.

@andlaus I have not been able to verify that my changes also work on Linux, don't think Jenkins checks it? Or maybe it does?

### DIFF
--- a/bin/runtest.sh
+++ b/bin/runtest.sh
@@ -21,7 +21,7 @@ validateResults() {
 
     for REFERENCE_RESULT in ${MY_DIR}/../tests/referencesolutions/$SIM_NAME*; do
         echo "Comparing with \"$REFERENCE_RESULT\"... "
-        if python "${MY_DIR}/fuzzycomparevtu.py" "$REFERENCE_RESULT" "$OUTPUT_FILE"; then
+        if python2 "${MY_DIR}/fuzzycomparevtu.py" "$REFERENCE_RESULT" "$OUTPUT_FILE"; then
             # SUCCESS!!!!!!
             echo "Result file '$OUTPUT_FILE' and reference '$REFERENCE_RESULT' are identical" 
             return 0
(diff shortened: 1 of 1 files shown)


atgeirr:
There is one more thing causing trouble when running the tests, but it is minor. I get the following
```
      Start 10: finger_immiscible_ecfv_adaptive
Could not find executable skip_test_dummy
Looked in the following places:
skip_test_dummy
skip_test_dummy
Release/skip_test_dummy
Release/skip_test_dummy
Debug/skip_test_dummy
Debug/skip_test_dummy
MinSizeRel/skip_test_dummy
MinSizeRel/skip_test_dummy
RelWithDebInfo/skip_test_dummy
RelWithDebInfo/skip_test_dummy
Deployment/skip_test_dummy
Deployment/skip_test_dummy
Development/skip_test_dummy
Development/skip_test_dummy
Unable to find executable: skip_test_dummy
10/58 Test #10: finger_immiscible_ecfv_adaptive .....***Not Run   0.00 sec
```

(and some more similar)

This is caused by the following in OpmSatellites.cmake:
```
    # the following causes the test to appear as 'skipped' in the
    # CDash dashboard. it this is removed, the test is just silently
    # ignored.
    if (NOT CURTEST_ONLY_COMPILE AND ADD_DISABLED_CTESTS)
      add_test(${_FANCY} skip_test_dummy)
    endif()
```

I find no reference to skip_test_dummy on the ENTIRE INTERNET other than this place. So is this a part of the build system we have silently lost?

andlaus:
everything is probably as it is supposed to be: `skip_test_dummy` is an -- err -- non-existing dummy target to tell ctest that this test ought to be skipped. with newer versions this can be done properly using a special exit code of the test driver, but for the ctest version that was current when this macro was written (2.6?), I could not find another way to do this. The need for backward compatibility and inertia has then kept it in...

andlaus:
if jenkins is happy, linux is golden ;)

andlaus:
jenkins build this please

andlaus:
it is because of the `-perm` parameter for find:

```
find: invalid mode ‘+0111’
```

the man page of the GNU version of that tool says:

```
       -perm +mode
              This is no longer supported (and has been deprecated since 2005).  Use -perm /mode instead.
```

`-executable` does not work on BSDs?

andlaus:
(I just verified that '-mode /0111` works fine. does this work for BSD find?)

atgeirr:
> (I just verified that `-mode /0111` works fine. does this work for BSD find?)

No, it does not... (assume you mean `-perm /0111`). But `-perm -0111` works, I think it should be ok, and will revise the PR. What a mess! This is why we should never use long shell scripts...

atgeirr:
jenkins build this please

andlaus:
with jenkins being happy, there's nothing which prevents merging this. thanks.

PR #459, opened 2019-01-11
WIP: Remove strict-parsing option

joakim-hove:
Seems this should be in flow/opm-simulators: https://github.com/OPM/opm-simulators/pull/1704

### DIFF
--- a/ebos/eclbasevanguard.hh
+++ b/ebos/eclbasevanguard.hh
@@ -65,7 +65,6 @@ NEW_PROP_TAG(Scalar);
 NEW_PROP_TAG(EclDeckFileName);
 NEW_PROP_TAG(OutputDir);
 NEW_PROP_TAG(EnableOpmRstFile);
-NEW_PROP_TAG(EclStrictParsing);
 NEW_PROP_TAG(EclOutputInterval);
 NEW_PROP_TAG(IgnoreKeywords);
 
@@ -73,7 +72,6 @@ SET_STRING_PROP(EclBaseVanguard, IgnoreKeywords, "");
 SET_STRING_PROP(EclBaseVanguard, EclDeckFileName, "");
 SET_INT_PROP(EclBaseVanguard, EclOutputInterval, -1); // use the deck-provided value
 SET_BOOL_PROP(EclBaseVanguard, EnableOpmRstFile, true);
-SET_BOOL_PROP(EclBaseVanguard, EclStrictParsing, false);
 
 END_PROPERTIES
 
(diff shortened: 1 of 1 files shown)


andlaus:
well, the parameter is still there. you can also access it in the `flow` context...

joakim-hove:
> well, the parameter is still there. you can also access it in the flow context...

OK - whatever; I don't really understand this. Wheh I combined `ewoms:master` with https://github.com/OPM/opm-simulators/pull/1704 I got the message:
```
terminate called after throwing an instance of 'std::logic_error'
  what():  Parameter EclStrictParsing registered twice with non-matching characteristics.
Aborted (core dumped)
```
so I concluded that it should only be in `flow` - but I appreciate guidance on this.

andlaus:
yes, must then remove the `EWOMS_REGISTER_PARAM()` call for `EclStrictParsing` from opm-simulators.

joakim-hove:
> yes, must then remove the EWOMS_REGISTER_PARAM() call for EclStrictParsing from opm-simulators.

OK - but will it then be accessable from opm-simulators; specifically before the parsing has started?

andlaus:
(registering the same parameter twice with different descriptions or types would lead to inconsistencies.)

andlaus:
> OK - but will it then be accessable from opm-simulators; specifically before the parsing has started?

I think so.

andlaus:
actually, I'm willing to bet a beer on this because the file name of the deck to be parsed is determined via the parameter system.

joakim-hove:
> I think so.

I hope you are right - but I think you are wrong.

> I'm willing to bet a beer on this because the file name of the deck to be parsed is determined via the parameter system.

Yes - but that is some uber-hack to avoid a Catch22 - I tried to hook into the same hack, but ....

joakim-hove:
I give up for now - will pick it up again. I have really tried, but getting the parameter system to adhere to my whims is not exactly walk in the park.

What I want to achieve is that `flow` supports an option - like `--ecl-strict-parsing`[1] and that I can get access to that `bool` setting [here](https://github.com/OPM/opm-simulators/pull/1704/files#diff-fa1067ba914130bd73d02fe3a49a4ad0R186) - might be walk in the park for you?

[1]: Preferably without having to add a `=true` - but that is not essential.

andlaus:
I suppose that programming a reservoir simulator is not for the faint of heart. merge joakim-hove/opm-simulators#3 and you are golden (be aware that is a PR for your opm-simulators PR).

PR #460, opened 2019-01-12
updating more quantities when restart

GitPaean:
lastRs_, lastRv_, maxDRs_, maxDRv_

to get better restart.

Trying to fix the restart issue related to SPE1CASE1 reported in https://github.com/OPM/opm-simulators/issues/1699

This PR is done with @hnil and @bska .

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -1868,12 +1868,26 @@ private:
             elemFluidState.setPvtRegionIndex(pvtRegionIndex(elemIdx));
             eclWriter_->eclOutputModule().initHysteresisParams(this->simulator(), elemIdx);
             eclWriter_->eclOutputModule().assignToFluidState(elemFluidState, elemIdx);
+            lastRs_[elemIdx] = elemFluidState.Rs();
+            lastRv_[elemIdx] = elemFluidState.Rv();
             if (enableSolvent)
                  solventSaturation_[elemIdx] = eclWriter_->eclOutputModule().getSolventSaturation(elemIdx);
             if (enablePolymer)
                  polymerConcentration_[elemIdx] = eclWriter_->eclOutputModule().getPolymerConcentration(elemIdx);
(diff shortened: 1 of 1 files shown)


ebos/eclproblem.hh
```diff
@@ -1868,13 +1868,27 @@ private:
             elemFluidState.setPvtRegionIndex(pvtRegionIndex(elemIdx));
             eclWriter_->eclOutputModule().initHysteresisParams(this->simulator(), elemIdx);
             eclWriter_->eclOutputModule().assignToFluidState(elemFluidState, elemIdx);
+            lastRs_[elemIdx] = elemFluidState.Rs();
+            lastRv_[elemIdx] = elemFluidState.Rv();
```
andlaus: in principle, these should not be necessary, because `begin{Episode,Timestep,Iteration}()` are called before the the real simulation re-starts.  It does not hurt, though.

ebos/eclproblem.hh
```diff
@@ -1868,13 +1868,27 @@ private:
...
             if (enableSolvent)
                  solventSaturation_[elemIdx] = eclWriter_->eclOutputModule().getSolventSaturation(elemIdx);
             if (enablePolymer)
                  polymerConcentration_[elemIdx] = eclWriter_->eclOutputModule().getPolymerConcentration(elemIdx);
             // if we need to restart for polymer molecular weight simulation, we need to add related here
         }
 
+        const int epsiodeIdx = this->simulator().episodeIndex();
```
andlaus: same here (?)

ebos/eclproblem.hh
andlaus: > lastRs_ is updated to be wrong value in the function updateCompositionChangeLimits_().

Ah, I see. the problem are the regions where oil and gas are present and where Rs != RsMax because of DRSDT.

> Are you sure the begin{Episode,Timestep,Iteration}() needs to be called before applyInitialSoltuion()?

no, I'm sure that it needs to be called after thereafter ;).

PR #461, opened 2019-01-15
newton method: call prepare() after preSolve()

andlaus:
because preSolve() may modify the residual and its Jacobian and the linear solver would only "see" the unmodified versions otherwise.

this fixes ebos for SPE1 and SPE9_CP, but ECL output still does not work. in that case https://github.com/OPM/opm-common/blob/master/src/opm/output/eclipse/AggregateConnectionData.cpp#L327-L329 triggers and the exception message is not at all clear to me:

```
ERROR: Uncaught std::exception when running tasklet: Inconsistent number of open connections I in vector<Opm::data::Connection*> (0) in Well PROD. Trying to continue.
```` 

this code was added by opm/opm-common#549 and is so deeply burried that I gave up my attempts at fixing this after a few hours. any hints how to fix this @jalvestad, @joakim-hove, @totto82 or @GitPaean?


andlaus:
jenkins build this please

bska:
> triggers and the exception message is not at all clear to me:

If that condition triggers it means that the `well->getActiveConnections()` has more open connections than there are connection results in [`data::Well::connections`](https://github.com/OPM/opm-common/blob/dc0ccc010ae41a73dd1094a688581c0e85fe5fd5/opm/output/data/Wells.hpp#L155) object created in the well-state's `report()` function.  The root cause is likely to be elsewhere.

GitPaean:
One places can remove some connections is https://github.com/OPM/ewoms/pull/283 . while the SPE1 and SPE9 cases should not have this problem.

bska:
> yes, that is my problem: where?

You need to look at the component that creates the [`RestartValue::wells`](https://github.com/OPM/opm-common/blob/dc0ccc010ae41a73dd1094a688581c0e85fe5fd5/opm/output/eclipse/RestartValue.hpp#L65) object&mdash;and in particular the `connections` subobject&mdash;that gets passed into [`EclipseIO::writeTimeStep()`](https://github.com/OPM/opm-common/blob/dc0ccc010ae41a73dd1094a688581c0e85fe5fd5/opm/output/eclipse/EclipseIO.hpp#L173-L180).  If the `data::Wells::connections` object legitimately holds fewer elements than there are active connections, then the (connection part of the) restart code must be completely rewritten.

andlaus:
okay, the issue with parallel VCFV turned out to be quite a bit more involved than anticipated: Since there are DOFs on the process boundary, the residual needs to be synchronized between processes before the error is calculated, and the model's `preSolve_()` method needs the final residual yet must to be called before the linear solver is started up. Since the synchronization of the Jacobian is next-to-impossible to do using DUNE grid communication, I changed the linear solver API back to something resembling the original one, but (hopefully) with clearer semantics:https://github.com/OPM/ewoms/pull/461/commits/7538199223b683201dfd4f96b781ec069ac8f54c

this now implies a small cleanup PR for opm-simulators. note that this cleanup is technically not required in the sense that `flow` compiles and works just fine without it, but not having it would prevent to use the standard eWoms linear solvers.

I also had another look at fixing the output issue. After looking at that code for half an hour, I got dizzy and gave up again. @bska: what is the simulator-provided data that needs to be changed to avoid this exception? (I can change it, but for this I need to know what causes the problem.)

andlaus:
jenkins build this opm-simulators=1712 please

andlaus:
okay, now that @totto82 has fixed the output (thanks!), I've added a regression test for plain `ebos` using SPE1CASE1.

@akva2: Some questions/remarks regarding the `compareECL` utillity surfaced during that excercise:
 - is there a way to compare EGRID files using `compareECL`?
 - it seems like `compareECL` is a bit inconsistent on the file extensions: e.g. they seem like they must be specified for the UNSMRY files, but they must be omitted when comparing INIT files.
 - `compareECL` segfaults for the restart files produced by plain ebos, yet the files should be okay because ResInsight can load them just fine.

andlaus:
jenkins build this opm-simulators=1712 please

andlaus:
jenkins build this opm-simulators=1712 please

andlaus:
ok. since this fixes a seriously annoying regression and jenkins is all green, I'll take the fire for self-merging this if jenkins is also happy with this stand-alone.

@atgeirr: please merge OPM/opm-simulators#1712

akva2:
okay, looking at the code in here, you meant restart files, not summary files. you used -t RST when you want -t UNRST. it's a bit messy, because RST and UNRST needs to be handled differently by underlying library (libecl).

andlaus:
>> is there a way to compare EGRID files using `compareECL`?
>
> not directly, tests are run as part of "normal" tests. see the comparator.gridCompare calls in the compareECL.cpp

i.e., the EGRID file is used anyway? if yes, I'm happy with the current state of affairs.

> okay, looking at the code in here, you meant restart files, not summary files. you used -t RST when you want -t UNRST. it's a bit messy, because RST and UNRST needs to be handled differently by underlying library (libecl).

okay, will open a PR to fix this. thanks for the heads-up.

andlaus:
cf #464

PR #462, opened 2019-01-18
Add trivial connection output to silence the output code

totto82:
In this PR only trivial data is added.


andlaus:
thanks! I tested this locally and it works as advertised. merging.

PR #463, opened 2019-01-21
convert FIP values to LAB scale

totto82:
Add unit conversion of FIP output in case of LAB unit system.

### DIFF
--- a/ebos/ecloutputblackoilmodule.hh
+++ b/ebos/ecloutputblackoilmodule.hh
@@ -1256,6 +1256,17 @@ private:
             fip[FipDataType::GasInPlaceInGasPhase] = Opm::unit::convert::to(fip[FipDataType::GasInPlaceInGasPhase], 1000*Opm::unit::cubic(Opm::unit::feet));
             fip[FipDataType::PoreVolume] = Opm::unit::convert::to(fip[FipDataType::PoreVolume], Opm::unit::stb);
         }
+        else if (units.getType() == Opm::UnitSystem::UnitType::UNIT_TYPE_LAB) {
+            Scalar scc = Opm::unit::cubic(Opm::prefix::centi * Opm::unit::meter); //standard cubic cm.
+            fip[FipDataType::WaterInPlace] = Opm::unit::convert::to(fip[FipDataType::WaterInPlace], scc);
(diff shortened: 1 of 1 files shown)


totto82:
jenkins build this please

andlaus:
Thanks for digging this up. It's kind of annoying that something like this needs to be done manually on such a high level of abstraction (yes, this is -- in my opinion -- another case of my favorite kind of issue: a layering violation), but the patch is fine and it does regress the status quo. I'll merge once jenkins is happy.

ebos/ecloutputblackoilmodule.hh
```diff
@@ -1256,6 +1256,17 @@ private:
...
         }
+        else if (units.getType() == Opm::UnitSystem::UnitType::UNIT_TYPE_LAB) {
+            Scalar scc = Opm::unit::cubic(Opm::prefix::centi * Opm::unit::meter); //standard cubic cm.
+            fip[FipDataType::WaterInPlace] = Opm::unit::convert::to(fip[FipDataType::WaterInPlace], scc);
+            fip[FipDataType::OilInPlace] = Opm::unit::convert::to(fip[FipDataType::OilInPlace], scc);
+            fip[FipDataType::OilInPlaceInLiquidPhase] = Opm::unit::convert::to(fip[FipDataType::OilInPlaceInLiquidPhase], scc);
+            fip[FipDataType::OilInPlaceInGasPhase] = Opm::unit::convert::to(fip[FipDataType::OilInPlaceInGasPhase], scc);
+            fip[FipDataType::GasInPlace] = Opm::unit::convert::to(fip[FipDataType::GasInPlace], scc);
+            fip[FipDataType::GasInPlaceInLiquidPhase] = Opm::unit::convert::to(fip[FipDataType::GasInPlaceInLiquidPhase], scc);
+            fip[FipDataType::GasInPlaceInGasPhase] = Opm::unit::convert::to(fip[FipDataType::GasInPlaceInGasPhase], scc);
```
joakim-hove: Why is this needed - the output layer converts from SI to whatever?

totto82: It is used for the .PRT .DBG output.

joakim-hove: OK - I guess the "proper" fix is to do the unit conversion in the PRT output routine, but that is of course a bigger change. I am not familiar with the PRT output.

PR #465, opened 2019-01-22
each phase needs to be above certain value to be treated to be present

GitPaean:
it helps to recover some RESTART running based on single precision format.

Because saturation values are used to determine saturated/unsaturated region distribution, with this patch, we are more likely to get more continuous region distribution, numerically, more continuous primary variable distribution. 

Physically, it does not matter, while numerically, the linear system is easier to solve.


GitPaean:
jenkins build this please

GitPaean:
jenkins build this please

GitPaean:
I added the re-scaling of the saturations to make sure the sum  is 1.0.  To avoid adding too much length to the original function, I put this part to be a separate function.

GitPaean:
jenkins build this please

atgeirr:
As the requested changes were made I think this should be merged. Also, it only affects (and improves) restarting, so there is little risk. @andlaus?

andlaus:
agreed. merging.

PR #467, opened 2019-01-31
Fix generic boundary conditions for blackoil model

totto82:
This makes it possible to set non-trivial boundary conditions in the blackoil model
Currently the boundary must be set explicit in boundary() in the eclproblem class 
and enableBoundaryMassFlux needs to be set to true in eclfluxmodule.hh

TODO
Make it settable from deck using customized keywords 

Note:
Currently it doesn't allow for polymer or solvent influx
on the boundary with the free flow option


totto82:
jenkins build this please

andlaus:
nice! Since Jenkins is happy, I'll merge tomorrow if there are no vetos.

PR #470, opened 2019-02-05
ebos: make it possible to account for external setup costs

andlaus:
this allows to take the setup time for external objects (of which about 99.5% is usually the parsing time for the deck) into account in the TCPU summary output. in itself, this patch does not change anything, but it makes it possible for `flow` to communicate these costs.


andlaus:
jenkins build this please

andlaus:
jenkins build this please

totto82:
Thanks for the patch. Now parsing time can be included in the TCPU output.  
Also thanks for cleaning up the depricated writeOutput class.

alfbr:
It looks like you are adding the parsing time to the TCPU vector. Being able to measure set-up time is certainly valuable, so thanks for the contribution! It seems clear that at least parts of the parsing time should be included in TCPU, so what is done here might very well be the right thing. However, we also have the vector ELAPSED which typically is closer to wall-time than TCPU. It is a bit unclear to me if we are now adding parts of what should go into ELAPSED rather than TCPU. Have you looked at this?

totto82:
The setup time belongs to TCPU. Currently we don't output ELAPSED. For single core run I thought they where equivalent?

alfbr:
> For single core run I thought they where equivalent?

Actually, they are not. The documentation of TCPU and ELAPSED is unfortunately not complete, so we are left guessing or measuring. The documentation does clearly state that time used on interprocess communication is not accounted for in TCPU, but accounted for in ELAPSED. However, for single core runs, it seems that ELAPSED indeed tries to capture wall-clock time while TCPU tries to capture the amount of time spent in user code. More precisely, using a stop watch I can reproduce the set-up time of the Norne model as ELAPSED while TCPU is about half of that.

andlaus:
The performance tracking machinery used by ewoms/OPM is not equivalent to the one of Eclipse. Reading the ECL documentation, I doubt that it is even possible to implement the eclipse scheme faithfully because the simulators work quite differently and many of these vectors thus simply don't make sense for us. renaming TCPU to ELAPSED is trivial, though.

PR #472, opened 2019-02-07
Add support for using free boundary conditions in Flow

totto82:
The OPM specific keywords FREEBC[XYZ[-]] can be used to specify
boundary cells that are open. Default is a closed boundary.

Depends on OPM/opm-common#641


totto82:
jenkins build this opm-common=641 please

andlaus:
ok, with OPM/opm-common#641 merged, there's nothing holding this back.

PR #471, opened 2019-02-06
Temporarily disable the SPE1 test.

atgeirr:
In order to re-enable it we should:
 - Use input and reference data in opm-tests.
 - Ensure that the jenkins update feature updates it.

I'll self-merge when green, unless someone beats me to it.

PR #473, opened 2019-02-12
do no longer validate the results of any test

### LINKED ISSUE #469: Updating Reference Solution for SPE1CASE1
Commit dfb0647 (PR #461) introduced a new regression test based on the SPE1CASE1 model from opm-tests.  As far as I can tell, there is no automatic way of updating the reference solution&mdash;e.g., in the context of adding new data to the `IWEL` vector&mdash;since this regression test does not plug into the `update_data` request of the Jenkins system.

As of right now that means that said regression test is blocking PR OPM/opm-common#610.  I'd prefer not to disable the test, but I also don't want to take on the responsibility of manually keeping the reference solutions in sync with developments elsewhere.  Can we reach a reasonable compromise here?

andlaus:
this resolves #469.

since the arguments brought forward in that issue equally apply to the tests which produce VTK files, and the willingness to manually update reference results for these is likely even smaller, the results produced by the eWoms integration tests are simply not validated anymore besides the fact that the produced files must be non-empty.

Note that this only applies to the default build system; the alternative system still does -- and will continue to -- validate the results.

@akva2, @alfbr, @atgeirr, @bska, @joakim-hove : I cede the trigger for this to you, since you were the ones who brought up #469 .


andlaus:
jenkins build this please

andlaus:
jenkins build this please

joakim-hove:
> @akva2, @alfbr, @atgeirr, @bska, @joakim-hove : I cede the trigger for this to you, since you were the ones who brought up #469 .

I feel this is part of political decision which I do not have large enough balls to handle; I will not push any button on this.

andlaus:
> I feel this is part of political decision which I do not have large enough balls to handle; I will not push any button on this.

I agree, this *is* a political decision: Either *all* results are validated or none or we allow ourselves to be inconsistent. Since I *really* hate being inconsistent, the decision was made with the merge of #471. 

Alteratively, I propose the SPE1 results to be validated again and just accept the consequences that this implies with the current infrastructure.

andlaus:
@alfbr: mind to merge? you demanded not to validate the results of a test in #469, so you should be in favour of this patch...

alfbr:
> @alfbr: mind to merge? you demanded not to validate the results of a test in #469, 

I did no such thing, and fail to see how this PR relates to #469. This is mostly about VTK-formatted tests.

totto82:
This PR suggests that for the research simulators the minimum requirement of just testing that it compiles and runs (and produces some output) is sufficient. For our end that is fine. That would make the tests easier to maintain and clean up some reference solutions in the ewoms repository. If nobody objects I will merge this tomorrow.

PR #475, opened 2019-02-13
ebos: Fix restart from ECL files

andlaus:
Some time loop stuff was missing in the doobly-doo.

Also, I noticed that using the current version of `opm-common` it seems to be impossible do ECL restarts using a different directory for the restart files than the one which contains the `.DATA` file. is this correct @joakim-hove? if no, how can I specify the directory for `UNRST` files for `opm-common`? If yes, this functionality should IMO be added or else the `--output-dir` parameter is meaningless for restarts if the input data is not copied to that directory before restart.

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -357,7 +357,6 @@ class EclProblem : public GET_PROP_TYPE(TypeTag, BaseProblem)
     enum { enableSolvent = GET_PROP_VALUE(TypeTag, EnableSolvent) };
     enum { enablePolymer = GET_PROP_VALUE(TypeTag, EnablePolymer) };
     enum { enablePolymerMolarWeight = GET_PROP_VALUE(TypeTag, EnablePolymerMW) };
-
     enum { enableTemperature = GET_PROP_VALUE(TypeTag, EnableTemperature) };
     enum { enableEnergy = GET_PROP_VALUE(TypeTag, EnableEnergy) };
     enum { enableThermalFluxBoundaries = GET_PROP_VALUE(TypeTag, EnableThermalFluxBoundaries) };
@@ -737,12 +736,12 @@ public:
         const auto& oilVaporizationControl = this->simulator().vanguard().schedule().getOilVaporizationProperties(epsiodeIdx);
         if (drsdtActive_())
             // DRSDT is enabled
-            for (size_t pvtRegionIdx = 0; pvtRegionIdx < maxDRs_.size(); ++pvtRegionIdx )
+            for (size_t pvtRegionIdx = 0; pvtRegionIdx < maxDRs_.size(); ++pvtRegionIdx)
                 maxDRs_[pvtRegionIdx] = oilVaporizationControl.getMaxDRSDT(pvtRegionIdx)*this->simulator().timeStepSize();
 
         if (drvdtActive_())
             // DRVDT is enabled
-            for (size_t pvtRegionIdx = 0; pvtRegionIdx < maxDRv_.size(); ++pvtRegionIdx )
+            for (size_t pvtRegionIdx = 0; pvtRegionIdx < maxDRv_.size(); ++pvtRegionIdx)
                 maxDRv_[pvtRegionIdx] = oilVaporizationControl.getMaxDRVDT(pvtRegionIdx)*this->simulator().timeStepSize();
 
         wellModel_.beginTimeStep();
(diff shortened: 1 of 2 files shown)


andlaus:
jenkins build this please

joakim-hove:
> it seems to be impossible do ECL restarts using a different directory for the restart files than the one which contains the `.DATA` file. is this correct @joakim-hove? if no, how can I specify the directory for `UNRST` files for `opm-common`?

Well - what you are talking about is the output dir - i.e. where the files generated by flow should go; in addition you need an `input_dir` setting - by default the path embedded in the deck will be used. The `IOConfig`class has [such a method](https://github.com/OPM/opm-common/blob/master/opm/parser/eclipse/EclipseState/IOConfig/IOConfig.hpp#L174), I think flow at a stage had code to call it - but that seems to have fallen off the shelf :-<; so right now I do not think what you want is possible to achieve.

andlaus:
the problem is that the UNRST is not located inside the `input_dir`. I know that this can be hacked around with some shell-scripting magic, but in the past we've added work-arounds  for smaller usability problems in c++.

joakim-hove:
> the problem is that the UNRST is not located inside the `input_dir`

Forget what I wrote - the support was only in my head; sorry. I agree this is something we should have - and I actually thought we already had it.

andlaus:
okay, the `init()` method for the well model was not called on restarts and this affects only the non-toy well model from opm-simulators. this lead me to discover a real bug in the current restart code, though: the initial fluid states and possibly the threshold pressures are not computed on current master, i.e. restarts for decks that are using e.g., non-trivial boundary conditions and (possibly) implicit threshold pressures are incorrect. anyway, this should be fixed in the current version.

andlaus:
jenkins build this please

andlaus:
oops. wrong button.

andlaus:
Also, I'm not sure if SCHEDULE events are handled correctly by the current restart machinery, but that's a story for another day.

andlaus:
jenkins build this please

andlaus:
```
Comparing OPM_IWEL...
Occurrence in first file    = 14
Occurrence in second file   = 8
Value index                 = 0
(first value, second value) = (0, 1)

Program threw an exception: [/var/lib/jenkins/workspace/ewoms-PR-builder/deps/opm-common/examples/test_util/EclRegressionTest.cpp:104] Values of int type differ.
```

`OPM_IWEL` is different. any idea what this means and/or what causes this?

totto82:
> the threshold pressures....

should be stored in the restart files, right? 
> non-trivial boundary conditions...

Do you mean the new FREEBC* stuff.  
> `OPM_IWEL` is different. any idea what this means and/or what causes this?
I can take a look next week if you don't find the solution until then.

andlaus:
>> the threshold pressures....
>
> should be stored in the restart files, right?

in theory yes, but are they in practice?

>> non-trivial boundary conditions...
>
> Do you mean the new FREEBC* stuff.

yes. that and the thermal boundaries in case you want to maintain the geothermal gradient during a simulation.

>> `OPM_IWEL` is different. any idea what this means and/or what causes this?
>
> I can take a look next week if you don't find the solution until then.

that would be fantastic!

andlaus:
ok, I've rebased it. but the jenkins error is caused by the XGRP and XWEL output fields. I could not find them mentioned anywhere in ewoms and opm-simulators (and nowadays the ways of opm-common are quite mysterious to me). how to fix this?

andlaus:
jenkins build this please

ebos/eclproblem.hh
```diff
@@ -1876,35 +1879,60 @@ private:
...
+        for (; elemIt != elemEndIt; ++elemIt) {
+            const auto& elem = *elemIt;
+            if (elem.partitionType() != Dune::InteriorEntity)
+                continue;
+
+            elemCtx.updatePrimaryStencil(elem);
+            int elemIdx = elemCtx.globalSpaceIndex(/*spaceIdx=*/0, /*timeIdx=*/0);
+            initial(sol[elemIdx], elemCtx, /*spaceIdx=*/0, /*timeIdx=*/0);
+        }
+
```
totto82: The solution needs to by synchronized. 
`this->model().syncOverlap()`

This should make Jenkins happy

andlaus: oops, right. Fixed. I guess I owe you a beer now.

ebos/eclproblem.hh
```diff
@@ -1869,20 +1854,33 @@ private:
 
     void readEclRestartSolution_()
     {
+        // Set the start time of the simulation
+        const auto& schedule = this->simulator().vanguard().schedule();
+        const auto& eclState = this->simulator().vanguard().eclState();
```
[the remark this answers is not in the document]
andlaus: both fixed. thanks. I'm wondering a bit why my compiler did not spit out a warning here...

andlaus:
jenkins build this please

PR #444, opened 2018-12-10
Extending the polymer module to handle the polymer molecular weight transport

GitPaean:
The molecular weight will be used to evaluate the polymer viscosity based on a different viscosity model. 

This PR requires the upstream PR https://github.com/OPM/opm-material/pull/317 .

PR #476, opened 2019-02-15
ebos: always use ECLIPSE compatible restart files

andlaus:
this is basically equivalent to #390, but it removes the fiddling-knob to set this.

If jenkins is happy, I intend to self-merge this.

### DIFF
--- a/ebos/eclbasevanguard.hh
+++ b/ebos/eclbasevanguard.hh
@@ -64,7 +64,6 @@ NEW_PROP_TAG(EquilGrid);
 NEW_PROP_TAG(Scalar);
 NEW_PROP_TAG(EclDeckFileName);
 NEW_PROP_TAG(OutputDir);
-NEW_PROP_TAG(EnableOpmRstFile);
 NEW_PROP_TAG(EclStrictParsing);
 NEW_PROP_TAG(EclOutputInterval);
 NEW_PROP_TAG(IgnoreKeywords);
@@ -72,7 +71,6 @@ NEW_PROP_TAG(IgnoreKeywords);
 SET_STRING_PROP(EclBaseVanguard, IgnoreKeywords, "");
 SET_STRING_PROP(EclBaseVanguard, EclDeckFileName, "");
 SET_INT_PROP(EclBaseVanguard, EclOutputInterval, -1); // use the deck-provided value
-SET_BOOL_PROP(EclBaseVanguard, EnableOpmRstFile, true);
 SET_BOOL_PROP(EclBaseVanguard, EclStrictParsing, false);
 
 END_PROPERTIES
@@ -109,8 +107,6 @@ public:
                              "The name of the file which contains the ECL deck to be simulated");
         EWOMS_REGISTER_PARAM(TypeTag, int, EclOutputInterval,
                              "The number of report steps that ought to be skipped between two writes of ECL results");
-        EWOMS_REGISTER_PARAM(TypeTag, bool, EnableOpmRstFile,
-                             "Include OPM-specific keywords in the ECL restart file to enable restart of OPM simulators from these files");
         EWOMS_REGISTER_PARAM(TypeTag, std::string, IgnoreKeywords,
                              "List of Eclipse keywords which should be ignored. As a ':' separated string.");
         EWOMS_REGISTER_PARAM(TypeTag, bool, EclStrictParsing,
@@ -471,8 +467,7 @@ private:
         // specify the directory output. This is not a very nice mechanism because
         // the eclState is supposed to be immutable here, IMO.
         ioConfig.setOutputDir(outputDir);
-
-        ioConfig.setEclCompatibleRST(!EWOMS_GET_PARAM(TypeTag, bool, EnableOpmRstFile));
+        ioConfig.setEclCompatibleRST(true);
     }
 
     Implementation& asImp_()


andlaus:
jenkins build this please

andlaus:
are there any objections? I'll intend to merge this before the ebos move to opm-simulators to avoid unnecessary work...

andlaus:

totto82:
@bska Any reasons for keeping the OPM specific option?

joakim-hove:
> Any reasons for keeping the OPM specific option?

The opm specific version can be used to enable perfect restarts; which save the solution as `double` values in SI units. I (think) that has been a valuable feature while debugging @atgeirr ?

andlaus:
On an only slightly related note I'm currently struggling with the ECL restart stuff in #475. the problem seems to be wells related, so advice is highly appreciated.

andlaus:
jenkins build this opm-tests=95 please

bska:
>> Any reasons for keeping the OPM specific option?
> 
> The opm specific version can be used to enable perfect restarts; which save the solution as double values in SI units. I (think) that has been a valuable feature while debugging @atgeirr ?

As @joakim-hove says, the ability to output solution variables as double precision is certainly valuable.  Another reason would be the ability to output vectors that aren't known to ECLIPSE, but which ResInsight would just pick up and visualise automatically due to being sized according to the number of active cells.  A prime example here would be the molecular weight variables used in @GitPaean's PRs #444 and OPM/opm-simulators#1680.

I do think we should be writing ECLIPSE compatible output files by default, though.

totto82:
What happens if you set --ecl-output-double-precision=true and --enable-opm-rst-file=false. Would Eclipse fail to restart due to double values, or would the double precision request be ignored? 

What happens if you a running a case with @GitPaean's polymer model and ask for output of it in your deck file (is this supported?) and  --enable-opm-rst-file=false. Would the output request be ignored or will Eclipse restart just fail. 

For both of this cases, you are probably not interested in restarting the simulator in eclipse anyway, so as long as there is nothing in the ecl restart feature that prevents the user for adding stuff that is not compatible with eclipse I don't think we need a explicit trigger for opm-rst-file.

bska:
> What happens if you set --ecl-output-double-precision=true and --enable-opm-rst-file=false.

At the moment, setting `opm-rst-file=false` would [override](https://github.com/OPM/opm-common/blob/a1c7e8f5d45c703767d6320fabcb4ee8649a0eda/src/opm/output/eclipse/RestartIO.cpp#L545-L547) `double-precision=true`.  You'd get an output file whose solution variables would be written in single precision (`float`).

> What happens if you a running a case with @GitPaean's polymer model and ask for output of it in your deck file (is this supported?) and  --enable-opm-rst-file=false.

We don't currently support outputting those vectors, though it would be easy to do.  As for what would happen in that case the answer is "It Depends".  Specifically, it depends on how we would classify the variables.  If they were classified as [solution variables](https://github.com/OPM/opm-common/blob/a1c7e8f5d45c703767d6320fabcb4ee8649a0eda/opm/output/data/Cells.hpp#L59), the vectors would be written as expected albeit in single precision.  If they were classiefied as [auxilliary](https://github.com/OPM/opm-common/blob/a1c7e8f5d45c703767d6320fabcb4ee8649a0eda/opm/output/data/Cells.hpp#L60), then they would be [discarded](https://github.com/OPM/opm-common/blob/a1c7e8f5d45c703767d6320fabcb4ee8649a0eda/src/opm/output/eclipse/RestartIO.cpp#L499-L501).

andlaus:
>> What happens if you set --ecl-output-double-precision=true and --enable-opm-rst-file=false.
>
>At the moment, setting `opm-rst-file=false` would [override](https://github.com/OPM/opm-common/blob/a1c7e8f5d45c703767d6320fabcb4ee8649a0eda/src/opm/output/eclipse/RestartIO.cpp#L545-L547) `double-precision=true`.  You'd get an output file whose solution variables would be written in single precision (`float`).

I'd say that this is quite surprising behavior. Anyway, I think that we should always write an E100 compatible restart file, next to a second one which is primarily targeted at visualization tools and restart from OPM. The only drawback I can see with this approach is that it takes a bit more harddisk space, but who does not have a few gigabytes in spare nowadays?

bska:
>>> What happens if you set --ecl-output-double-precision=true and --enable-opm-rst-file=false.
> 
>> At the moment, setting `opm-rst-file=false` would override `double-precision=true`.  You'd get an output file whose solution variables would be written in single precision (float).
> 
> I'd say that this is quite surprising behavior.

Maybe so, but it is mandated.  Requesting ECLIPSE-compatible restart files is fundamentally opposed to double-precision solution variables.  One of those requests must be relaxed and we've elected to have ECLIPSE-compatibility take precendece.  The alternative would be to refuse to write any output at all.

> [W]e should always write an E100 compatible restart file, next to a second one which is primarily targeted at visualization tools and restart from OPM. The only drawback I can see with this approach is that it takes a bit more harddisk space,

There is one more downside.  We would have to teach the visualisation tools about the new file.  It's certainly not an insurmountable task, but might still be a nontrivial amount of work.

andlaus:
> We would have to teach the visualisation tools about the new file. It's certainly not an insurmountable task, but might still be a nontrivial amount of work.

not really: we could produce `OPM_$CASENAME.{EGRID,INIT,UNRST}` files and point ResInsight to them.

totto82:
Based on the discussion, I suggest that we close this PR and instead merge #390. Then we keep the option of output to ecl format without getting under the strict ecl tyranny.  We can remove it later if its never used.

andlaus:
agreed. somebody who's going to do the data update required?

PR #477, opened 2019-02-15
ebos: remove the DisableWells property

andlaus:
the purpose of this was a hack to be able to manipulate the Jacobian to include the wells matrix directly from exterior code. Since `flow` has been converted to the eWoms wells API, this is not required anymore.


andlaus:
jenkins build this please

totto82:
Thanks for the clean-up. We should try to keep the number of parameters as small as possible. For Ebos and Flow disabling the wells should be done in the deck.

PR #478, opened 2019-02-15
ebos: remove the code to calculate a mass conservative initial condtion if the specified condition is inconsistent

andlaus:
As far as I can see, this was always set to `false` anyway.


andlaus:
jenkins build this please

totto82:
Why is the mpi.powerinjection_forchheimer_fd affected by this PR?

andlaus:
it should not be because this is PR at the other end of the spectrum of code used by this test. the error seems to be unrelated and somewhere in libc. maybe there's a rarely triggered race condition somewhere. let's try again.

andlaus:
jenkins build this please

andlaus:
yay! it worked! no changes whatsoever!

totto82:
Thanks. Jenkins ways are sometimes mysterious.

PR #479, opened 2019-02-15
ebos: implement partial support for the TUNING keyword

andlaus:
this only implements the parameters that are currently supported by `flow`.

large parts of this patch are due to a collaboration with @totto82. Thanks!

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -1533,6 +1533,10 @@ public:
      */
     Scalar nextTimeStepSize() const
     {
+        // allow external code to do the timestepping
+        if (this->nextTimeStepSize_ > 0.0)
+            return this->nextTimeStepSize_;
+
         int episodeIdx = this->simulator().episodeIndex();
 
         // for the initial episode, we use a fixed time step size
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -917,18 +917,6 @@ public:
         eclWriter_->writeOutput(isSubStep);
     }
 
-    // this method is DEPRECATED!!!
-    void writeOutput(Opm::data::Wells& dw, Scalar t, bool substep, Scalar totalSolverTime, Scalar nextstep, bool verbose = true)
-    {
-        // use the generic code to prepare the output fields and to
-        // write the desired VTK files.
-        ParentType::writeOutput(verbose);
-
-        // output using eclWriter if enabled
-        if (eclWriter_)
-            eclWriter_->writeOutput(dw, t, substep, totalSolverTime, nextstep);
-    }
-
     /*!
      * \copydoc FvBaseMultiPhaseProblem::intrinsicPermeability
      */
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -139,10 +139,11 @@ NEW_PROP_TAG(EnableThermalFluxBoundaries);
 // The class which deals with ECL aquifers
 NEW_PROP_TAG(EclAquiferModel);
 
-NEW_PROP_TAG(MaxTimeStepSizeAfterWellEvent);
-NEW_PROP_TAG(RestartShrinkFactor);
-NEW_PROP_TAG(MaxFails);
-NEW_PROP_TAG(EnableEclTuning);
+// time stepping parameters
+NEW_PROP_TAG(EclMaxTimeStepSizeAfterWellEvent);
+NEW_PROP_TAG(EclRestartShrinkFactor);
+NEW_PROP_TAG(EclMaxFails);
+NEW_PROP_TAG(EclEnableTuning);
 
 // Set the problem property
 SET_TYPE_PROP(EclBaseProblem, Problem, Ewoms::EclProblem<TypeTag>);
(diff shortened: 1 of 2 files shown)


andlaus:
jenkins build this please

totto82:
The test failure is due to difference in suggested next time step size stored in the restart file. If #476 is merged this PR would hopefully not fail. But we should check if this effects restart before merging.

andlaus:
Thanks for the tip. any idea why happens? I thought `flow` did its time stepping completely independently...

totto82:
Maybe dtNext is outputted before is overwritten (set) by code in opm-simulators? If that is the case the current output of dtNext is maybe not correct either. I guess it is easy to test, just check if the value in OPM-extra is what the timestepper in ewoms suggest or the one in flow. This used to work, and should have been tested in the restart tests, so alternatively this PR messes up the ordering of the setNextTimeStepLenght(...) between code in opm-simulators and ewoms. Anyway it should not be hard to fix.

andlaus:
thanks. I'll have look.

andlaus:
jenkins build this opm-simulators=1731 please

andlaus:
I slowly get the impression that the reference results are incorrect here. can anybody else have a look?

andlaus:
> This used to work, and should have been tested in the restart tests

the restart tests only should be affected if dtNext is smaller than the length of the next report step. I guess that for the tests this was neither the case for plain-ebos nor flow.

alfbr:
Does this impact flow in any way? We needed to store the length of the next time step in order to successfully converge after a restart. Storing it the same location as Eclipse was necessary to use the same file format. This is rather critical, so please inform if you are changing this in any way. I am out of office this week, but if you need to dig out what was done for flow, then @GitPaean probably knows.

andlaus:
> Does this impact flow in any way?

as with #475, it shouldn't because `flow` implements its own time stepper and newton method. That said, it obviously does influence the results because some tests fail. I bet a beer that the reason for both of these failures is somewhere the output code (it is quite a beast). Maybe it even is a bug in the current code, so far nobody nows...

totto82:
I will test this.

totto82:
All tests pass on my machine (without opm-simulators=1731) I also didn't rebase the branch on top of current master (i.e. it uses opm restart files as default) Lets see what jenkins says.

andlaus:
@totto82: I suffer from the same symptoms. (but not all tests pass on my machine, not even on master)

I stared a bit more on what Jenkins has to say here, and I became more and more convinced that the current reference data is wrong: the failing `OPMEXTRA`output field supposedly contains the size of the next time step, right?

andlaus:

alfbr:
OPMEXTRA I believe is only for the old restart format, with Eclipse compatible restart it will not be written out and the size of the next time step is stored in a header. @totto82 , Torbjørn is in office and can probably help you if you need to know where it is stored and read in now.

andlaus:
> OPMEXTRA I believe is only for the old restart format, with Eclipse compatible restart it will not be written out and the size of the next time step is stored in a header.

maybe. but if it is not supposed to be there, why does Jenkins still complain about it?

andlaus:
jenkins build this opm-simulators=1731 update_data please

totto82:
> why does Jenkins still complain about it


The reason is that enable-opm-rst-file=true is hardcoded in the tests. That is probably why Jenkins didn't fail when default was changed.

andlaus:
@totto82 thanks that explains why #390 did not require a data update: it did not change anything w.r.t. the tests

andlaus:
jenkins build this opm-simulators=1731 opm-tests=100 please

andlaus:
*yay*

totto82:
I would like to understand why we need to update the referance solution for this before it is merged. I will do it tomorrow.

joakim-hove:
I would like to merge this family: 
```txt
opm-tests=97 opm-simulators=1732
```
Is that OK for you; or will it create trouble in your ongoing `update_data` endeavours?

andlaus:
go ahead, we can tell jenkins to update the refdata again...

andlaus:
jenkins build this opm-simulators=1731 update_data please

andlaus:
jenkins build this opm-simulators=1731 opm-tests=100 please

ebos/eclproblem.hh
```diff
@@ -139,6 +139,11 @@ NEW_PROP_TAG(EnableThermalFluxBoundaries);
 // The class which deals with ECL aquifers
 NEW_PROP_TAG(EclAquiferModel);
 
+NEW_PROP_TAG(MaxTimeStepSizeAfterWellEvent);
+NEW_PROP_TAG(RestartShrinkFactor);
+NEW_PROP_TAG(MaxFails);
+NEW_PROP_TAG(EnableEclTuning);
+
```
totto82: These new parameters should be hidden from flow.

ebos/eclproblem.hh
```diff
@@ -547,14 +577,26 @@ public:
...
+        // the "NOGRAV" keyword from Frontsim or setting the EnableGravity to false
+        // disables gravity, else the standard value of the gravity constant at sea level
+        // on earth is used
         this->gravity_ = 0.0;
-
-        // the "NOGRAV" keyword from Frontsim disables gravity...
         const auto& deck = simulator.vanguard().deck();
-        if (!deck.hasKeyword("NOGRAV") && EWOMS_GET_PARAM(TypeTag, bool, EnableGravity))
+        if (!deck.hasKeyword("NOGRAV") || !EWOMS_GET_PARAM(TypeTag, bool, EnableGravity))
             this->gravity_[dim - 1] = 9.80665;
```
totto82: Good catch

andlaus:
jenkins build this please

andlaus:
jenkins build this please

ebos/eclproblem.hh
andlaus: right. I'm thinking about adding an `Ecl` prefix as well?!

totto82: Ok, then I will wait with merging this.

andlaus:
okay. the parameters have been renamed and OPM/opm-simulators#1733 hides them from `flow` users.

andlaus:
jenkins build this opm-simulators=1733 please

andlaus:
jenkins build this please

andlaus:
I hope this was the last rebase that is necessary before merge: this PR adhered to Murphy's law in the sense that it conflicted with every single PR which was merged into master since it was opened ;)

PR #480, opened 2019-02-15
ebos: clarify startup message

andlaus:
this hopefully makes the purpose of `ebos` crystal clear in its description. this should be interpreted as "if you use ebos in production, you are on your own and you should only expect a very limited amount of support (or even sympathy) if something breaks".

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -500,8 +500,11 @@ public:
             return
                 "The Ecl-deck Black-Oil reservoir Simulator (ebos); a hydrocarbon "
                 "reservoir simulation program that processes ECL-formatted input "
-                "files which is provided by the Open Porous Media project "
-                "(https://opm-project.org).";
+                "files and is provided by the Open Porous Media project "
+                "(https://opm-project.org).\n"
+                "\n"
+                "THE `ebos` SIMULATOR IS FOR RESEARCH PURPOSES ONLY! For industrial "
+                "applications, use `flow`."
         else
             return briefDescription_;
     }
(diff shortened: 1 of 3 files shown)


andlaus:
jenkins build this please

andlaus:
jenkins build this please

ebos/eclproblem.hh
```diff
@@ -507,8 +507,11 @@ public:
...
                 "The Ecl-deck Black-Oil reservoir Simulator (ebos); a hydrocarbon "
                 "reservoir simulation program that processes ECL-formatted input "
-                "files which is provided by the Open Porous Media project "
-                "(https://opm-project.org).";
+                "files and is provided by the Open Porous Media project "
+                "(https://opm-project.org).\n"
+                "\n"
+                "THE `ebos` SIMULATOR IS FOR RESEARCH PURPOSES ONLY; DO *NOT* USE "
+                "IT IN PRODUCTION! INSTEAD, USE `flow` THERE!";
         else
```
totto82: I think we should avoid the shouting !!!!!!
I suggest.  
-- DO *NOT* USE "
    "IT IN PRODUCTION! INSTEAD, USE `flow` THERE!";  
++  "FOR INDUSTRIAL USAGE, USE INSTEAD 'flow' ";

andlaus: I'd prefer bold, but using plain ASCII, that's not possible. the FSF seems to be of the same opinion, cf our license headers, e.g. https://github.com/OPM/ewoms/blob/master/ebos/ebos.cc#L11-L14 

I changed to the formulation of the last sentence a bit. I hope it is good enough now.

totto82: The -- and ++ was just to show what line I would like to change.

What I suggest is to remove `DO *NOT* USE IT IN PRODUCTION!` and the last `!` 

I am not a writer, but having to many !!!  seems unnecessary. I am sure the `FOR RESEARCH PURPOSE ONLY `makes it clear enough.

andlaus: okay, I see. I hope the current version fixes to that.

totto82: Thanks!

PR #481, opened 2019-02-19
introduce an "experimental mode"

andlaus:
this is a compile time switch with the intention to be able to more easily turn experimental features that are not yet considered to be production quality on and off. DUNE has a similar mechanism (i.e., the `DUNE_GRID_EXPERIMENTAL_GRID_EXTENSIONS` macro), but this relies on the preprocessor.

For now, the property does not have any effect.

### DIFF
--- a/ebos/ebos.cc
+++ b/ebos/ebos.cc
@@ -36,6 +36,11 @@ BEGIN_PROPERTIES
 
 NEW_TYPE_TAG(EclProblem, INHERITS_FROM(BlackOilModel, EclBaseProblem));
 
+// Enable experimental features for ebos: ebos is the research simulator of the OPM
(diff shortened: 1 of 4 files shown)


andlaus:
jenkins build this please

andlaus:
jenkins build this please

totto82:
Wouldn't it be better to have explicit compile time triggers for each new feature? Anyway, I am not against merging this. We can always remove it later if it is not used.

andlaus:
one does not exclude the other. "experimental mode" is just supposed to be a convenient way to switch off everything which is not deemed to be mature enough. for flow/ebos "mature enough" is defined as "ready for production use".

PR #483, opened 2019-02-19
Add option for setting mass boundary rate from deck

totto82:
Usage

BCRATE
1 1 1 1 1 10 X WATER 1e-7 /

This will inject 1e-7 of water (mass/time/length/length) on the x side of the
boundary cells with cartesian index [1 1 1] to [1 1 10]

Depends on OPM/opm-common#648

### DIFF
--- a/ebos/eclproblem.hh
+++ b/ebos/eclproblem.hh
@@ -1260,36 +1260,43 @@ public:
             unsigned indexInInside  = context.intersection(spaceIdx).indexInInside();
             unsigned interiorDofIdx = context.interiorScvIndex(spaceIdx, timeIdx);
             unsigned globalDofIdx = context.globalSpaceIndex(interiorDofIdx, timeIdx);
+            unsigned pvtRegionIdx = pvtRegionIndex(context, spaceIdx, timeIdx);
             switch (indexInInside) {
             case 0:
                 if (freebcXMinus_[globalDofIdx])
                     values.setFreeFlow(context, spaceIdx, timeIdx, initialFluidStates_[globalDofIdx]);
-
+                else
+                    values.setMassRate(massratebcXMinus_[globalDofIdx], pvtRegionIdx);
                 break;
             case 1:
                 if (freebcX_[globalDofIdx])
                     values.setFreeFlow(context, spaceIdx, timeIdx, initialFluidStates_[globalDofIdx]);
-
+                else
+                    values.setMassRate(massratebcX_[globalDofIdx], pvtRegionIdx);
                 break;
             case 2:
                 if (freebcYMinus_[globalDofIdx])
                     values.setFreeFlow(context, spaceIdx, timeIdx, initialFluidStates_[globalDofIdx]);
-
+                else
+                    values.setMassRate(massratebcYMinus_[globalDofIdx], pvtRegionIdx);
                 break;
             case 3:
                 if (freebcY_[globalDofIdx])
                     values.setFreeFlow(context, spaceIdx, timeIdx, initialFluidStates_[globalDofIdx]);
-
+                else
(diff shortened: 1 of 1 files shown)


ebos/eclproblem.hh
```diff
@@ -2290,6 +2297,87 @@ private:
...
+                        throw std::logic_error("invalid direction for BCRATE");
+
+                    int i1 = ratebc.getRecord(record).getItem("I1").template get< int >(0) - 1;
+                    int i2 = ratebc.getRecord(record).getItem("I2").template get< int >(0) - 1;
+                    int j1 = ratebc.getRecord(record).getItem("J1").template get< int >(0) - 1;
+                    int j2 = ratebc.getRecord(record).getItem("J2").template get< int >(0) - 1;
+                    int k1 = ratebc.getRecord(record).getItem("K1").template get< int >(0) - 1;
+                    int k2 = ratebc.getRecord(record).getItem("K2").template get< int >(0) - 1;
+                    const Evaluation rate = ratebc.getRecord(record).getItem("RATE").getSIDouble(0);
+                    for (int i = i1; i <= i2; ++i) {
```
totto82: For know - rates means that mass is added to the reservoar and +rates in the deck means that mass is removed. Is it better to have it the other way around?

akva2: do that and you piss off all mathematicians in sight ;) unicorns, bunnies and puppies will die.

andlaus:
jenkins build this opm-common=648 please

ebos/eclproblem.hh
```diff
@@ -1260,36 +1260,43 @@ public:
             unsigned indexInInside  = context.intersection(spaceIdx).indexInInside();
             unsigned interiorDofIdx = context.interiorScvIndex(spaceIdx, timeIdx);
             unsigned globalDofIdx = context.globalSpaceIndex(interiorDofIdx, timeIdx);
+            unsigned pvtRegionIdx = pvtRegionIndex(context, spaceIdx, timeIdx);
             switch (indexInInside) {
             case 0:
                 if (freebcXMinus_[globalDofIdx])
                     values.setFreeFlow(context, spaceIdx, timeIdx, initialFluidStates_[globalDofIdx]);
 
+                values.setMassRate(massratebcXMinus_[globalDofIdx], pvtRegionIdx);
```
andlaus: if you do it in this order, the rates computed by the free-flow BC will unconditionally be overwritten by some forced mass rate...

ebos/eclproblem.hh
```diff
@@ -2290,6 +2297,87 @@ private:
...
+                    std::string compName = ratebc.getRecord(record).getItem("COMPONENT").getTrimmedString(0);
+                    int compIdx = -999;
+
+                    if (compName == "OIL")
+                        compIdx = oilCompIdx;
+                    else if (compName == "GAS")
+                        compIdx = gasCompIdx;
+                    else if (compName == "WATER")
+                        compIdx = waterCompIdx;
+                    else if (compName == "SOLVENT")
```
andlaus: If you deal with solvent here, probably polymer should be handled as well?

totto82: polymer added

ebos/eclproblem.hh
totto82: Good point. Added a else. No freebc will overwrite massbc if both are set for the same cell and same direction.

andlaus:
jenkins build this please

andlaus:
jenkins build this opm-comon=648 please

andlaus:
from my POV this is ready to be merged if Jenkins concurs. Merge whenever after OPM/opm-common#648 is in.

andlaus:
OPM/opm-common#648 is merged, let's get the simulator side into shape. thanks for the patch @totto82.

PR #485, opened 2019-02-28
start(): call MPI_Init() after dealing with the command line parameters

andlaus:
some MPI libraries (i.e., openMPI 3.1.1.0, the one installed on my system) have the annoying habit of gobbling up about 1 second for `MPI_Init()`. While this is not a big deal when running a simulation, it is really annoying for tab completion. Thus, deal with the parameters before calling `MPI_Init()` so that the help message can be printed instantaneously.

Note that this assumes that `mpirun` does not mess around with the command line parameters. This seems to be the case for at least openMPI and MPICH, but -- as far as I know -- this behavior is not guaranteed by the MPI standard.

A better approach would be to detect if the simulator is started via `mpirun` and only defer the call to `MPI_Init()` if this is not the case. Unfortunately, I am not aware of any way to do this generically. (It is possible for specific implementations, e.g., for openMPI the presence of certain environment variables indicate this.)

Finally, `flow` does not use this code path, so it is unaffected.

### DIFF
--- a/ewoms/common/start.hh
+++ b/ewoms/common/start.hh
@@ -124,9 +124,6 @@ static inline int setupParameters_(int argc, const char **argv, bool registerPar
 
     // first, get the MPI rank of the current process
     int myRank = 0;
-#if HAVE_MPI
-    MPI_Comm_rank(MPI_COMM_WORLD, &myRank);
-#endif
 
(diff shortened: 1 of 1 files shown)


andlaus:
jenkins build this please

PR #487, opened 2019-03-04
remove ebos from packaging

akva2:


andlaus:
thanks. possibly, it should be added to the opm-simulators package as well. probably not -- the target audience is supposed to compile it themselves anyway...

akva2:
opm-simulators already package all installed applications so nothing has to be done.

andlaus:
will the `ebos` binary be included in it automatically?

andlaus:
forget my last comment, I slightly misread yours. Sorry.

PR #489, opened 2019-03-06
changed to have pressure as first variable for 3ph blackoil

hnil:
Move pressure to be first variable.
This together with setting water equation as first equation make some implementations of cpr simpler. 
The pullrequest also adds the spessializations need for using our matrixes with matrix market.

### DIFF
--- a/ewoms/linear/matrixblock.hh
+++ b/ewoms/linear/matrixblock.hh
@@ -232,6 +232,17 @@ public:
 
     BaseType& asBase()
     { return static_cast<BaseType&>(*this); }
+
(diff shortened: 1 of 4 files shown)


atgeirr:
jenkins build this opm-material=324

andlaus:
I'm not sure if this is a good idea, see my comment for OPM/opm-material#324.

andlaus:
(also, jenkins is currently unhappy)

andlaus:
With the CPR stuff merged to opm-simulators, I'm pretty sure that this is obsolete. Closing tentatively.

PR #490, opened 2019-03-13
catch when a process throws

fgfuchs:
Exit the simulation, if load balancing results in a process with zero cells.
depends on a branch with the same name in opm-grid
fixes https://github.com/OPM/opm-simulators/issues/1726 when combined with OPM/opm-grid#361

### DIFF
--- a/ewoms/common/simulator.hh
+++ b/ewoms/common/simulator.hh
@@ -119,11 +119,29 @@ public:
 
         if (verbose_)
             std::cout << "Instantiating the vanguard\n" << std::flush;
-        vanguard_.reset(new Vanguard(*this));
+        int exceptionThrown = 0;
+        try {
+            vanguard_.reset(new Vanguard(*this));
+        }
+        catch (const std::exception& e) {
+            exceptionThrown = 1;
+        }
+        const auto& cc = Dune::MPIHelper::getCollectiveCommunication();
+        if (cc.max(exceptionThrown) == 1) {
+            throw std::runtime_error("Could not instantiate the vanguard.");
+        }
 
         if (verbose_)
             std::cout << "Distributing the vanguard data\n" << std::flush;
-        vanguard_->loadBalance();
+        try {
+            vanguard_->loadBalance();
+        }
+        catch (const std::exception& e) {
+            exceptionThrown = 1;
+        }
+        if (cc.max(exceptionThrown) == 1) {
+            throw std::runtime_error("Could not distribute the vanguard data.");
+        }
 
         if (verbose_)
             std::cout << "Allocating the model\n" << std::flush;
@@ -135,11 +153,27 @@ public:
 
         if (verbose_)
             std::cout << "Finish init of the model\n" << std::flush;
-        model_->finishInit();
+        try {
+            model_->finishInit();
+        }
+        catch (const std::exception& e) {
+            exceptionThrown = 1;
+        }
+        if (cc.max(exceptionThrown) == 1) {
+            throw std::runtime_error("Could not finish init of the model.");
+        }
 
         if (verbose_)
             std::cout << "Finish init of the problem\n" << std::flush;
-        problem_->finishInit();
+        try {
+            problem_->finishInit();
+        }
+        catch (const std::exception& e) {
+            exceptionThrown = 1;
+        }
+        if (cc.max(exceptionThrown) == 1) {
+            throw std::runtime_error("Could not finish init of the problem.");
+        }
 
         setupTimer_.stop();
 


atgeirr:
jenkins build this opm-grid=361 please

ewoms/common/simulator.hh
```diff
@@ -117,34 +127,62 @@ public:
 
         finished_ = false;
 
-        if (verbose_)
+        if (verbose_) {
```
andlaus: in eWoms the preferred style is not to use curly braces no need for braces for single statement blocks -- IMO they are slightly ugly in that they add to the visual noise. (I won't insist on removing them, but please don't complain about the code-churn if somebody else does in a different context.)

ewoms/common/simulator.hh
```diff
@@ -73,6 +72,17 @@ namespace Ewoms {
  * which the problem exhibits boundary conditions and source terms
  * that do not depend on time.
  */
+
+namespace {
+    inline void checkForExceptionsAndThrow(const int exception_thrown, const std::string& message)
+    {
+        auto cc = Dune::MPIHelper::getCollectiveCommunication();
```
andlaus: this is an instance where I want a `const` (and a reference):

```c++
const auto&o cc = Dune::MPIHelper::getCollectiveCommunication();
```

also, there's no reason to put the exception handling code into a seperate function instead of just add it in the simulator directly? Copying two lines of code a few times does not hurt too much and makes the user functions much easier to read. (please enlighten me if I'm wrong.)

finally, why is the collective communication object retrieved from the MPIHelper and not from the gridView? the `MPIHelper` will probably not work properly with dune-fem which is a problem here because the `Simulator` class is as general as it can get.

ewoms/common/simulator.hh
```diff
@@ -117,34 +127,62 @@ public:
...
 
-        if (verbose_)
+        if (verbose_) {
             std::cout << "Instantiating the vanguard\n" << std::flush;
-        vanguard_.reset(new Vanguard(*this));
+        }
+        int exception_thrown = 0;
+        try {
+            vanguard_.reset(new Vanguard(*this));
+        } catch (const std::exception& e) {
```
andlaus: new line after `}`, please.

ewoms/common/simulator.hh
```diff
@@ -117,34 +127,62 @@ public:
 
         finished_ = false;
 
-        if (verbose_)
+        if (verbose_) {
             std::cout << "Instantiating the vanguard\n" << std::flush;
-        vanguard_.reset(new Vanguard(*this));
+        }
+        int exception_thrown = 0;
```
andlaus: camelCase instead of_underscores!

ewoms/common/simulator.hh
```diff
@@ -73,6 +72,17 @@ namespace Ewoms {
  * which the problem exhibits boundary conditions and source terms
  * that do not depend on time.
  */
+
+namespace {
+    inline void checkForExceptionsAndThrow(const int exception_thrown, const std::string& message)
```
andlaus: camelCase instead of_underscores.

ewoms/common/simulator.hh
```diff
@@ -117,34 +127,62 @@ public:
...
+        checkForExceptionsAndThrow(exception_thrown, "Could not instantiate the vanguard.");
 
-        if (verbose_)
+        if (verbose_) {
             std::cout << "Distributing the vanguard data\n" << std::flush;
-        vanguard_->loadBalance();
+        }
+        try {
+            vanguard_->loadBalance();
+        } catch (const std::exception& e) {
```
andlaus: extra new line, please.

ewoms/common/simulator.hh
fgfuchs: ok

ewoms/common/simulator.hh
fgfuchs: agree

ewoms/common/simulator.hh
fgfuchs: ok

fgfuchs:
@andlaus I have incorporated your requests for change

atgeirr:
jenkins build this opm-grid=361 please

andlaus:
thanks. All good now, merge at leasure.

ewoms/common/simulator.hh
fgfuchs: gridView().comm().max behaves differently from  Dune::MPIHelper::getCollectiveCommunication().max
The first one does not work, meaning
`const auto& cc = Dune::MPIHelper::getCollectiveCommunication();
        if (cc.max(exceptionThrown) == 1) {
            throw std::runtime_error("message.");
        }
` throws on all processes but
`if (gridView().comm().max(exceptionThrown) == 1) {
            throw std::runtime_error("message.");
        }
` does not.
 Do you know what the differenc is @andlaus

fgfuchs:
jenkins build this opm-grid=361 please

ewoms/common/simulator.hh
andlaus: I don't. So far, I assumed they were identical. This seems like a DUNE-pecularity, so maybe @blattms or @dr-robertk know more.

Anyway, instead of relying on the MPIHelper, you could use the `MPI_Allreduce()` function directly. Just make sure that no processes have been terminated before you call it. (this might also explain your observation, BTW.)

ewoms/common/simulator.hh
```diff
@@ -135,11 +153,27 @@ public:
...
             std::cout << "Finish init of the model\n" << std::flush;
-        model_->finishInit();
+        try {
+            model_->finishInit();
+        }
+        catch (const std::exception& e) {
+            exceptionThrown = 1;
+        }
+        if (cc.max(exceptionThrown) == 1) {
+            throw std::runtime_error("Could not finish init of the model.");
```
andlaus: you probably need to call `MPI_Finalize()` here before you throw, else you will get undefined behavior due to some processes being terminated while others still want to communicate.

ewoms/common/simulator.hh
fgfuchs: Does not the following line include MPIHelper anyway?
```#include <dune/common/parallel/mpihelper.hh>```
blame shows that this was committed on 19 Jun 2015 by @andlaus

Comment (andlaus) [reply ewoms/common/simulator.hh]: if dune-fem is found, the dune-fem mechanism for initializing MPI is used, i.e., Dune::MPIHelper is called. maybe they have some code in dune-fem which makes Dune::MPIHelper work correctly, but I would not bet on this. what's the state of affairs in this regard, @dr-robertk?

Comment (fgfuchs) [reply ewoms/common/simulator.hh]: I see the line
```
verbose_ = verbose && Dune::MPIHelper::getCollectiveCommunication().rank() == 0;
```
being used in the same file.
Does not this mean that I can use 
Dune::MPIHelper::getCollectiveCommunication().max() as well?

atgeirr:
Looks fine, merging!

If the MPIHelper issue turns out to be important, it must be handled for the existing usage anyway.

PR #491, opened 2019-03-13
Add support for water induced rock compaction

totto82:


ewoms/models/blackoil/blackoilproblem.hh
```diff
@@ -140,6 +149,15 @@ public:
                                  unsigned timeIdx OPM_UNUSED) const
     { return 1e5; }
 
+    template <class LhsEval, class Context>
+    LhsEval getPoreVolumeMultiplier(const LhsEval& pressure OPM_UNUSED,
+                                       const Context& context OPM_UNUSED,
```
andlaus: if you pass a context object, you don't need to pass the pressure explicitly because the latter is available via `context.intensiveQuantities(spaceIdx, timeIdx).fluidState().pressure(oilPhaseIdx)`.

also: the indentation of this is slightly wrong. (just a nit.)

ewoms/models/blackoil/blackoilsolventmodules.hh
```diff
@@ -1510,15 +1510,18 @@ public:
 
         Scalar faceArea = elemCtx.stencil(timeIdx).interiorFace(scvfIdx).area();
         const IntensiveQuantities& up = elemCtx.intensiveQuantities(solventUpstreamDofIdx_, timeIdx);
+        const auto& problem = elemCtx.problem();
+        const auto& stencil = elemCtx.stencil(timeIdx);
+        const Evaluation transModified = trans*problem.getTransmissibiltyMultiplier(up.fluidState().pressure(FluidSystem::oilPhaseIdx), timeIdx, stencil.globalSpaceIndex(solventDownstreamDofIdx_));
```
andlaus: I think it is better to see this as an implementation detail of `Problem::transmissibility()` and move the code there. As far as I can see, this would even make the functionality fully contained in `opm-simulators`?

ewoms/disc/common/fvbasediscretization.hh
```diff
@@ -673,7 +673,7 @@ public:
             !intensiveQuantityCacheUpToDate_[timeIdx][globalIdx])
             return 0;
 
-        if (timeIdx > 0 && enableStorageCache_)
+        if (timeIdx > 0 && enableStorageCache_ && simulator_.problem().recycleFirstIterationStorage())
```
andlaus: I guess that the base problem class will need a `recycleFirstIterationStorage()` to make the the eWoms tests compile?

(also, it would be great if this method could be avoided entirely. I'll have a look tomorrow.)

PR #492, opened 2019-03-15
Add transpose to MatrixBlock.

atgeirr:
This is needed for the CPR work.

### DIFF
--- a/ewoms/linear/matrixblock.hh
+++ b/ewoms/linear/matrixblock.hh
@@ -232,6 +232,16 @@ public:
 
     BaseType& asBase()
     { return static_cast<BaseType&>(*this); }
+
+    MatrixBlock<Scalar, m, n> transpose() const {
+        MatrixBlock<Scalar, m, n> out;
+        for (int i=0; i<n; i++){
+            for (int j=0; j<m; j++){
+                out[j][i] = (*this)[i][j];
+            }
+        }
+        return out;
+    }
 };
 
 } // namespace Ewoms


andlaus:
I'm not sure if doing it like this is a terribly good idea: it will prevent using `Dune::FieldMatrix` for code which uses this. Making this a free function (e.g., `Ewoms::transposeMatrix(M)`), would be a better approach as far as I can see.

andlaus:
or -- even better -- move this free function to the CPR code in opm-simulators.

atgeirr:
> it will prevent using `Dune::FieldMatrix` for code which uses this.

I think the problem is the opposite: Dune::FieldMatrix.transpose() is defined and used, but fails for MatrixBlock. (@hnil please correct me if I misunderstood this.)

blattms:
It might fail because the code might not see FieldMatrix but only
DenseMatrix (which does not have the transpose function)

andlaus:
I'm not sure which version of DUNE you use, but for Dune 2.6, the following snippet does not compile on my machine:

```c++
#include <dune/common/fmatrix.hh>

int main(int argc, char **argv)
{
    Dune::FieldMatrix<double, 2, 2> foo;
    foo.transpose();
    return 0;
}
```

blattms:
I did not check and just believed somebody claiming it existed. Indeed it does not exist and I was talking about an imaginary version.

andlaus:
@atgeirr: please close this PR if it has become obsolete.

PR #493, opened 2019-03-21
do not use BCRSMatrix::entry()

andlaus:
this method seems to be intended only for BCRS matrices which are build in implicit mode. This might be a mistake in DUNE because the method works fine for any matrices provided that they are fully set up.

This assert is only triggered if the `DUNE_ISTL_WITH_CHECKING` macro is set, so the impacts of this mistake should be quite limited.


andlaus:
jenkins build this please

totto82:
Thanks for the fix.

PR #494, opened 2019-03-21
black oil newton method: allow the compiler to see at compile time if  a given extension is enabled

andlaus:
i.e., we now check for `enable$BLACKOIL_EXTENSION` compile time variables. this should allow the compiler to produce slightly better optimized machine code if the respective extension is disabled.

### DIFF
--- a/ewoms/models/blackoil/blackoilnewtonmethod.hh
+++ b/ewoms/models/blackoil/blackoilnewtonmethod.hh
@@ -284,6 +284,9 @@ protected:
                     nextValue[pvIdx] = 0.0;
             }
 
+            // keep the temperature above 100 and below 1000 Kelvin
+            if (enableEnergy && pvIdx == Indices::temperatureIdx)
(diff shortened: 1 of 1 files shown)


andlaus:
jenkins build this please

andlaus:
jenkins build this please

andlaus:
okay, there was a stupid mistake here: if you want the minimal temperature to be 100 K, you have to use `std::max(100, T)`, not `std::min(100, T)`...

andlaus:
jenkins build this please

andlaus:
with the stupidity fixed, it actually works. Wow?

@totto82: if you're okay with this, I'd appreciate if you pressed Green.

totto82:
Does this fix the convergence issue for the energy model? 

Anyway, looks good, and Jenkins is happy.

andlaus:
thanks for merging!

> Does this fix the convergence issue for the energy model?

not that I know of. While doing this, I just stumbled over the fact that all other blackoil extensions had crash barriers around their update paths but not the energy extension. I thought that it was a good idea to fix this while at it...

PR #495, opened 2019-03-22
simulator: print the error messages of the thrown exceptions in case something goes wrong during init

andlaus:
this only affects binaries which do use set `verbose_` flag, i.e., not `flow`.


andlaus:
jenkins build this please

andlaus:
since Jenkins is happy and this does not affect `flow`, I'll shamelessly self-merge.

PR #497, opened 2019-03-25
fixing a small mistake in getSkprwatTable

GitPaean:
fixing error reported in https://github.com/OPM/ewoms/issues/496 by @bska


GitPaean:
jenkins build this please

andlaus:
Jenkins is happy and the patch is rather, so there's no reason not to merge. Thanks for finding and fixing!

PR #498, opened 2019-03-25
simulator: properly handle exceptions in the pre and postprocessing routines in parallel runs

andlaus:
`flow` is unaffected by this, because it uses its own main loop.


andlaus:
jenkins build this please

PR #499, opened 2019-03-26
make all headers self sufficient

andlaus:
this is part of the pre-release maintainance: all eWoms headers should now be includable without strings attached, i.e., in any order and without having to include something before.


andlaus:
jenkins build this please

totto82:
Thanks for clean up.

PR #500, opened 2019-04-03
simulator: simplify the time step handling a bit

andlaus:
`timeStepSize()` now always returns the value specified via `setTimeStepSize()`, i.e., the time loop is now solely responsible for not crossing episode boundaries in the middle of time steps and for not going past the specified end time of the simulation.


andlaus:
jenkins build this please

totto82:
Looks good.

PR #501, opened 2019-04-09
black oil: fix the determination of the maximum oil saturation

andlaus:
the oil saturation for the current iteration can be larger than the historic maximum. since oil saturation typically goes down over time in oil reservoirs, the impact of this is likely very limited...

this PR depends on OPM/opm-material#332 (but the inverse is not the case)

### DIFF
--- a/ewoms/io/vtkblackoilmodule.hh
+++ b/ewoms/io/vtkblackoilmodule.hh
@@ -193,8 +193,6 @@ public:
         if (!EWOMS_GET_PARAM(TypeTag, bool, EnableVtkOutput))
             return;
 
-        typedef Opm::MathToolbox<Evaluation> Toolbox;
-
         for (unsigned dofIdx = 0; dofIdx < elemCtx.numPrimaryDof(/*timeIdx=*/0); ++dofIdx) {
             const auto& fs = elemCtx.intensiveQuantities(dofIdx, /*timeIdx=*/0).fluidState();
             typedef typename std::remove_const<typename std::remove_reference<decltype(fs)>::type>::type FluidState;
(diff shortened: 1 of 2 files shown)


andlaus:
jenkins build this opm-material=332 please

GitPaean:
What is `SoMax` used for? Is this PR trying to fix a problematic scenario?

I am a little worrying about the generality of the PR.  Extreme scenarios like oil injecting, not that extreme scenarios, like flooding. 

Hopefully, there is some theoretical support for this PR.

andlaus:
> What is SoMax used for?

VAPPARS

> Is this PR trying to fix a problematic scenario?

no: I just stumbled over this while doing something that requires the maximum water saturation.

> I am a little worrying about the generality of the PR. Extreme scenarios like oil injecting, not that extreme scenarios, like flooding.

can you be more specific? note that even before this PR, the solution will be correct, only the convergence may be slower.

> Hopefully, there is some theoretical support for this PR.

as stated above, the theory is that the maximum oil saturation can occur for the most recent iteration. Note that this does not necessarily mean "oil injection" but that oil can be displaced into cells with low Sw by water or gas injectors.

GitPaean:
Okay. My bad. I thought you were stating `the oil saturation for the current iteration can be larger than the historic maximum` is a problem. Your PR is trying to enforce the above statement. 

I take back what I said.

andlaus:
mea culpa! I see that I also made a mistake while commenting: I thought your comment was for OPM/opm-material#332. The answers above mostly apply anyway, except for the fact that the converged solution may be slightly different in highly exotic cases. (i.e., if oil saturation increases with time.)

andlaus:
> max oil saturation should be removed from the element context and added to fluidState. 

I severely disagree: the max oil saturation is an ugly ECL specific hack, while fluid states are supposed to represent the thermodynamic state of a given location and I never saw a "maximum oil saturation" mentioned in any thermodynamics text book.

> As it is done now it will miss derivatives if it is increasing.

I agree if you mean "as done in `master`". The whole point of this PR and OPM/opm-material#332 is fixing this, though.

GitPaean:
I guess he meant `maxOilSaturation` return a `Scalar`. Do we need to keep some derivatives here?

andlaus:
the problem's `maxOilSaturation()` function is a slight misnomer now: it should be called something like `historicMaxOilSaturation()`. (this does not depend on the current solution, agree?)

GitPaean:
okay. then `fluidState_.saturation(oilPhaseIdx),` is the current solution. 

Yes. better naming of the function of comments help the reading.

PR #503, opened 2019-04-24
Removes unsused function parameter warning

blattms:
Would be nice if this would get into the release.


blattms:
jenkins build this please

andlaus:
thanks. hint: to faciliate documenting these parameters with doxygen, it is preferable to use the `OPM_UNUSED` macros...

PR #506, opened 2019-05-10
Check each command line parameter for a correctly specified value

blattms:
and print a meaningful error message. ewoms expects the format
--param=value. Some users might be used to --param value and will
get an exception thrown when using it.

Without this running 'flow --output-dir dir SPE1CASE2' results in the
rather cryptic message (when compiled wih debugging symbols, otherwise
it should be even less usable):

```
terminate called after throwing an instance of 'std::out_of_range'
  what():  basic_string::substr: __pos (which is 1) > this->size() (which is 0)
[smaug:111393] *** Process received signal ***
[smaug:111393] Signal: Aborted (6)
[smaug:111393] Signal code:  (-6)
[smaug:111393] [ 0] /lib/x86_64-linux-gnu/libpthread.so.0(+0x12730)[0x7f8172fe3730]
[smaug:111393] [ 1] /lib/x86_64-linux-gnu/libc.so.6(gsignal+0x10b)[0x7f817119c7bb]
[smaug:111393] [ 2] /lib/x86_64-linux-gnu/libc.so.6(abort+0x121)[0x7f8171187535]
[smaug:111393] [ 3] /usr/lib/x86_64-linux-gnu/libstdc++.so.6(+0x8c983)[0x7f81713ff983]
[smaug:111393] [ 4] /usr/lib/x86_64-linux-gnu/libstdc++.so.6(+0x928c6)[0x7f81714058c6]
[smaug:111393] [ 5] /usr/lib/x86_64-linux-gnu/libstdc++.so.6(+0x92901)[0x7f8171405901]
[smaug:111393] [ 6] /usr/lib/x86_64-linux-gnu/libstdc++.so.6(+0x92b34)[0x7f8171405b34]
[smaug:111393] [ 7] /usr/lib/x86_64-linux-gnu/libstdc++.so.6(+0x8e891)[0x7f8171401891]
[smaug:111393] [ 8] /usr/lib/x86_64-linux-gnu/libstdc++.so.6(_ZNKSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEE6substrEmm+0x57)[0x7f8171493767]
[smaug:111393] [ 9] ../../opm-simulators/opm-parallel-debug/bin/flow(_ZN5Ewoms10Parameters23parseCommandLineOptionsINS_10Properties4TTag13FlowEarlyBirdEFiRSt3setINSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEESt4lessISB_ESaISB_EERSB_iPPKciiEEESB_iSK_RKSB_RKT0_+0x7f3)[0x5646ea539642]
[smaug:111393] [10] ../../opm-simulators/opm-parallel-debug/bin/flow(+0x1579a9b)[0x5646ea530a9b]
[smaug:111393] [11] ../../opm-simulators/opm-parallel-debug/bin/flow(_ZN3Opm12FlowMainEbosIN5Ewoms10Properties4TTag13FlowEarlyBirdEE16setupParameters_EiPPc+0x1e5)[0x5646ea525171]
[smaug:111393] [12] ../../opm-simulators/opm-parallel-debug/bin/flow(main+0xfa)[0x5646ea4e1c95]
[smaug:111393] [13] /lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xeb)[0x7f817118909b]
[smaug:111393] [14] ../../opm-simulators/opm-parallel-debug/bin/flow(_start+0x2a)[0x5646ea4e11ca]
[smaug:111393] *** End of error message ***
```

### DIFF
--- a/ewoms/common/parametersystem.hh
+++ b/ewoms/common/parametersystem.hh
@@ -720,6 +720,16 @@ std::string parseCommandLineOptions(int argc,
         }
         seenKeys.insert(paramName);
 
+        if (s.empty() || s[0] != '=' || s.size()==1) {
+            std::string msg =
+                std::string("Parameter '")+paramName+"' is missing a value. "
+                +" Please use "+argv[i]+"=value.";
+
+            if (!helpPreamble.empty())
+                printUsage<TypeTag>(helpPreamble, msg, std::cerr);
+            return msg;
+        }
+
         paramValue = s.substr(1);
 
         // Put the key=value pair into the parameter tree


blattms:
Not finished yet as the errror is message is also show if one uses an unknown parameter.

blattms:
That was such an embarrassing mistake.

blattms:
jenkins build this please

atgeirr:
> Not finished yet as the errror is message is also show if one uses an unknown parameter.

It is not clear to me: did you address this? When I run with an unknown parameter the simulator starts, but will print a message about the unknown parameter before the main simulation loop. I guess this is just as before? Did you aim to change this? Perhaps we should just refuse to run if there are unknown parameters, as they are guaranteed to be typing mistakes.

blattms:
Sorry for not being clear. Unknown parameters are still accepted and simulator
runs through. I will address this in an upcoming PR.

The bug that I experienced with a previous version and triggered the comment
is gone now. IMHO this is finished.

atgeirr:
> IMHO this is finished.

Good! There is just one nitpick: spaces around binary ops. Will merge when addressed. Also assuming you want #507 to be dealt with separately?

blattms:
> Good! There is just one nitpick: spaces around binary ops.

Sigh. No offence, but I am getting confused and am annoyed by constantly demanding different coding styles. I guess we need to decide whether ebos stays like it was in ewoms or the same coding standards now apply to it as to the rest of opm-simulators.

Do not get me wrong I am all for the latter option (including *.hh becoming *.hpp etc.) as this seems to be less annoying then constantly switching between styles. But last week someone else insisted on the former and no other developer opposed to insisting on ewoms coding style within ebos. Therefore this time I opted to do this right away. But I seem to be  wrong again and wasted time.

I just want to know what I should do in advance.

blattms:
Well one should look/understand before shooting. Turns out I did not comprehend what binary operators are. Indeed there always are spaces around them in both styles. Will fix somewhen.

andlaus:
NB: I am of the considerate opinion that aborting the run on encountering an unknown parameter is the policy of the simulator must be thus done there. For `flow` this would mean adding `std::abort()` to were the unknown parameters are printed, IOW the parameter system ought not be affected by this.

blattms:
> NB: I am of the considerate opinion that aborting the run on encountering an unknown parameter is the policy of the simulator must be thus done there.

Not sure how this is related with this PR (other than one of my comments). But good to know.

Please be aware that even your current code will abort directly in ewoms if the unknow parameter has e.g. underlines. Just run `flow --output_dir=bla <case>` and see for yourself.

andlaus:
I'll merge this now: I'm not enough of a nitpicker to care enough about the missing spaces and while I don't consider the current master's code being buggy in a strict sense (it does not produce incorrect behavior or  segfaults), the present PR is certainly a nice usability improvement. thanks for the patch!

blattms:
guess I am too late for fixing the space around binary operatores. thanks for merging.

PR #507, opened 2019-05-13
Allow '--parameter value' passing on the command line.

blattms:
Up to now only '--parameter=value' was allowed. This was about
annoying as most Linux/Unix users are accustomed to both possible
variations.

### DIFF
--- a/ewoms/common/parametersystem.hh
+++ b/ewoms/common/parametersystem.hh
@@ -52,6 +52,7 @@
 #include <iostream>
 #include <fstream>
 #include <unordered_map>
+#include <cstring>
 
 #include <unistd.h>
 #include <sys/ioctl.h>
@@ -704,8 +705,20 @@ std::string parseCommandLineOptions(int argc,
             return oss.str();
         }
 
-        // copy everything after the "--" into a separate string
-        std::string s(argv[i] + 2);
+        std::string s;
+
+        if (strchr(argv[i], '=') == nullptr && i < argc - 1
+            && strlen(argv[i+1]) && argv[i+1][0]!='-' ){
+            // User seems to use spaces between param and value
+            // copy everyting after "--" and use the next argument as value
+            std::ostringstream oss;
+            oss << (argv[i] + 2) << '=' << argv[i+1];
+            s = oss.str();
+            ++i; // skip the next argument
+        } else {
+            // copy everything after the "--" into a separate string
+            s = std::string(argv[i] + 2);
+        }
 
         // parse argument
         paramName = transformKey_(parseKey_(s), /*capitalizeFirst=*/true);


blattms:
jenkins build this please

atgeirr:
How will this treat the special positional parameter for the input deck?

blattms:
I do not understand your question. What are positional parameters?

atgeirr:
> What are positional parameters?

Maybe not a well-defined term, but what I was thinking of was that a single parameter is allowed to be *not* of the form --key=value, and if such a parameter exists, it is used for the EclDeckFileName (or ecl-deck-file-name) parameter. With your change, any number of parameters can be standalone values, I just wanted to know that it did not interfere with the ability to set the deck as a standalone parameter.

andlaus:
> What are positional parameters?

these are parameters that do not follow the `--key=value` convention, i.e., they depend on the their relative position. as @atgeirr mentioned this currently means specifying the deck file without specifying the parameter name explicitly (i.e., using `ebos DECK.DATA` instead of `ebos --ecl-deck-file-name=DECK.DATA`).

andlaus:
while I do not think this issue technically fulfills to the definition of a bug because the current master neither misbehaves nor segfaults, the patch constitutes a nice usability improvement IMO. Jenkins must be made happy before it can be merged, though.

andlaus:
oops. I mistook this pr for #506 while writing my last comment. rapid-fire PRs for almost the same thing are quite confusing...

andlaus:
I'll close this for now. If there is a generic solution (which preferably should also deal with stuff like `--help` and `--version` in a generic fashion), this issue should IMO be revisited.

blattms:
Maybe I am missing something. Are there any keywords not taking a value besides --help and --version?
The former works like a charm. The latter not but that is because it needs to be handled differently in opm-simulators, which is the only place where it is queried. There is a check missing  that there are no other parameters besides --version.

Also specifying the deckfile as the last argument on the command line works as long as there is no parameter without a value before it. So the question really is are we supporting these besides --help and --version.

I totally agree with having a generic system like getop by the way, but I kinf of have the feeling that this is not what you meant.

andlaus:
> Are there any keywords not taking a value besides --help and --version?

not that I know of anyway.

> The former works like a charm. The latter not but that is because it needs to be handled differently in opm-simulators, which is the only place where it is queried. There is a check missing that there are no other parameters besides --version.

yes. the reason for this is that the latter was "bolted" on flow, cf. OPM/opm-simulators#1677

> Also specifying the deckfile as the last argument on the command line works as long as there is no parameter without a value before it.

yes. the first positional argument is interpreted as the deck file by `ebos` and its derivatives. I think this is how it is supposed to be.

> So the question really is are we supporting these besides --help and --version.

I'm not aware of anything at the moment. That said, it can be argued that generic support for flag arguments a.l.a. `--disable-ecl-output` instead of `--enable-ecl-output=false` is desirable.

> I totally agree with having a generic system like getopt by the way, but I kind of have the feeling that this is not what you meant.

go ahead and provide a patch. As usua,l I have to insist on a generic solution and reasonable code quality if it affects anything besides `flow`, though.

blattms:
Well, then this branch should work as expected together ewoms and with OPM/opm-simulators#1842 even for flow.

PR #508, opened 2019-05-13
Let Parameters::printValues() return a list of unknown keywords.

blattms:
This will allow programs to print this list and abort the run.

### DIFF
--- a/ewoms/common/parametersystem.hh
+++ b/ewoms/common/parametersystem.hh
@@ -52,6 +52,7 @@
 #include <iostream>
 #include <fstream>
 #include <unordered_map>
+#include <tuple>
 
 #include <unistd.h>
 #include <sys/ioctl.h>
@@ -797,14 +798,17 @@ void parseParameterFile(const std::string& fileName, bool overwrite = true)
 
 /*!
  * \ingroup Parameter
- * \brief Print values of the run-time parameters.
+ * \brief Creates list of keywords parsed by the parameter system.
  *
- * \param os The \c std::ostream on which the message should be printed
+ * \return A tuple consisting of the list of valid runtime keywords,
+ *         a list of all runtime parameters parsed, and a list of
+ *         unknown keywords parsed.
  */
 template <class TypeTag>
-void printValues(std::ostream& os = std::cout)
+std::tuple<std::list<std::string>,std::list<std::string>,std::list<std::string> >
+getParsedKeywords()
 {
-    typedef typename GET_PROP(TypeTag, ParameterMetaData) ParamsMeta;
+    using ParamsMeta = typename GET_PROP(TypeTag, ParameterMetaData);
 
     const Dune::ParameterTree& tree = ParamsMeta::tree();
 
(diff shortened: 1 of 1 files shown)


blattms:
Marked wip until other PR is tested.

blattms:
Finished. Should be tested with OPM/opm-simulators#1840

andlaus:
not everybody wants to print something before being able to inspect the unknown parameters. better introduce a `EWOMS_GET_UNKNOWN_PARAMS()` macro to that end.

blattms:
You do not have to. there is a function without printing that is even used by
flow. Or do you suddenly care about milliseconds?

andlaus:
> You do not have to.

**my** point is that while it is certainly possible to use printValues_() without it producing any terminal output by specifying a non-printing `std::ostream`, it is certainly not the intended use case and -- at least as long as I have something to say here -- such ugly hacks do not have a place in eWoms if their only benefit is the author being able to be a bit lazier.

> there is a function without printing that is even used by flow.

yes. it's purpose is to retrieve a message which can be diverted to `Opm::Log`. that's quite a different usecase.

> Or do you suddenly care about milliseconds?

no, I do not care about milliseconds. do you?

blattms:
> **my** point is that while it is certainly possible to use printValues_() without it producing any terminal output by specifying a non-printing `std::ostream`, it is certainly not the intended use case and -- at least as long as I have something to say here -- such ugly hacks do not have a place in eWoms if their only benefit is the author being able to be a bit lazier.
>

Sorry I still do not get your point. Maybe you can point out what is a hack in the code in your opinion. Nobody is using a non-printing `std::ostream to just get a list of unknown keywords
(or at least my changes do not do that if it was not there before).

andlaus:
the hack is to misuse a function that has the word `print` in its name to do something decidedly non-printy!

blattms:
Isn't that a bit unreasonable? Anyway, I will stop here and I am out now.

andlaus:
yes, I think the hack is unreasonable considering the fact that its only "advantage" is that it allows to be slopier when implementing the functionallity you currently seem to be working on. (That said, if you want the code to be impenetrable on purpose, I see your point.)

joakim-hove:
Andreas your constant stream of snide remarks like:

>That said, if you want the code to be impenetrable on purpose, I see your point.)

Is extremely detrimental to the opm project in general, and specifically your standing as a reasonable developer.

Position yourself in front of mirror, take a deep breath and read this sentence out loud: "All the developers in the opm project work hard and do their best for the project to become a success. Sometimes we disagree on technical matters, that is fair enough - but everyone have the best intentions".

OK - so how did that feel?

andlaus:
> Andreas your constant stream of snide remarks like: [...]

I accept at most 50% of the blame for the present PR: just not reusing the print function to retrieve the list of unknown parameters would have saved all involved a lot of excitement. Anyway, while I will not merge this PR in its current form, I'm sure that you'll be glad to hear that you will not have to endure my pesky comments for much longer.

blattms:
@andlaus You are stalling progress. You might disagree but at least the big
picture of this change (stopping a run when unknown keywords are encountered,
OPM/opm-simulators#1840) seems to an improvement for at least 3 OPM people.

For me it is still not clear why return a list of keywords from a print
function is a bug. operator<<() returns a stream not the same thing, but the
function name does not say it. For people calling
```
printValues();
```
Nothing changes, not even the runtime as the copy should get optimized away.
People calling
```
auto list = printValues()
```
get the list basically for free (not additional parsing run). That is why this
route was taken.

I am still not sure I understand your complaint 100%. But lets try to make you
happy/find a compromise.
To make you happy should I
- rename printValues to printValuesAndReturnUnknownKeywords. I am a bit
  reluctant to do that because this looks like an interface change to me
  that would break other peoples code.
- duplicate the code in this method and use it to implement a method
  get unknownKeywords() that just creates the a list of unknown keywords

or maybe you have another option that I do not see?

OPMUSER:
>  You might disagree but at least the big picture of this change (stopping a run when unknown keywords are encountered, OPM/opm-simulators#1840) seems to an improvement for at least 3 OPM people.

+1

blattms:
... and people just interested in the list of keywords parsed can call `getParsedKeywords()`

PR #510, opened 2019-05-22
Enable onephase

hnil:
Changes to make one phase with and without thermal option run

### DIFF
--- a/ewoms/common/simulator.hh
+++ b/ewoms/common/simulator.hh
@@ -125,7 +125,7 @@ public:
         time_ = 0.0;
         endTime_ = EWOMS_GET_PARAM(TypeTag, Scalar, EndTime);
         timeStepSize_ = EWOMS_GET_PARAM(TypeTag, Scalar, InitialTimeStepSize);
-
+        assert(timeStepSize_>0);
         const std::string& predetTimeStepFile =
             EWOMS_GET_PARAM(TypeTag, std::string, PredeterminedTimeStepsFile);
         if (!predetTimeStepFile.empty()) {
@@ -400,7 +400,12 @@ public:
      * \param timeStepSize The new value for the time step size \f$\mathrm{[s]}\f$
      */
     void setTimeStepSize(Scalar value)
-    { timeStepSize_ = value; }
+    {
+        if( not(value>0)){
+            std::cout << "Time step set to zero" << std::endl;
+        }
+        timeStepSize_ = value;
+    }
 
     /*!
      * \brief Set the current time step index to a given value.
@@ -443,7 +448,7 @@ public:
      */
     bool finished() const
     {
-        assert(timeStepSize_ >= 0.0);
+        assert(timeStepSize_ > 0.0);
         Scalar eps =
             std::max(Scalar(std::abs(this->time())), timeStepSize())
             *std::numeric_limits<Scalar>::epsilon()*1e3;
--- a/ewoms/common/simulator.hh
+++ b/ewoms/common/simulator.hh
@@ -401,10 +401,7 @@ public:
      */
     void setTimeStepSize(Scalar value)
     {
-        if( not(value>0)){
-            std::cout << "Time step set to zero" << std::endl;
-        }
-        timeStepSize_ = value;
+         timeStepSize_ = value;
     }
 
     /*!
@@ -448,7 +445,7 @@ public:
      */
     bool finished() const
     {
-        assert(timeStepSize_ > 0.0);
+        assert(timeStepSize_ >= 0.0);
         Scalar eps =
             std::max(Scalar(std::abs(this->time())), timeStepSize())
             *std::numeric_limits<Scalar>::epsilon()*1e3;
@@ -754,7 +751,7 @@ public:
                 else
                     // ask the problem to provide the next time step size
                     dt = std::min(maxTimeStepSize(), problem_->nextTimeStepSize());
-
+                assert(dt>0);
                 setTimeStepSize(dt);
             }
             prePostProcessTimer_.stop();
--- a/ewoms/disc/common/fvbaselocalresidual.hh
+++ b/ewoms/disc/common/fvbaselocalresidual.hh
@@ -579,8 +579,10 @@ protected:
 
             // Use the implicit Euler time discretization
             for (unsigned eqIdx = 0; eqIdx < numEq; ++eqIdx) {
+                double dt = elemCtx.simulator().timeStepSize();
+                assert(dt>0);
                 tmp[eqIdx] -= tmp2[eqIdx];
-                tmp[eqIdx] *= scvVolume / elemCtx.simulator().timeStepSize();
+                tmp[eqIdx] *= scvVolume / dt; 
 
                 residual[dofIdx][eqIdx] += tmp[eqIdx];
             }
--- a/ewoms/models/blackoil/blackoilenergymodules.hh
+++ b/ewoms/models/blackoil/blackoilenergymodules.hh
@@ -376,7 +376,7 @@ public:
         // and the thermal condictivity coefficients
         for (int phaseIdx = 0; phaseIdx < numPhases; ++ phaseIdx) {
             if (!FluidSystem::phaseIsActive(phaseIdx)) {
-                fs.setEnthalpy(phaseIdx, 0.0);
+                //fs.setEnthalpy(phaseIdx, 0.0);
                 continue;
             }
 
--- a/ewoms/models/blackoil/blackoilenergymodules.hh
+++ b/ewoms/models/blackoil/blackoilenergymodules.hh
@@ -376,7 +376,6 @@ public:
         // and the thermal condictivity coefficients
         for (int phaseIdx = 0; phaseIdx < numPhases; ++ phaseIdx) {
             if (!FluidSystem::phaseIsActive(phaseIdx)) {
-                //fs.setEnthalpy(phaseIdx, 0.0);
                 continue;
             }
 
--- a/ewoms/models/blackoil/blackoilintensivequantities.hh
+++ b/ewoms/models/blackoil/blackoilintensivequantities.hh
@@ -124,9 +124,13 @@ public:
 
         // extract the water and the gas saturations for convenience
         Evaluation Sw = 0.0;
-        if (waterEnabled)
-            Sw = priVars.makeEvaluation(Indices::waterSaturationIdx, timeIdx);
-
+        if (waterEnabled){
+            if(priVars.primaryVarsMeaning() == PrimaryVariables::px){
+                Sw = 1.0;
+            }else{
+                Sw = priVars.makeEvaluation(Indices::waterSaturationIdx, timeIdx);
+            }
+        }
         Evaluation Sg = 0.0;
         if (compositionSwitchEnabled)
         {
@@ -251,8 +255,8 @@ public:
             else
                 fluidState_.setRv(0.0);
         }
-        else {
-            assert(priVars.primaryVarsMeaning() == PrimaryVariables::Sw_pg_Rv);
+        else if (priVars.primaryVarsMeaning() == PrimaryVariables::Sw_pg_Rv) {
+            //assert(priVars.primaryVarsMeaning() == PrimaryVariables::Sw_pg_Rv);
 
             const auto& Rv = priVars.makeEvaluation(Indices::compositionSwitchIdx, timeIdx);
             fluidState_.setRv(Rv);
@@ -271,6 +275,8 @@ public:
             }
             else
                 fluidState_.setRs(0.0);
+        }else{
+            assert(priVars.primaryVarsMeaning() == PrimaryVariables::px);
         }
 
         typename FluidSystem::template ParameterCache<Evaluation> paramCache;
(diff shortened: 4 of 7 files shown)


totto82:
Good initiative. Have you done the adaptation in the well model for this?

ewoms/common/simulator.hh
```diff
@@ -125,7 +125,7 @@ public:
         time_ = 0.0;
         endTime_ = EWOMS_GET_PARAM(TypeTag, Scalar, EndTime);
         timeStepSize_ = EWOMS_GET_PARAM(TypeTag, Scalar, InitialTimeStepSize);
-
+        assert(timeStepSize_>0);
```
andlaus: It's probably better to up the game here and throw an exception. (also, isn't this redundant with the change to `Simulator::setTimeStepSize()` from below?)

ewoms/common/simulator.hh
```diff
@@ -400,7 +400,12 @@ public:
      * \param timeStepSize The new value for the time step size \f$\mathrm{[s]}\f$
      */
     void setTimeStepSize(Scalar value)
-    { timeStepSize_ = value; }
+    {
+        if( not(value>0)){
```
andlaus: weird white spacing. Also use an exclamation mark instead of `not`...

ewoms/common/simulator.hh
```diff
@@ -400,7 +400,12 @@ public:
...
+            std::cout << "Time step set to zero" << std::endl;
```
andlaus: throw an exception (e.g.  `std::runtime_error`) here instead...

ewoms/common/simulator.hh
```diff
@@ -443,7 +448,7 @@ public:
      */
     bool finished() const
     {
-        assert(timeStepSize_ >= 0.0);
+        assert(timeStepSize_ > 0.0);
```
andlaus: the equals is actually there for a reason here: when writing the initial solution, the time step size is zero and if the simulator is finished before it starts, having a zero-length time step here is legitimate. (also: this function is not used by `flow`!?)

ewoms/disc/common/fvbaselocalresidual.hh
```diff
@@ -579,8 +579,10 @@ protected:
 
             // Use the implicit Euler time discretization
             for (unsigned eqIdx = 0; eqIdx < numEq; ++eqIdx) {
+                double dt = elemCtx.simulator().timeStepSize();
+                assert(dt>0);
```
andlaus: the assert is pretty redundant if you make sure that this does not happen in `setTimeStepSize()`?

ewoms/models/blackoil/blackoilenergymodules.hh
```diff
@@ -376,7 +376,7 @@ public:
         // and the thermal condictivity coefficients
         for (int phaseIdx = 0; phaseIdx < numPhases; ++ phaseIdx) {
             if (!FluidSystem::phaseIsActive(phaseIdx)) {
-                fs.setEnthalpy(phaseIdx, 0.0);
+                //fs.setEnthalpy(phaseIdx, 0.0);
```
andlaus: please do not litter code. (remove the line outright. I'm wondering why it was there in the first place...)

ewoms/models/blackoil/blackoilintensivequantities.hh
```diff
@@ -198,9 +202,13 @@ public:
         // update the Saturation functions for the blackoil solvent module.
         asImp_().solventPostSatFuncUpdate_(elemCtx, dofIdx, timeIdx);
 
-        const Evaluation& SoMax =
-            Opm::max(fluidState_.saturation(oilPhaseIdx),
-                     elemCtx.problem().maxOilSaturation(globalSpaceIdx));
+        Evaluation SoMax=0;
+        //const Evaluation&
```
andlaus: remove dead code.

ewoms/models/blackoil/blackoilintensivequantities.hh
```diff
@@ -274,11 +282,15 @@ public:
             }
             else
                 fluidState_.setRs(0.0);
+        }else{
```
andlaus: line break is inconsistent with the ones above.

ewoms/models/blackoil/blackoilintensivequantities.hh
```diff
@@ -254,8 +262,8 @@ public:
             else
                 fluidState_.setRv(0.0);
         }
-        else {
-            assert(priVars.primaryVarsMeaning() == PrimaryVariables::Sw_pg_Rv);
+        else if (priVars.primaryVarsMeaning() == PrimaryVariables::Sw_pg_Rv) {
+            //assert(priVars.primaryVarsMeaning() == PrimaryVariables::Sw_pg_Rv);
```
andlaus: remove dead code.

ewoms/models/blackoil/blackoilintensivequantities.hh
```diff
@@ -274,11 +282,15 @@ public:
...
                 fluidState_.setRs(0.0);
+        }else{
+            assert(priVars.primaryVarsMeaning() == PrimaryVariables::px);
         }
 
         typename FluidSystem::template ParameterCache<Evaluation> paramCache;
         paramCache.setRegionIndex(pvtRegionIdx);
-        paramCache.setMaxOilSat(SoMax);
+        if(FluidSystem::phaseIsActive(FluidSystem::oilPhaseIdx)){
+            paramCache.setMaxOilSat(SoMax);
```
andlaus: this should not matter because the `paramCache` always has memory for the maximum oil saturation, but the `if` does not hurt either IMO...

ewoms/models/blackoil/blackoilintensivequantities.hh
```diff
@@ -335,7 +347,16 @@ public:
         Scalar rockCompressibility = problem.rockCompressibility(elemCtx, dofIdx, timeIdx);
         if (rockCompressibility > 0.0) {
             Scalar rockRefPressure = problem.rockReferencePressure(elemCtx, dofIdx, timeIdx);
-            Evaluation x = rockCompressibility*(fluidState_.pressure(oilPhaseIdx) - rockRefPressure);
+            Evaluation x;
+            if(FluidSystem::phaseIsActive(oilPhaseIdx)){
+                x = rockCompressibility*(fluidState_.pressure(oilPhaseIdx) - rockRefPressure);
+            }else{
+                if(FluidSystem::phaseIsActive(waterPhaseIdx)){
```
andlaus: please make this a flat `if () else if ()` chain...

ewoms/models/blackoil/blackoilonephaseindices.hh
```diff
@@ -0,0 +1,153 @@
...
+#include <cassert>
+
+namespace Ewoms {
+
+/*!
+ * \ingroup BlackOilModel
+ *
+ * \brief The primary variable and equation indices for the black-oil model.
+ */
+template <unsigned numSolventsV, unsigned numPolymersV, unsigned numEnergyV, unsigned PVOffset, unsigned enableCanonicalCompIdx>
```
andlaus: rename `enableCanonicalCompIdx` to `canonicalCompIdx`. (it is not a boolean!)

ewoms/models/blackoil/blackoilonephaseindices.hh
```diff
@@ -0,0 +1,153 @@
...
+struct BlackOilOnePhaseIndices
+{
+    //! Is phase enabled or not
+    static const bool oilEnabled = (enableCanonicalCompIdx == 0);
```
andlaus: hardcoding phase/component indices is not a terribly good idea. (cf. e.g. `ebos_altidx`)

ewoms/models/blackoil/blackoilonephaseindices.hh
```diff
@@ -0,0 +1,153 @@
...
+        enableEnergy ? PVOffset + numPhases + numSolvents + numPolymers : - 1000;
+
+    //////////////////////
+    // Equation indices
+    //////////////////////
+
+    //! \brief returns the index of "active" component
+    static unsigned canonicalToActiveComponentIndex(unsigned compIdx)
+    {
+        return 0;
```
andlaus: some `assert()` would be good...

ewoms/models/blackoil/blackoilonephaseindices.hh
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    static unsigned activeToCanonicalComponentIndex(unsigned compIdx)
+    {
+        // assumes canonical oil = 0, water = 1, gas = 2;
+        assert(compIdx < 2);
```
andlaus: `assert(compIdx == 0)`

ewoms/models/blackoil/blackoilonephaseindices.hh
```diff
@@ -0,0 +1,153 @@
...
+            return 1;
+        } else {
+            assert(oilEnabled);
+        }
+        return 0;
+    }
+
+    //! Index of the continuity equation of the first phase
+    static const int conti0EqIdx = PVOffset + 0;
+    // one continuity equation follows
```
andlaus: no: no continuity equation follows ;)

ewoms/models/blackoil/blackoilprimaryvariables.hh
```diff
@@ -109,6 +109,7 @@ public:
         Sw_po_Sg, // threephase case
         Sw_po_Rs, // water + oil case
         Sw_pg_Rv, // water + gas case
+        px      , // onephase case
```
andlaus: That's a pretty bad name, IMO. Also: aligning ascii-art style aligning should be avoided  in eWoms...

ewoms/models/blackoil/blackoilprimaryvariables.hh
```diff
@@ -284,7 +288,16 @@ public:
         }
 
         // assign the actual primary variables
-        if (primaryVarsMeaning() == Sw_po_Sg) {
+        if (primaryVarsMeaning() == px ) {
+            if (waterEnabled){
+                (*this)[waterSaturationIdx] = FsToolbox::value(fluidState.saturation(waterPhaseIdx));
+                (*this)[pressureSwitchIdx] = FsToolbox::value(fluidState.pressure(waterPhaseIdx));
```
andlaus: is there a reason why water is used? (note that the fluid state might throw if water is not active.)

ewoms/models/blackoil/blackoilratevector.hh
```diff
@@ -81,16 +81,16 @@ public:
     BlackOilRateVector(Scalar value) : ParentType(Toolbox::createConstant(value))
     {}
 
-    template <class Eval = Evaluation>
-    BlackOilRateVector(const typename std::enable_if<std::is_same<Eval, Evaluation>::value, Evaluation>::type& value) : ParentType(value)
-    {}
+    // template <class Eval = Evaluation>
+    // BlackOilRateVector(const typename std::enable_if<std::is_same<Eval, Evaluation>::value, Evaluation>::type& value) : ParentType(value)
+    // {}
```
andlaus: dead code.

ewoms/models/blackoil/blackoilratevector.hh
```diff
@@ -81,16 +81,16 @@ public:
...
 
     /*!
      * \copydoc ImmiscibleRateVector::ImmiscibleRateVector(const
      * ImmiscibleRateVector& )
      */
-    BlackOilRateVector(const BlackOilRateVector& value) : ParentType(value)
-    {}
+    // BlackOilRateVector(const BlackOilRateVector& value) : ParentType(value)
+    // {}
```
andlaus: dead code.

also: there must have been a reason why these two constructors were around. does everything still compile without them and if yes, have you checked that the performance does not suffer? (maybe they were just required due to a bug in an older compiler.)

ewoms/common/simulator.hh
hnil: The problem is that the curren code requite stting timeStepSize_=0. I think it is only used to write out state at start of time step.

ewoms/common/simulator.hh
hnil: value=0 is use to write out state at start of timestep.

ewoms/common/simulator.hh
hnil: Ok. I will remove it. the other assert will catch if it is a problem.

ewoms/models/blackoil/blackoilonephaseindices.hh
hnil: I did not understand this point.

ewoms/models/blackoil/blackoilprimaryvariables.hh
hnil: sudjestions?

ewoms/models/blackoil/blackoilonephaseindices.hh
hnil: just following the current code..

ewoms/common/simulator.hh
andlaus: I see. Does this also happen for `ebos`? If no, I guess this would be a good excuse to fix `flow`...

ewoms/models/blackoil/blackoilonephaseindices.hh
hnil: done

ewoms/models/blackoil/blackoilonephaseindices.hh
andlaus: in a single-phase model, you only have one continuity equation, so the "one continuity equation follows" comment is wrong.

ewoms/models/blackoil/blackoilprimaryvariables.hh
andlaus: `OnePhase_p` ? The other names should probably be prefixed by `ThreePhase_` then...

ewoms/models/blackoil/blackoilprimaryvariables.hh
hnil: I inted the code to work with all phases. However tests is done only with water so far. It is only water which is make primary variable if settings for one phase is proper. for the others the values are set without derivatives. I have added asserts for the gas case in case some one use onephase with compoistion switch on ...

ewoms/common/simulator.hh
hnil: removed the lines. Hopefully the other asserts catch the problem.

atgeirr:
Replaced by #530.

PR #511, opened 2019-05-22
Add FemSolverBackend and AMGXSolverBackend.

dr-robertk:
This PR adds a FemSolverBackend for using linear solvers implemented in dune-fem, including everything PETSc has to offer and an AMGXSolverBackend for the AMGXSolver.


dr-robertk:
@andlaus: Hello Herr Dr. andlaus, could you take a look a the part in fvbasdiscretization.hh and sugest a better way to select the different sparse matrix adapter options and solvers?

dr-robertk:
@andlaus: PS: no bike shedding plz ;-)

andlaus:
Generally, I'm in favor of this. When it comes to the concrete implementation, I'm wondering quite a bit if it is really necessary that so many files which have nothing to do with the linear solver are affected... (I haven't yet looked at the code in depth, or tried to use it, though.) 

linear-solver-selection mechanism: create a new type tag for each of these classes, set all the Properties which you need on this type tag and finally splice it into your problem's type tag using the `LinearSolverSplice`. you can do this completely analogous to e.g. the standard ISTL-AMG linear solver backend:

type tag definition: https://github.com/OPM/ewoms/blob/master/ewoms/linear/parallelamgbackend.hh#L50-L66
usage: https://github.com/OPM/ewoms/blob/master/tests/problems/co2injectionproblem.hh#L145

for the nitpicking: I promise not care about coding style consistency during review of this PR if you vow to support the subsequent cleanup PR...

blattms:
So dune-fem is mandatory to use any of this?

dr-robertk:
@blattms: dune-fem contains a couple of solver interface that include bindings for PETSc. dune-fem was already an optional dependency for ewoms, so that seemed like a good solution to avoid re-coding the complete bindings. These bindings to PETSc allow to access solvers provided by PETSc but are also needed for the AMGXWrapper (https://github.com/barbagroup/AmgXWrapper) which provides an interface to AMGX using PETSc. I think eventually this wrapper will become part of PETSc. 

In addition, the use of this solvers allowed to design the SparseMatrixAdapter in ewoms and flow to allow for different implementations of sparse matrix data structures and linear solver packages without changing the assembly code and without copying of matrices. 

Parts of these developments are already merged and with this PR I'm just putting out the rest that was missing.

blattms:
that is a lot of depencndencies (dune-fem, AMGXWrapper, Petsc, AMGX) just to use AMGX. Probably too much for me.

alfbr:
Yes, AMGX is still quite painful to get going. We have a dialogue with Nvidia on it, hopefully they can help us make it easier to use AMGX. The build issues are challenging at this point. I suggest we do a video meet-up to discuss a way forward. I am very grateful or all the work done in this PR, it is a very good starting point for exploring the potential in GPUs.

dr-robertk:
> that is a lot of depencndencies (dune-fem, AMGXWrapper, Petsc, AMGX) just to use AMGX. Probably too much for me.

I agree. That is why this is and should stay an optional dependency.

hnil:
This will be very good!! Merge it as soon as possible so we do not miss it. Together with the more flexible dune interface and good? cpr implementations this will make linear solver possibilities much more complete. I would be interested in setting up the similar cpr preconditioner in petsc for comparison...

atgeirr:

atgeirr:
The tests fail because "ebos/femcpgridcompat.hh" no longer exists in ewoms. The ebos dir was moved to opm-simulators some time ago. Perhaps that file should be moved back to ewoms (evidently it is not only used for ebos/flow)? Or maybe another solution is more appropriate?

dr-robertk:
@atgeirr: I removed the ebos/femcpgridcompat.hh since the code seems to be compiling without it.

dr-robertk:

ewoms/nonlinear/newtonmethod.hh
```diff
@@ -371,9 +371,12 @@ public:
                 asImp_().linearizeAuxiliaryEquations_();
                 linearizeTimer_.stop();
 
+                // finalize the linearization process if not done already
+                linearizer.finalize();
```
atgeirr: This looks like a potentially important change?

ewoms/disc/common/fvbasediscretization.hh
```diff
@@ -130,7 +136,66 @@ SET_TYPE_PROP(FvBaseDiscretization, DiscExtensiveQuantities, Ewoms::FvBaseExtens
...
+    typedef Dune::Fem::PetscLinearOperator< DiscreteFunction, DiscreteFunction > LinearOperator;
+#elif USE_DUNE_FEM_VIENNACL_SOLVERS
+#warning "Using Dune-Fem ViennaCL solvers"
+    typedef Dune::Fem::SparseRowLinearOperator < DiscreteFunction, DiscreteFunction > LinearOperator;
+#else
+#warning "Using Dune-Fem ISTL solvers"
+    typedef Dune::Fem::ISTLLinearOperator < DiscreteFunction, DiscreteFunction > LinearOperator;
+#endif
+
+    struct FemMatrixBackend : public LinearOperator
```
atgeirr: It is odd to place this code in the fvbasediscretization.hh file, why is it here?

ewoms/nonlinear/newtonmethod.hh
dr-robertk: For the ISTL backend this calls compress of matrix if implicit build mode is used and for PETSc it calls finalize assembly. For the current use of ISTL this function does nothing.

ewoms/disc/common/fvbasediscretization.hh
dr-robertk: I agree. What is a better place and way to select the different solvers?

andlaus: > What is a better place and way to select the different solvers?

the problem. [instructions](https://github.com/OPM/ewoms/pull/511#issuecomment-494867000).

andlaus:
ok: I tried to give this a whirl, but I failed miserably because it seems like a lot of dune-fem infrastructure is simply not there in the 2.6 branch. does this only work for the latest DUNE and dune-fem masters?

dr-robertk:
@andlaus: Yes everything master at the latest version. You could have asked but I guess it was to much to ask for ;-)

dr-robertk:
@atgeirr: I fixed the linearizer.finalize(); which actually should go inside the linearizer timer section.

dr-robertk:

dr-robertk:
@andlaus: Like the ISTLMatrixBackend where should I put the code from fvbasediscretization to make the changes in fvbasediscretization only a few lines?

andlaus:
> @andlaus: Yes everything master at the latest version. You could have asked but I guess it was to much to ask for ;-)

okay, before this can be merged, it needs to compile with Dune 2.5 and 2.6, IMO. (it needs to *compile* but does not necessarily need to *work* with them.)

> @andlaus: Like the ISTLMatrixBackend where should I put the code from fvbasediscretization to make the changes in fvbasediscretization only a few lines?

I think the `FemSolverBackend` class should create the `DiscreteFunction` and `DiscreteFunctionSpace` objects internally on a as-needed basis. Since the grid has a sequence number already, this is not going to be less efficient.

If you want, I can provide you with a patch to that end.

dr-robertk:
It should build with 2.5 and 2.6, I tested 2.5 and the build worked yesterday. So don't know what went wrong on your side.

andlaus:
> It should build with 2.5 and 2.6

does it also compile if dune-fem is detected by the build system and if some tests activates the dune-fem solver backend if it is available?

dr-robertk:
So you are asking whether the code compiles with the 2.5 release of dune-fem when it needs things that are only implemented in the current master. Without testing I can answer: NO! But I may need to add some version checks here and there.

andlaus:
no, I'm asking if if compiles if (a) dune and dune-fem are available and on a release branch which we ought to support and (b) some test in eWoms has been modified to use the new fem linear solver backend. in the case of (b) the test should probably fall back to the standard backend if dune-fem is not on master...

dr-robertk:
Ok maybe I make it separate: The fem-solver backend needs the dune-fem master.

dr-robertk:
This means that using dune-fem 2.5 and wanting to use the fem solver backend will not work. Does this answer your question?

dr-robertk:
So to conclude: I will add some DUNE_VERSION... checks to make clear, which version is to used.

andlaus:
I mean I want grid apdaptivity to continue working with dune-2.5 and I want some test to excercise the new code if dune-fem > 2.6 is available.

> So to conclude: I will add some DUNE_VERSION... checks to make clear, which version is to used.

yes, that will do the trick...

dr-robertk:
Just to confirm, the tests compile with with dune 2.5 and dune-fem found without using the fem solvers.

andlaus:
> Just to confirm, the tests compile with with dune 2.5 and dune-fem found without using the fem solvers.

ok, but it seems like there is nothing in `make test-suite` which uses the new code?!

dr-robertk:
Yes, because first I wanted to get all the changes in place without disturbing the existing stuff. Once that is done I will introduce an example.

andlaus:
I do not want dangling code in the repository because past experience has shown that it *always* bit-rots. (@akva2: how much effort would it be to do nightly coverage runs on jenkins?)

anyway, modifying one of the tests to use each of the new backend if it is available should not be too much effort if you go with the splicing mechanism.

andlaus:
okay, I fixed most of the issues I'm bugged with in this PR in dr-robertk/ewoms#1. Note that the Petsc and AMG-X backends almost certainly will not work with that. (I have little motivation in getting these prerequisites working.)

andlaus:
During doing dr-robertk/ewoms#1, I've noticed that these new backends currently produce a shipload of compiler warnings on GCC 9.1.1. (No, I did not enable masochistic warning flags.) These need to be fixed before merge.

akva2:
@andlaus coverage (gcov) is run in the static analysis jobs now, e.g. https://ci.opm-project.org/job/opm-grid-static-analysis/

dr-robertk:
@blattms: I would like this to be merged before the release, if possible. I'll fix the conflicts.

blattms:
In that case you would need to resolve the merge conflicts, make the tests pass and get it merged into master. (Seem like quite some work to me.)

Judging from the conversation it seems that both Petsc and AMGXSolver will only work if DUNE < 2.6 (i.e. master) including dune-fem is available. Is that correct.

dr-robertk:
I actually did that already, just the PETSc stuff. It's in a different branch. I'll try to fix this on Wednesday.

blattms:
> I actually actually did that already, just the PETSc stuff. It's in a different branch.

That might very well be the case, but there is no PR for it or I could not find it.

dr-robertk:
I pushed the updated version. I'll assess whether it's a good idea to put this into the release or not.

dr-robertk:
@blattms: There is a problem with some parameters and I won't get this ready until the release. So lets do it afterwards.

PR #513, opened 2019-05-28
let the linear solver backend determine the type of the sparse matrix adapter

andlaus:
so far, this property was specified by the level of the base discretization. because explicitly set properties overwrite the stuff inherited from splices, so far it was not possible to specify linear solver backends which require a custom sparse matrix adapter unless the sparse matrix adapter class was explicitly set again at the problem level -- thus making the splice much less generic than it ought to be. (also, the whole point of the sparse matrix adapter abstraction is that the discretization does not need to know what the sparse matrix code does internally.)

### DIFF
--- a/ewoms/disc/common/fvbasediscretization.hh
+++ b/ewoms/disc/common/fvbasediscretization.hh
@@ -50,7 +50,6 @@
 #include <ewoms/parallel/gridcommhandles.hh>
 #include <ewoms/parallel/threadmanager.hh>
 #include <ewoms/linear/nullborderlistmanager.hh>
-#include <ewoms/linear/istlsparsematrixadapter.hh>
 #include <ewoms/common/simulator.hh>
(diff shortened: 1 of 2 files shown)


andlaus:
jenkins build this please

dr-robertk:
Makes absolutely sense. I was going to introduce this into my PR. So now I can just rebase on top of this.

andlaus:
feel free to merge once jenkins is done with its stuff...

andlaus:
the jenkins server seems to have some problems. ("HTTP ERROR 404; not found?!". are you aware @akva2 ?). maybe it is just a heisenbug:

andlaus:
jenkins build this please

andlaus:
jenkins build this opm-simulators=1868 please

andlaus:
jenkins build this opm-simulators=1868 please

PR #514, opened 2019-06-03
parameter system: some fixes and improvements

andlaus:
this are "drive by" fixes while I was working on something else last week;

- make code which uses the `EWOMS_PARAM_IS_SET` macro compile
- make it possible to reset the parameter machinery. this is an  advanced function which might eat children! only use it if you know  what you're doing!


andlaus:
jenkins build this please

totto82:
Shouldn't change anything for the current code, and adds needed functionality.

PR #516, opened 2019-06-07
first attempt to fixing the Evaluation with value 1.0

GitPaean:
@andlaus , how do you think?

### DIFF
--- a/ewoms/models/blackoil/blackoilpolymermodules.hh
+++ b/ewoms/models/blackoil/blackoilpolymermodules.hh
@@ -845,14 +845,19 @@ public:
         const Scalar eps = 1e-14;
         // return 1.0 if the polymer has no effect on the water.
         if (std::abs((viscosityMultiplier - 1.0)) < eps){
-            return ToolboxLocal::createBlank(v0) + 1.;
+            Evaluation v = v0;
+            v = 1.;
+            return v;
         }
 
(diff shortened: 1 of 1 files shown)


GitPaean:
jenkins build this please

GitPaean:
jenkins build this please

andlaus:
I see what you mean, this is not exactly pretty. how about handling `DynamicEvaluations` with zero derivatives as constants. This would require a few case distinctions, but would allow to instantiate `DynamicEvaluation`s that represent  constants without having to specify a size...

atgeirr:
> I see what you mean, this is not exactly pretty. how about treating `DynamicEvaluations` with zero derivatives as constants? This would require a few case distinctions, but would allow to instantiate `DynamicEvaluation`s that represent constants without having to specify a size...

I suggested this a while ago, and still think it is a good idea, so I support that!

GitPaean:
> DynamicEvaluations with zero derivatives as constants? 

Sorry I did not see your comment yesterday.  Then we will support a constructor that can create a DynamicEvaluation from a Scalar value, then it is possible we will recover some old grammar before introduced DynamicEvaluation.  The DynamicEvalution makes us have to specify the number of the primary variables at its current form.

andlaus:
this has become obsolete with #517. closing.

PR #517, opened 2019-06-13
using createConstantOne instead of createBlank

GitPaean:
to obtain more well defined behavoir

PR #518, opened 2019-06-13
do no longer call createVariable() for plain floating point types

andlaus:
with OPM/opm-material#340, this will throw an exception. Never try to fix more than one thing at once...

Right now, this PR can be merged as-is as it is, and it is IMO a good idea to do so, because it is IMO good style to not to give the impression that a plain floating point object can represent an AD variable. It is required before for OPM/opm-material#340 can go in.

### DIFF
--- a/ewoms/disc/common/fvbaseprimaryvariables.hh
+++ b/ewoms/disc/common/fvbaseprimaryvariables.hh
@@ -89,10 +89,15 @@ public:
      */
     Evaluation makeEvaluation(unsigned varIdx, unsigned timeIdx) const
     {
-        if (timeIdx == 0)
-            return Toolbox::createVariable((*this)[varIdx], varIdx);
-        else
-            return Toolbox::createConstant((*this)[varIdx]);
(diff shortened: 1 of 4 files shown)


andlaus:
jenkins build this please

andlaus:
note that, ironically, this issue affected only the non-AD code paths, i.e., everything in opm-simulators compiles with just OPM/opm-material#340 merged.

andlaus:
because #517 was merged by @atgeirr, this PR and OPM/opm-material#340 are now mutually dependent...

andlaus:
jenkins build this opm-material=340 please

GitPaean:
> Never try to fix more than one thing at once...

This is pure golden wisdom.

atgeirr:
jenkins build this opm-material=340 please

atgeirr:
I want a new jenkins run for this after merging other things, but Jenkins seems not to take. Trying again...

atgeirr:
Merging, all green.

PR #520, opened 2019-06-17
simulator: make episode{Is,WillBe}Over() work if startTime() is negative

andlaus:
for some reason, this happens with SPE-10 model 2.


andlaus:
jenkins build this please

totto82:
Thanks for the fix. This may also fix a model I have been testing. Does not affect Flow.

PR #522, opened 2019-06-24
improve and generalize the time integration

andlaus:
the time integration code now tries to update once with exactly the minumum time step size before bailing out, the error message when aborting the run because of encountering a time step size smaller than the minimum now makes sense, and the parameters for the time integration are not fixed anymore but queried from the implementation of the problem, i.e., they can change over the course of the simulation.

### DIFF
--- a/ewoms/disc/common/fvbaseproblem.hh
+++ b/ewoms/disc/common/fvbaseproblem.hh
@@ -509,28 +509,25 @@ public:
      */
     void timeIntegration()
     {
-        unsigned maxFails = EWOMS_GET_PARAM(TypeTag, unsigned, MaxTimeStepDivisions);
-        Scalar minTimeStepSize = EWOMS_GET_PARAM(TypeTag, Scalar, MinTimeStepSize);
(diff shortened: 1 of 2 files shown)


andlaus:
jenkins build this please

totto82:
In general I think this PR is fine. But note that our interpretation of minimum time step size may be different than what the engineers expects. In Eclipse I think it means that the simulation continues whatever residual if the time-step is small enough. We instead give up.

andlaus:
> In Eclipse I think it means that the simulation continues whatever residual if the time-step is small enough.

hm. I think we should continue to guarantee that the requested tolerance is obtained by default, but possibly introduce a parameter that instructs the simulator to continue anyway?

andlaus:
> In Eclipse I think it means that the simulation continues whatever residual if the time-step is small enough.

hm. I think we should continue to guarantee that the requested tolerance is obtained by default, but possibly introduce a parameter that instructs the simulator to continue anyway?

totto82:
Lets to the extra parameter in another PR.

PR #524, opened 2019-06-26
Ewoms::start(): abort the run when encountering unknown parameters

andlaus:
This patch is part of the quest to reduce surprising/unexpected behaviors and it aborts the simulation with a hopefully useful error message if unknown parameters are specified at runtime. The implementation of this is general, i.e. it works for any simulators that use `Ewoms::start()` without any scaffolding around it: with no modifications to opm-simulators `mebos` will give up if it encounters a parameter that is unknown to the black-oil variant and it will print the usage message multiple times on `--help`. `flow` will work as before but it will start complaining about unknown parameters. OPM/opm-simulators#1916 rectifies these issues.

### DIFF
--- a/ewoms/common/start.hh
+++ b/ewoms/common/start.hh
@@ -118,7 +118,11 @@ static inline void registerAllParameters_(bool finalizeRegistration = true)
  * \param argv Array with the command line argument strings
  */
 template <class TypeTag>
-static inline int setupParameters_(int argc, const char **argv, bool registerParams = true)
+static inline int setupParameters_(int argc,
+                                   const char **argv,
+                                   bool registerParams=true,
+                                   bool allowUnused=false,
+                                   bool handleHelp = true)
 {
     typedef typename GET_PROP_TYPE(TypeTag, Problem) Problem;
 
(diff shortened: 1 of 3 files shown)


andlaus:
jenkins build this please

ewoms/common/start.hh
```diff
@@ -171,6 +175,36 @@ static inline int setupParameters_(int argc, const char **argv, bool registerPar
...
+                std::cerr << "   " << keyValue.first << "=\"" << keyValue.second << "\"\n";
+            std::cerr << "\n";
+
+            std::cerr << "Use\n"
+                      << "\n"
+                      << "  " << argv[0] << " --help\n"
+                      << "\n"
+                      <<"to obtain the list of recognized command line parameters.\n\n";
+        }
+        return /*status=*/2;
```
blattms: Mmh, this will result in `start()` returning `0` at [start.hh#L310](https://github.com/OPM/ewoms/blob/f5f65f3871a2ace1d5098955aea617679baf096c/ewoms/common/start.hh#L310). I might be mistaking but I would assume that a return values of `0` indicates success

andlaus: fixed

andlaus:
jenkins build this please

andlaus:
> Should this make OPM/opm-simulators#1845 obsolete?

not in its current form: I took some pains not to change the behavior of `flow`, but `flow` can easily be modified to handle unused parameters using the generic code...

andlaus:
jenkins build this please

andlaus:
jenkins build this please

totto82:
Thanks for the patch.

PR #525, opened 2019-06-27
remove const qualifier on return value

akva2:
meaningless and causes warnings.


andlaus:
thanks. (BTW: what's the difference between nothing and constant nothing ;))

akva2:
A: only one of them makes gcc spew.

PR #528, opened 2019-07-04
Add foam module

atgeirr:
This adds a foam module, which should be considered experimental for the time being. It supports a subset of foam features: most importantly gas mobility reduction tables (FOAMMOB) and foam adsorption (FOAMADS, FOAMROCK).

A companion PR in opm-simulators is needed.

### DIFF
--- a/ewoms/models/blackoil/blackoilfoammodules.hh
+++ b/ewoms/models/blackoil/blackoilfoammodules.hh
@@ -32,8 +32,15 @@
 //#include <ewoms/io/vtkblackoilfoammodule.hh>
 #include <ewoms/models/common/quantitycallbacks.hh>
 
-//#include <opm/material/common/Tabulated1DFunction.hpp>
+#include <opm/material/common/Tabulated1DFunction.hpp>
 //#include <opm/material/common/IntervalTabulated2DFunction.hpp>
+
+#if HAVE_ECL_INPUT
+#include <opm/parser/eclipse/Deck/Deck.hpp>
+#include <opm/parser/eclipse/EclipseState/EclipseState.hpp>
+#include <opm/parser/eclipse/EclipseState/Tables/FoamadsTable.hpp>
+#include <opm/parser/eclipse/EclipseState/Tables/FoammobTable.hpp>
+#endif
 
 #include <opm/material/common/Valgrind.hpp>
 #include <opm/material/common/Unused.hpp>
@@ -68,13 +75,11 @@ class BlackOilFoamModule
 
     typedef Opm::MathToolbox<Evaluation> Toolbox;
 
-    //typedef typename Opm::Tabulated1DFunction<Scalar> TabulatedFunction;
-    //typedef typename Opm::IntervalTabulated2DFunction<Scalar> TabulatedTwoDFunction;
+    typedef typename Opm::Tabulated1DFunction<Scalar> TabulatedFunction;
 
     static constexpr unsigned foamConcentrationIdx = Indices::foamConcentrationIdx;
     static constexpr unsigned contiFoamEqIdx = Indices::contiFoamEqIdx;
     static constexpr unsigned gasPhaseIdx = FluidSystem::gasPhaseIdx;
-
 
     static constexpr unsigned enableFoam = enableFoamV;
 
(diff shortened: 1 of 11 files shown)


andlaus:
sorry, I can't properly review this anymore, i.e., @totto82 is in charge.

one question, though: isn't this supposed to be an experimental feature; at least initially?

atgeirr:
> one question, though: isn't this supposed to be an experimental feature; at least initially?

Yes it has not been tested (much). Maybe add a warning?

andlaus:
> Yes it has not been tested (much). Maybe add a warning?

that would be one option. the IMO better alternative is to throw an `std::runtime_error` somewhere in the initialization if foam is requested by the deck and the `EnableExperiments` property is not `true`...

atgeirr:
> the `EnableExperiments` property

Will not this also cause other experimental things to be activated, that could change behaviour?

andlaus:
> Will not this also cause other experimental things to be activated, that could change behaviour?

yes, the other experimental stuff might lead slightly different behavior in some cases. that's kind of the point of the `EnableExperiments` property, isn't it?

in case you're wondering: yes, declaring foam experimental would deactivate it for `flow`, but encouraging people to use it for their production decks seems like a *very* bad idea to me at the current stage of development. That said, if you want to bear the grunt of supporting these people for the stuff that will almost inevitably blow up, I also don't mind if you decide to declare this feature as "production ready".

atgeirr:
My point is that I do not necessarily want the other experimental features on at the same time as running/testing foam. Perhaps the best is then a new parameter, `EnableFoam` defaulting to false, so if you want to run a foam case you need to be explicit about it?

ewoms/models/blackoil/blackoilfoammodules.hh
```diff
@@ -0,0 +1,598 @@
...
+     */
+    static void registerOutputModules(Model& model,
+                                      Simulator& simulator)
+    {
+        if (!enableFoam)
+            // foam have been disabled at compile time
+            return;
+
+        //model.addOutputModule(new Ewoms::VtkBlackOilFoamModule<TypeTag>(simulator));
+    }
```
totto82: Maybe a warning that Vtk is not supported for foam

ewoms/models/blackoil/blackoilfoammodules.hh
```diff
@@ -0,0 +1,598 @@
...
+            // When in the future adding support for the functional
+            // model, FOAMMOB will not be required anymore (functional
+            // family of keywords can be used instead, FOAMFSC etc.).
+            throw std::runtime_error("FOAMMOB must be specified in FOAM runs");
+        }
+        if (numPvtRegions != foammobTables.size()) {
+            throw std::runtime_error("Inconsistent sizes, number of PVT regions differ from the "
+                                     "number of FOAMMOB tables.");
+        }
+
```
totto82: I am missing the FOAMOPTS keyword that specify the type of foam model to use. Currently we only support the default FOAMOPTS options. (GAS; TAB) The simulator should throw with a appropriate message if somebody tries to use non-default options for FOAMOPTS.

ewoms/models/blackoil/blackoilfoammodules.hh
```diff
@@ -0,0 +1,598 @@
...
+                               unsigned timeIdx)
+    {
+        const PrimaryVariables& priVars = elemCtx.primaryVars(dofIdx, timeIdx);
+        foamConcentration_ = priVars.makeEvaluation(foamConcentrationIdx, timeIdx);
+        const auto& fs = asImp_().fluidState_;
+
+        // Compute gas mobility reduction factor
+        Evaluation mobilityReductionFactor = 1.0;
+        if (false) { // TODO: allow this model
+            // The functional model is used.
```
totto82: What is missing?

ewoms/models/blackoil/blackoilfoammodules.hh
```diff
@@ -0,0 +1,598 @@
...
+            Evaluation F2 = pow((fm_oil-S_o)/(fm_oil-fl_oil), ep_oil);
+            Evaluation F3 = pow(fm_cap/Ca, ep_cap);
+            Evaluation F7 = 0.5 + atan(ep_dry*(S_w-fm_dry))/M_PI;
+
+            mobilityReductionFactor = 1./(1. + fm_mob*F1*F2*F3*F7);
+        } else {
+            // The tabular model is used.
+            const auto& gasMobilityMultiplier = FoamModule::gasMobilityMultiplierTable(elemCtx, dofIdx, timeIdx);
+            mobilityReductionFactor = gasMobilityMultiplier.eval(foamConcentration_, /* extrapolate = */ true);
+        }
```
totto82: Add a comment that modification due to velocity and pressure is missing? What happens if FOAMMOBS is in the deck?

ewoms/models/blackoil/blackoilindices.hh
```diff
@@ -46,6 +46,8 @@ struct BlackOilIndices
     static const bool waterEnabled = true;
     static const bool gasEnabled = true;
 
+    //static const bool foamEnabled = foamEnabled___;
+
     //! Are solvents involved?
```
totto82: remove this

ewoms/models/blackoil/blackoilfoammodules.hh
atgeirr: Primarily the ability to use water instead of gas as the transport phase.

ewoms/models/blackoil/blackoilfoammodules.hh
atgeirr: As written now it will be ignored. I'll add a warning and a comment.

atgeirr: Actually, the warning will not be added here, it is just a usual keyword not implemented case for FOAMMOBS and FOAMMOBP.

atgeirr:
I have now made the requested modifications (I think). I'll rerun jenkins from the opm-simulators PR.

PR #529, opened 2019-07-08
changed: quell some more unused parameter warnings without MPI

akva2:
missed originally as they are only emitted from clang, not gcc


akva2:

andlaus:
thanks

PR #530, opened 2019-08-12
Enable onephase revised

atgeirr:
This is a revised version of #510. Basically it is what I would have requested if reviewing it.

### DIFF
--- a/opm/models/blackoil/blackoilenergymodules.hh
+++ b/opm/models/blackoil/blackoilenergymodules.hh
@@ -376,7 +376,6 @@ public:
         // and the thermal condictivity coefficients
         for (int phaseIdx = 0; phaseIdx < numPhases; ++ phaseIdx) {
             if (!FluidSystem::phaseIsActive(phaseIdx)) {
-                fs.setEnthalpy(phaseIdx, 0.0);
                 continue;
             }
 
--- a/opm/models/blackoil/blackoilenergymodules.hh
+++ b/opm/models/blackoil/blackoilenergymodules.hh
@@ -376,7 +376,7 @@ public:
         // and the thermal condictivity coefficients
         for (int phaseIdx = 0; phaseIdx < numPhases; ++ phaseIdx) {
             if (!FluidSystem::phaseIsActive(phaseIdx)) {
-                fs.setEnthalpy(phaseIdx, 0.0);
+                //fs.setEnthalpy(phaseIdx, 0.0);
                 continue;
             }
 
--- a/opm/models/blackoil/blackoilenergymodules.hh
+++ b/opm/models/blackoil/blackoilenergymodules.hh
@@ -376,6 +376,7 @@ public:
         // and the thermal condictivity coefficients
         for (int phaseIdx = 0; phaseIdx < numPhases; ++ phaseIdx) {
             if (!FluidSystem::phaseIsActive(phaseIdx)) {
+                fs.setEnthalpy(phaseIdx, 0.0);
                 continue;
             }
 
--- a/opm/models/blackoil/blackoilenergymodules.hh
+++ b/opm/models/blackoil/blackoilenergymodules.hh
@@ -376,7 +376,6 @@ public:
         // and the thermal condictivity coefficients
         for (int phaseIdx = 0; phaseIdx < numPhases; ++ phaseIdx) {
             if (!FluidSystem::phaseIsActive(phaseIdx)) {
-                //fs.setEnthalpy(phaseIdx, 0.0);
                 continue;
             }
 
--- a/opm/models/blackoil/blackoilintensivequantities.hh
+++ b/opm/models/blackoil/blackoilintensivequantities.hh
@@ -140,6 +140,7 @@ public:
         {
             if (priVars.primaryVarsMeaning() == PrimaryVariables::Sw_po_Sg)
                 // -> threephase case
+                assert( not(priVars.primaryVarsMeaning() == PrimaryVariables::px) )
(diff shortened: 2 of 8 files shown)


ewoms/models/blackoil/blackoilintensivequantities.hh
```diff
@@ -335,7 +344,14 @@ public:
         Scalar rockCompressibility = problem.rockCompressibility(elemCtx, dofIdx, timeIdx);
         if (rockCompressibility > 0.0) {
             Scalar rockRefPressure = problem.rockReferencePressure(elemCtx, dofIdx, timeIdx);
-            Evaluation x = rockCompressibility*(fluidState_.pressure(oilPhaseIdx) - rockRefPressure);
+            Evaluation x;
+            if (FluidSystem::phaseIsActive(oilPhaseIdx)) {
+                x = rockCompressibility*(fluidState_.pressure(oilPhaseIdx) - rockRefPressure);
```
totto82: Maybe we should add a method in the fluidState that return the refereance pressure and temperature.  Just fluidState_.pressure() and fluidState_.temperature(). And then set the reference phase in the BlackoilXXXIndices class. 
The need to access the reference values applies to many places in both the simulator part and the well part of the code.

ewoms/models/blackoil/blackoilprimaryvariables.hh
```diff
@@ -284,7 +289,16 @@ public:
...
 
         // assign the actual primary variables
-        if (primaryVarsMeaning() == Sw_po_Sg) {
+        if (primaryVarsMeaning() == OnePhase_p) {
+            if (waterEnabled) {
+                (*this)[waterSaturationIdx] = FsToolbox::value(fluidState.saturation(waterPhaseIdx));
+                (*this)[pressureSwitchIdx] = FsToolbox::value(fluidState.pressure(waterPhaseIdx));
+            } else {
+                throw std::logic_error("Only pure ware is presently allowed");
+            }
```
totto82: Part of the code implements onephase more general then only for water.  Do we have a plan to also support only oil and only gas?

ewoms/models/blackoil/blackoilratevector.hh
```diff
@@ -81,17 +81,6 @@ public:
     BlackOilRateVector(Scalar value) : ParentType(Toolbox::createConstant(value))
     {}
 
-    template <class Eval = Evaluation>
-    BlackOilRateVector(const typename std::enable_if<std::is_same<Eval, Evaluation>::value, Evaluation>::type& value) : ParentType(value)
-    {}
-
```
totto82: Why are this removed

ewoms/models/blackoil/blackoilenergymodules.hh
```diff
@@ -376,7 +376,6 @@ public:
         // and the thermal condictivity coefficients
         for (int phaseIdx = 0; phaseIdx < numPhases; ++ phaseIdx) {
             if (!FluidSystem::phaseIsActive(phaseIdx)) {
-                fs.setEnthalpy(phaseIdx, 0.0);
                 continue;
```
totto82: Why this change

ewoms/models/blackoil/blackoilintensivequantities.hh
atgeirr: I think that might be a good idea, although we should be careful not to use the wrong pressure. I do not know a simple general rule that tells me which quantities should be evaluated at phase pressures and which should be at "reference" pressure (putting reference in quotes here, as the term is also used for things like rockRefPressure in the above code, which is a different concept...).

ewoms/models/blackoil/blackoilprimaryvariables.hh
atgeirr: I don't think there are concrete plans now, but it should not be excluded (as you cannot do everything with water, such as table-based PVT).

OPMUSER:
throw std::logic_error("Only pure *ware* is presently allowed");

water

OPM User
opmuser0@gmail.com

On 10-Oct-19 14:36, Atgeirr Flø Rasmussen wrote:
> throw std::logic_error("Only pure ware is presently allowed");



-- 
This email has been checked for viruses by Avast antivirus software.
https://www.avast.com/antivirus

ewoms/models/blackoil/blackoilratevector.hh
atgeirr: This constructor caused errors for single-phase cases, and seems to be unused for other cases.

ewoms/models/blackoil/blackoilenergymodules.hh
atgeirr: Temporary reverted, will check in testing.

akva2:
This broke the post-build.

ewoms/models/blackoil/blackoilintensivequantities.hh
totto82: I still think the solution you merged is not very robust. Now the code that picks the right pressure depending on the available phase (or model) is spread around in the code.  In this code the priority is oil water gas. Any reason at all for that particular choice?

atgeirr: I agree (still) that we should move this logic to the fluid state object, but I did not think it was a good idea to add that refactoring to these PRs (I see now that you did not agree with that...)

The "oil->water->gas" priority order is arbitrary, but not more so than our previous choice "oil (only)". Did you already have an idea or prototype for changing the fluid state? If so, a PR for that would be the natural place to discuss and test this. Instead of a hard-coded order, the "reference" phase could be a (compile-time) property for example.

PR #532, opened 2019-08-23
changed: rename to opm-models

akva2:
This is the initial renaming of this module to opm-models. It is a minimal effort thing. We cannot test it before the project name is changed on github due to git urls.


atgeirr:
I think it should be safe to do the following:
 - Rename the repo on github. (No-one is hurt because the old name still works.)
 - Test these PRs on Jenkins.
 - Merge all when green.

The only thing we cannot really test is verifying that all works fine even after the old name stops working (not sure how long that takes but months I guess).

akva2:
we will deal with whatever fallout rears its head.

atgeirr:
jenkins build this opm-common=970 opm-grid=385 opm-material=349 opm-simulators=1985 please

atgeirr:
jenkins build this opm-common=970 opm-grid=385 opm-material=349 opm-simulators=1985 please

atgeirr:
All merged, thanks @akva2

PR #535, opened 2019-09-09
changed: ewoms/linear -> opm/simulators/linalg

akva2:
A small contribution towards unifying code base some more. Use same directory as these components use in simulators.


akva2:
jenkins build this opm-simulators=1998 please

akva2:
now sits on top of https://github.com/OPM/opm-models/pull/536

akva2:
jenkins build this opm-simulators=1998 please

akva2:
jenkins build this opm-simulators=1998 please

akva2:
jenkins build this opm-simulators=1998 please

blattms:
I do not like this change. It gives the impression that this code is used by vanilla flow. Last time
I checked this was not the case. But I might be wrong.
If I am right, this will be quite surprising for developers when searching for bugs,
becoming accustomed to the code base, etc.

atgeirr:
Thanks, merging.

joakim-hove:
> I do not like this change. It gives the impression that this code is used by vanilla flow. Last time
I checked this was not the case. But I might be wrong.
If I am right, this will be quite surprising for developers when searching for bugs,
becoming accustomed to the code base, etc.

I do not know the facts of the matter here - but if @blattms is correct I think his point is very good; finding your way in a large codebase like opm *is* challenging to a newcomer, and good names etc certainly help. If there are parts of the opm-models which are not used in the flow codepath retaining e.g. ewoms or something else in both the filesystem and namespace seems like a good idea.

atgeirr:
I forgot to reload the web page before merging, so I did not see the comment by @blattms when I merged, in spite of it being made before. I assumed there were no comments, terribly sorry for that. I would certainly not have merged without considering the discussion.

Do you think this is serious enough to mandate a reversion?

To the matter at hand: I think we should unify things such as namespaces etc., and this is a step in the right direction, even if much of the code is not used by Flow (as far as I can tell, some parts are used including the MatrixBlock and the backend stuff). There are many parts used by both Flow and non-Flow programs/problems/tests/ebos-variants, I think they should all be in the Opm namespace. Another matter is if some of this should be pruned, probably the answer is yes?

atgeirr:
No comments, so I guess I'll leave it in for now? My offer to revert was sincere and will remain open.

PR #536, opened 2019-09-09
changed: explicitly list installed sources

akva2:
no globbing. like in all other modules.


akva2:
jenkins build this please

PR #547, opened 2019-09-16
changed: ewoms/models/richards -> opm/models/richards

akva2:


akva2:
jenkins build this please

PR #550, opened 2019-09-19
Move ewoms/* into an opm/models/* structure

akva2:
Collection of the previous PRs. No objections received, so this is it then.


akva2:
jenkins build this opm-simulators=2016 please

PR #552, opened 2019-10-07
fix bug in PLYROCK

totto82:
Should fix OPM/opm-simulators#2032


totto82:
jenkins build this please

totto82:
This fixes a reported bug. I will self-merge.

PR #554, opened 2019-10-08
enable rockcomp in flow

totto82:


blattms:
backported to release via #564

PR #556, opened 2019-10-14
fixed: install blackoilonephaseindices.hh

akva2:


blattms:
Just checked. Needs to go to he release, too. Please remember to tag such PRs or we might either miss them or do double the work.

PR #573, opened 2019-10-28
update redhat packaging

akva2:
new boost::filesystem usage requires a newer boost.


akva2:
@blattms looks like you forgot a module.

blattms:
backported to release via #575

PR #581, opened 2019-11-17
Use DeckItem::hasValue() before accessing Todd Langstaff parameters

joakim-hove:
While working on refactoring defaults - to refactor the 3D system I encountered this problem. There is bug/feature in the `DeckItem::getData()` function which is easy to get fooled by. The opm-common Deck data structures try quite hard to keep control over what data is assigned/defautled/and... and so on - but by using:
```
std::vector<T> data& = deck_item::getData<T>(0);
```
you can bypass virtually all checks and get invalid data[1]. When things were refactor the code addressed in this PR produced undefined values.

This PR can (and should!) be merged right away - indepent of the ongoing efforts in opm-common; but it must be merged before/along with https://github.com/OPM/opm-common/pull/1230.

[1]: This is being tightened up.

### DIFF
--- a/opm/models/blackoil/blackoilsolventmodules.hh
+++ b/opm/models/blackoil/blackoilsolventmodules.hh
@@ -294,16 +294,18 @@ public:
 
                 for (unsigned regionIdx = 0; regionIdx < numMiscRegions; ++regionIdx) {
                     const auto& tlmixparRecord = deck.getKeyword("TLMIXPAR").getRecord(regionIdx);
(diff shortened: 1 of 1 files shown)


joakim-hove:
jenkins build this please

joakim-hove:
At some point we should probably internalize the relevant parameters in a `EclipseState` class - for future safety.

atgeirr:
Thanks, merging.

PR #586, opened 2020-01-21
Add and use gatherStrings() for exception messages.

atgeirr:
Ensures user sees any useful exception messages during initialization. This scratches an old itch, probably not just for me, but also for frustrated users.

As of now, for certain types of failure that for example throw during fluid system construction or equilibration, the user only sees a generic "Could not initialize the problem" message. The .DBG file also just gets the same message.

With the proposed code, we gather the error strings (if any) from each process, and print the first one. This should be appropriate, as usually all processes throw for the same reason in this phase. If not, printing the first should let the user fix the problem (in the deck usually) and try again, eventually seeing all problems.

To implement this, I basically imported and refactored a bit of code from opm-simulators (the gatherDeferredLogger.cpp file), making it a bit more general in the process. It is a small step towards having a small MPI helper library. As can be seen in the code, it does not attempt to be complete, only supporting the particular case needed, with the exception of the `Packer<std::vector>` partial specialization.

### DIFF
--- a/opm/models/parallel/mpiutil.hh
+++ b/opm/models/parallel/mpiutil.hh
@@ -1,0 +1,211 @@
+// -*- mode: C++; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*-
+// vi: set et ts=4 sw=4 sts=4:
+/*
+  This file is part of the Open Porous Media project (OPM).
+
+  OPM is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 2 of the License, or
+  (at your option) any later version.
+
+  OPM is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with OPM.  If not, see <http://www.gnu.org/licenses/>.
+
+  Consult the COPYING file in the top-level source directory of this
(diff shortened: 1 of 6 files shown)


atgeirr:
jenkins build this please

atgeirr:
Missing `#include <cassert>` caused the test failure, trying again.

atgeirr:
jenkins build this please

opm/models/parallel/mpiutil.hh
```diff
@@ -0,0 +1,231 @@
...
+        {
+            return MPI_UNSIGNED;
+        }
+    };
+
+    template <>
+    struct MpiType<char>
+    {
+        // static constexpr MPI_Datatype type = MPI_CHAR;
+        static MPI_Datatype type()
```
akva2: these can be taken from dune, Dune::MPITraits<T>::getType()

atgeirr: Will fix.

atgeirr:
jenkins build this please

blattms:
Overall this looks like a nice improvement.

You might want to consider using [MPITraits](https://dune-project.org/doxygen/master/mpitraits_8hh.html) to get the MPI type and save some code.

Maybe in the long run opm/simulator/utils/ParallelRestart.[hc]pp and use the
packing from there to save some more code.

atgeirr:
@blattms The MPITraits are used now, I was not aware of them before pointed out by you and @akva2.

The ParallelRestart code is not used because it is downstream from here, I agree it would be better to have just one code in principle. I discussed the issue beforehand with Arne Morten, settling on the current approach as sufficient.

I had to make a small addition to the test runner, that will hopefully take care of the test failure.

atgeirr:
jenkins build this please

PR #587, opened 2020-01-24
Wrap the art2dgf utility in if (BUILD_EXAMPLES) ...  endif()

joakim-hove:
Had problems with this utility at one location - should not really be important; so wrapped in a conditinal.


joakim-hove:
jenkins build this please

PR #588, opened 2020-01-30
fix plmax for > 1 regions

totto82:


totto82:
jenkins build this please

totto82:
This is a simple fix, I will merge this later today, if nobody complains.

GitPaean:
It looks correct to me. Merge as will.

PR #599, opened 2020-03-04
Remove deck usage from BlackoilSolventModule

akva2:
Thus we remove the need for serialization.


akva2:
jenkins build this opm-simulators=2401 please

PR #603, opened 2020-03-17
Fix compilation issues due to removed Dune::shared_ptr.

blattms:
Those methods return a unique_ptr since at least DUNE 2.6.

Probably I am having weired options such that these code paths even get compiled.

### DIFF
--- a/opm/models/io/cubegridvanguard.hh
+++ b/opm/models/io/cubegridvanguard.hh
@@ -72,7 +72,7 @@ class CubeGridVanguard : public BaseVanguard<TypeTag>
     typedef typename GET_PROP_TYPE(TypeTag, Simulator) Simulator;
     typedef typename GET_PROP_TYPE(TypeTag, Grid) Grid;
 
-    typedef Dune::shared_ptr<Grid> GridPointer;
+    typedef std::unique_ptr<Grid> GridPointer;
     typedef typename Grid::ctype CoordScalar;
     enum { dimWorld = Grid::dimensionworld };
     typedef Dune::FieldVector<CoordScalar, dimWorld> GlobalPosition;
--- a/opm/models/io/simplexvanguard.hh
+++ b/opm/models/io/simplexvanguard.hh
@@ -65,7 +65,7 @@ class SimplexGridVanguard
     typedef typename GET_PROP_TYPE(TypeTag, Simulator) Simulator;
     typedef typename GET_PROP_TYPE(TypeTag, Grid) Grid;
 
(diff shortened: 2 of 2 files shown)


blattms:
jenkins build this please

bska:
> Probably I am having weird options such that these code paths even get compiled.

Maybe so, but there is an implicit conversion from `std::unique_ptr<T>` to `std::shared_ptr<T>`&mdash;one that even deals with custom deleters&mdash;so for instance the following compiles and runs without problems
```C++
std::shared_ptr<T> p = std::make_unique<T>(args);
```

blattms:
Maybe I was'nt clear. The compile error was due to not Dune::shared_ptr available for current DUNE.

My fix was the most straight forward since these methods return a unique_ptr and other vanguards
already use this.

What would the advantage of using a shared_ptr be?

blattms:
Could you check where your trilinos is from? Assuming the naming scheme is the same on Debian the command would be

apt-cache policy libtrilinos-zoltan-dev

Maybe @akva2 can tell whether we ship our own or not. Judging from https://launchpad.net/~opm/+archive/ubuntu/ppa we don't for bionic. That would mean we might be mixing differently configured mpi versions. Maybe we should ship our own?

akva2:
Until proven otherwise my gut feeling is local issue.
I have run bionic with my rebuilt openmpi packages for 1.5y using various packages built against the vanilla ones without any hitches.

We indeed only ship zoltan for xenial, it is in main repos for bionic.

bska:
> Until proven otherwise my gut feeling is local issue.

It's not.  It's an Ubuntu 18.04 issue&mdash;specifically another manifestation of the OpenMPI package issue.  I built new OpenMPI libraries and binaries using sources from open-mpi.org, built a PT-Scotch library linking to those new OpenMPI libraries (painful experience), and then built a new Zoltan library linking against that PT-Scotch library.  Then, I built the entire OPM stack and now I have a Flow binary that runs debug in parallel without issue.  The only annoying thing now is that I have to use full paths to `mpirun`/`mpiexec` since the system version ones are broken.

akva2:
just to be sure; you did use my rebuilt packages from the ppa before your manual build?

akva2:
this is a obviously a fix. strictly speaking dune 2.6 still has Dune::shared_ptr but the factories do return a std::unique_ptr. i tested locally, with the 'weird' code path (as in, have dune-alugrid available). it still builds on 2.6 without this fix, but it also builds with this fix and i assume the dune ptr's were killed in 2.7.

I'll merge.

PR #608, opened 2020-04-24
[bugfix][GridCommHandle] fix incorrect signature of overloaded interface method.

dr-robertk:
This fixes an occuring segfault caused by incorrect interface method overload. It occurs with recent DUNE 2.7 (including backport) or later versions. This must go into the release.

### DIFF
--- a/opm/models/parallel/gridcommhandles.hh
+++ b/opm/models/parallel/gridcommhandles.hh
@@ -111,14 +111,14 @@ public:
     {
     }
 
-    bool contains(unsigned dim OPM_UNUSED, unsigned codim) const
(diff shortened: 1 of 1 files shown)


dr-robertk:
jenkins build this please

atgeirr:
Thanks for the fix, merging. Just curious: what goes wrong without this?

dr-robertk:
You'll get a segfault, because the wrong method will call itself infinitely many times.

dr-robertk:
This was introduced in a recent backport, because in the data handle the method fixedsize had been changed to fixedSize (capital S) some years ago and the deprecation warning did not work. I'll probably have to look into this again.

PR #609, opened 2020-05-07
normalizing saturations

hnil:
In progress. Aim at normalizing the saturations to avoid the danger of haveing bad results. Initial test give worse performance on Norne, also norne output still have negative oil saturation in water cells.

### DIFF
--- a/opm/models/blackoil/blackoilprimaryvariables.hh
+++ b/opm/models/blackoil/blackoilprimaryvariables.hh
@@ -363,6 +363,8 @@ public:
         if (primaryVarsMeaning() == OnePhase_p){
             return false;
         }
+        Scalar eps_sat=eps;// eps to represent near the phase disapear 
+        Scalar eps_dis=eps*100;//eps to represent near saturated
         Scalar Sw = 0.0;
         if (waterEnabled)
             Sw = (*this)[Indices::waterSaturationIdx];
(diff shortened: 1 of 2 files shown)


hnil:
This is work in progress.

hnil:
With the option it the only real change is that it do not project the solvent saturation in the default setting. My opinion is that all saturation should be treated consistently if we do not have a very specific reason

bska:
jenkins build this please

bska:
jenkins build this please

opm/models/blackoil/blackoilprimaryvariables.hh
```diff
@@ -363,8 +363,8 @@ public:
         if (primaryVarsMeaning() == OnePhase_p){
             return false;
         }
-        Scalar eps_sat=eps;// eps to represent near the phase disapear 
-        Scalar eps_dis=eps*100;//eps to represent near saturated
+        Scalar eps_sat=0.0//eps;// eps to represent near the phase disapear 
+        Scalar eps_dis=0.0//eps*100;//eps to represent near saturated
```
bska, line 367: You need to reinsert the semicolons here.

hnil:
I will close this pullrequest to get one more inline with OPM policy: the new will be https://github.com/OPM/opm-models/pull/610

PR #610, opened 2020-05-13
Pure saturation chopping

hnil:
New pullrequest for introducing the possibility of saturation chopping. This should not modify the solution with default setting. To be able to do this it is slightly unlogical for surfactant runs.

### DIFF
--- a/opm/models/blackoil/blackoilnewtonmethod.hh
+++ b/opm/models/blackoil/blackoilnewtonmethod.hh
@@ -39,10 +39,12 @@ BEGIN_PROPERTIES
 NEW_PROP_TAG(DpMaxRel);
 NEW_PROP_TAG(DsMax);
 NEW_PROP_TAG(PriVarOscilationThreshold);
+NEW_PROP_TAG(ProjectSaturations);
 
 SET_SCALAR_PROP(NewtonMethod, DpMaxRel, 0.3);
 SET_SCALAR_PROP(NewtonMethod, DsMax, 0.2);
 SET_SCALAR_PROP(NewtonMethod, PriVarOscilationThreshold, 1e-5);
+SET_BOOL_PROP(NewtonMethod, ProjectSaturations, false);
 
 END_PROPERTIES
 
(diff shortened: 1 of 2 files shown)


bska:
Looks good at first glance.  I'll run a build check just to be sure, and then I'll have a closer look.

bska:
jenkins build this please

opm/models/blackoil/blackoilprimaryvariables.hh
```diff
@@ -565,6 +565,83 @@ public:
...
+                Sg = (*this)[Indices::compositionSwitchIdx];
+            Scalar Ssol = 0.0;
+            if (enableSolvent)
+                Ssol =(*this) [Indices::solventSaturationIdx];
+            
+            Scalar So = 1.0 - Sw - Sg - Ssol;
+            Sw = std::min(std::max(Sw,0.0),1.0);
+            So = std::min(std::max(So,0.0),1.0);
+            Sg = std::min(std::max(Sg,0.0),1.0);
+            Ssol = std::min(std::max(Ssol,0.0),1.0);
```
bska, line 588: Just for your information, this is a thing that became slightly more straightforward in C++17, with the introduction of the [`std::clamp()`](https://en.cppreference.com/w/cpp/algorithm/clamp) library function (from the `<algorithm>`) header:
```C++
Sw = std::clamp(Sw, 0.0, 1.0);
So = std::clamp(So, 0.0, 1.0);
Sg = std::clamp(Sg, 0.0, 1.0);
Ssol = std::clamp(Ssol, 0.0, 1.0);
```
GCC supports `std::clamp()` since version 7 so there's no restriction in using it for OPM.  By the way, I also had momentary lapse of consciousness and considered doing something like
```C++
for (auto& s : { Sw, So, Sg, Ssol }) {
    s = std::clamp(s, 0.0, 1.0);
}
```
but this obviously doesn't work because initializer lists effectively behave a lot like containers of `const` objects.

opm/models/blackoil/blackoilnewtonmethod.hh
```diff
@@ -309,6 +313,9 @@ protected:
 
         if (wasSwitched_[globalDofIdx])
             ++ numPriVarsSwitched_;
+        if(projectSaturations_){
+            bool chopped = nextValue.chopAndNormalizeSaturations();
```
bska, line 317: This variable (`chopped`) is guaranteed to be unused in the current version.  Did you mean to use it for something else here?

bska:
jenkins build this please

opm/models/blackoil/blackoilnewtonmethod.hh
hnil: We can remove it, but I think the right thing to do is to make a genneral system wich also should be used for the switching. Forexample the above function should take chopped as a variable for the cell (or that nexValue have this as a member) and then first time if chopped false one done forexample 0+eps and if chopped true and for the same limmit one do 0.  For now I can remove it.

bska:
jenkins build this please

bska:
Bulid check green.  I'll merge this now. Thanks a lot!

PR #613, opened 2020-05-29
Fix vtk output bugs

atgeirr:


opm/models/io/vtkmultiwriter.hh
```diff
@@ -85,9 +85,11 @@ class VtkMultiWriter : public BaseOutputWriter
 
             // determine name to write into the multi-file for the
             // current time step
+            // The file names in the pvd file are relative, the path should therefore be stripped.
+            const auto localFileName = fileName.substr(fileName.find_first_of('/') + 1);
```
akva2, line 89: i'd use the shiny filesystem class for this. it would stop me from wondering 'why the first slash'.

atgeirr: Ok, what is the best way to get the local filename from the string?

akva2: https://en.cppreference.com/w/cpp/filesystem/path - relative.

atgeirr: Thanks. In the end I went with the filename() function for simplicity, as the output directory is the same.

atgeirr:
jenkins build this please

atgeirr:
It turns out that the local tests here combined the vtk output code with fluid systems that did not have the phaseIsActive() function. Resolved by adding trivial such function to the base fluid system upstream.

atgeirr:
jenkins build this opm-material=420 please

atgeirr:
I will take a chance and self-merge this and the upstream PR. The changes are small and simple, review comments have been incorporated, and we solve a real problem with the VTK output.

PR #614, opened 2020-06-02
Make gas-oil related outputs conditional.

atgeirr:
This includes such things as saturation pressures, rs, rv etc.

This is necessary in addition to #613, for water-oil cases.


atgeirr:
jenkins build this please

PR #615, opened 2020-06-03
Use negative status to indicate --help, --print-properties, --print-parameters

blattms:
That is the documented convention and needed to be able to mark these runs as successful.


blattms:
Shameless selfmerge since test ran successfully downstream at OPM/opm-simulators#2650

PR #617, opened 2020-06-08
Replace property macro calls

berndflemisch:
Replace all property macro calls by native C++ code.


opm/models/blackoil/blackoilboundaryratevector.hh
```diff
@@ -42,28 +42,28 @@ namespace Opm {
...
-    typedef typename GET_PROP_TYPE(TypeTag, RateVector) RateVector;
-    typedef typename GET_PROP_TYPE(TypeTag, Indices) Indices;
-
-    enum { numEq = GET_PROP_VALUE(TypeTag, NumEq) };
-    enum { numPhases = GET_PROP_VALUE(TypeTag, NumPhases) };
-    enum { numComponents = GET_PROP_VALUE(TypeTag, NumComponents) };
-    enum { enableSolvent = GET_PROP_VALUE(TypeTag, EnableSolvent) };
-    enum { enablePolymer = GET_PROP_VALUE(TypeTag, EnablePolymer) };
-    enum { enableEnergy = GET_PROP_VALUE(TypeTag, EnableEnergy) };
+    typedef GetPropType<TypeTag, Properties::RateVector> ParentType;
```
akva2, line 47: for consideration: since you're touching all of this anyways, consider switching to using clauses.

berndflemisch: Good idea, I'll do that.

berndflemisch: Note to me: Use `sed -r 's/typedef (.+) ([^ ]+);/using \2 = \1;/g'`.

akva2:
jenkins build this please

atgeirr:
Thanks a lot for your effort! I'll merge this now.

berndflemisch:
Perfect! I'll proceed with `opm-simulators` once time permits.

akva2:
as there is a lot of more churn in opm-simulators, for your own sanity, i would really recommend a bite size approach there.

berndflemisch:
Would you recommend to make PRs replacing
* all macro calls in one subfolder like `opm/simulators/flow`,
* one type of macro calls like `GET_PROP_TYPE` in all subfolders.
Most likely the first one?

akva2:
even higher granularity if possible i'd say. if you deal with a file or two and it still builds, flush it out before you call it a day. the point is to allow for quick processing so you don't get stuck in a never-ending rebase hell.

atgeirr:
> * all macro calls in one subfolder like `opm/simulators/flow`,

I think this would probably work, that way you would usually be stepping on only one set of active toes at a time...

berndflemisch:
I'll see. My scripts catch pretty much all one-liners, I have to do manual tweaking for the multi-liners. I guess I'll distribute the PRs according to the manual effort.

PR #619, opened 2020-06-19
Add LinearizationType, which will be used for sequential methods.

atgeirr:
This consists of changes made to support a sequential solution approach. The biggest changes will be downstream in opm-simulators and not here. This PR can be merged independently, so that is what we should do. It should not change results or performance at all.

This branch is revised by me, I considered it as "including my review comments".


GitPaean:
jenkins build this please

atgeirr:
I can confirm that I get identical results with this PR. Tested on Norne with 8 processes.

opm/models/blackoil/blackoilpolymermodules.hh
```diff
@@ -884,6 +884,7 @@ public:
         // Use log(v0) as initial value for u
         auto u = v0AbsLog;
         bool converged = false;
+        // TODO make this into parameters
```
GitPaean, line 887: Not clear what this comment is about.

atgeirr: It refers to the method parameters on the next lines.

GitPaean: Sorry if I am too slow on this. I do not see `method parameters` on the next lines.  Could you please just copy the lines or give a little more details? Thanks.

opm/models/discretization/common/linearizationtype.hh
```diff
@@ -0,0 +1,44 @@
+// -*- mode: C++; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*-
```
GitPaean, line 1: Is it necessary to have this file in CMakeLists_files.cmake?

GitPaean:
FYI, I could not reproduce the jenkins build failure locally. @akva2 , any suggestions?

akva2:
it is like you point out, missing installation of the new header. jenkins uses installed builds, not super or sibling build.

GitPaean:
> it is like you point out, missing installation of the new header. jenkins uses installed builds, not super or sibling build.

Sure, Thanks. I think both me and @atgeirr using sibling build.

atgeirr:
> it is like you point out, missing installation of the new header

Amended and force pushed.

atgeirr:
jenkins build this please

opm/models/blackoil/blackoilpolymermodules.hh
GitPaean: My best guess here is to have a newton solver function as a template parameter.

GitPaean:
It looks good to me, I have no more comments.

atgeirr:
> It looks good to me, I have no more comments.

Thanks, I'll merge it then.

PR #620, opened 2020-07-07
Fixed bug in calculation of gravity contribution to boundary flux

rbe051:
The darcyfluxmodule had a  sign mistake and a missing division by distance in the calculation of the contribution of gravity to the flux on boundaries. The fix is consistent with the implementation of the gravity on internal faces (in calculateGradients_(...) )


totto82:
jenkins build this please

totto82:
Thanks for the fix

PR #625, opened 2020-09-21
Remove unused #include of Exceptions

joakim-hove:
See: https://github.com/OPM/opm-material/pull/428


joakim-hove:
jenkins build this please

PR #626, opened 2020-09-23
Make a method that invalidate and updates the intensive quantities

totto82:


opm/models/discretization/common/fvbasediscretization.hh
```diff
@@ -765,6 +765,29 @@ public:
...
+    {
+        invalidateIntensiveQuantitiesCache(timeIdx);
+
+        // loop over all elements...
+        ThreadedEntityIterator<GridView, /*codim=*/0> threadedElemIt(gridView_);
+#ifdef _OPENMP
+#pragma omp parallel
+#endif
+        {
+            unsigned threadId = ThreadManager::threadId();
```
atgeirr, line 778: This variable is not used. Not sure if that is a sign of a bug, or that it should just be removed?

opm/models/discretization/common/fvbasediscretization.hh
```diff
@@ -765,6 +765,29 @@ public:
...
+            ElementContext elemCtx(simulator_);
+            ElementIterator elemIt = threadedElemIt.beginParallel();
+            for (; !threadedElemIt.isFinished(elemIt); elemIt = threadedElemIt.increment()) {
+                // create an element context (the solution-based quantities are not
```
atgeirr, line 782: Could you elaborate (preferably in the comment) a little on why the solution-based quantities are not available?

atgeirr (commented):
I support the idea here wholeheartedly. I do have a few questions though.

opm/models/discretization/common/fvbasediscretization.hh
totto82: copy/paste bug

opm/models/discretization/common/fvbasediscretization.hh
totto82: copy/paste bug

PR #628, opened 2020-10-12
Alternative solvent extension for the black oil model.

osae:
A first version, depending on corresponding pr for opm-simulators (extboSimulators) and opm-common (kwPVTEXTBO).

### DIFF
--- a/opm/models/blackoil/blackoilextbomodules.hh
+++ b/opm/models/blackoil/blackoilextbomodules.hh
@@ -809,9 +809,7 @@ public:
      *
      * At this point the pressures and saturations of the fluid state are correct.
      */
-    void zPvtUpdate_(const ElementContext& elemCtx,
-                           unsigned scvIdx,
-                           unsigned timeIdx)
+    void zPvtUpdate_()
     {
         const auto& iq = asImp_();
         auto& fs = asImp_().fluidState_;
@@ -904,9 +902,7 @@ class BlackOilExtboIntensiveQuantities<TypeTag, false>
 
 public:
 
-    void zPvtUpdate_(const ElementContext& elemCtx OPM_UNUSED,
-                     unsigned scvIdx OPM_UNUSED,
-                     unsigned timeIdx OPM_UNUSED)
+    void zPvtUpdate_()
     { }
 
     void zFractionUpdate_(const ElementContext& elemCtx OPM_UNUSED,
--- a/opm/models/blackoil/blackoilextbomodules.hh
+++ b/opm/models/blackoil/blackoilextbomodules.hh
@@ -23,7 +23,11 @@
 /*!
  * \file
  *
- * \brief Contains the classes required to extend the black-oil model by solvents.
+ * \brief Contains the classes required to extend the black-oil model by solvent component.
+ *  For details, refer:
+ *  [*] T.H. Sandve, O. Sævareid and I. Aavatsmark: “Improved Extended Blackoil Formulation
+ *  for CO2 EOR Simulations.” in ECMOR XVII – The 17th European Conference on the
+ *  Mathematics of Oil Recovery,  September 2020.
  */
 #ifndef EWOMS_BLACK_OIL_EXTBO_MODULE_HH
 #define EWOMS_BLACK_OIL_EXTBO_MODULE_HH
@@ -31,11 +35,9 @@
 #include "blackoilproperties.hh"
 
 //#include <opm/models/io/vtkBlackOilExtboModule.hh> //TODO: Missing ...
-#include <opm/models/io/vtkblackoilsolventmodule.hh>
 
 #include <opm/models/common/quantitycallbacks.hh>
 
-#include <opm/material/fluidsystems/blackoilpvt/SolventPvt.hpp>
 #include <opm/material/common/Tabulated1DFunction.hpp>
 #include <opm/material/common/UniformXTabulated2DFunction.hpp>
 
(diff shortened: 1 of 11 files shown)


opm/models/blackoil/blackoilextbomodules.hh
```diff
@@ -0,0 +1,1014 @@
...
+namespace Opm {
+/*!
+ * \ingroup BlackOil
+ * \brief Contains the high level supplements required to extend the black oil
+ *        model.
+ */
+template <class TypeTag, bool enableExtboV = getPropValue<TypeTag, Properties::EnableExtbo>()>
+class BlackOilExtboModule
+{
+    typedef typename GET_PROP_TYPE(TypeTag, Scalar) Scalar;
```
akva2, line 71: please avoid property macros, see https://github.com/OPM/opm-models/commit/2a3389100016c25fdb679983dec380f23f22a52d

opm/models/blackoil/blackoilextbomodules.hh
```diff
@@ -0,0 +1,1014 @@
...
+
+
+public:
+    /*!
+     * \brief Called before the saturation functions are doing their magic
+     *
+     * At this point, the saturations of the fluid state correspond to those if the phases
+     * were pure hydrocarbons.
+     */
+    void zFractionUpdate_(const ElementContext& elemCtx,
```
totto82, line 752: The text is is old I think. 
zFractionUpdate_ updates the internal variable stored in the iq of the extended blackoil model. Maybe we should merge the code in zPvtUpdate_ in here and change the name to extendedBlackoilUpdate(....)

opm/models/blackoil/blackoilextbomodules.hh
```diff
@@ -0,0 +1,1014 @@
...
+              Sw = priVars.makeEvaluation(Indices::waterSaturationIdx, timeIdx).value();
+
+           if (Sw >= thresholdWaterFilledCell)
+              rs_ = 0.0;  // water only, zero rs_ ...
+        }
+
+        if (priVars.primaryVarsMeaning() == PrimaryVariables::Sw_po_Rs) {
+           rs_ = priVars.makeEvaluation(Indices::compositionSwitchIdx, timeIdx);
+           const Evaluation zLim = 0.7; //TODO: Approx. one-phase liq region ...
+           if (zFraction_ > zLim) {
```
totto82, line 794: Is the zLim value valid for different models?

opm/models/blackoil/blackoilextbomodules.hh
```diff
@@ -0,0 +1,1014 @@
...
+             pbub_ = ExtboModule::pbubRs(pvtRegionIdx, zLim, rs_);
+           } else {
+             pbub_ = ExtboModule::pbubRs(pvtRegionIdx, zFraction_, rs_);
+           }
+           //TODO: Undersaturated compressibility ...
+           bo_ = ExtboModule::bo(pvtRegionIdx, pbub_, zFraction_) - 4.0e-9*(fs.pressure(oilPhaseIdx)-pbub_);
+
```
totto82, line 801: This TODO needs to be addressed before merging. The value is not generic for different input data.

opm/models/blackoil/blackoilextbomodules.hh
```diff
@@ -0,0 +1,1014 @@
...
+           xvalue_ = ExtboModule::xvalue(pvtRegionIdx, pbub_, zFraction_);
+        }
+
+        if (priVars.primaryVarsMeaning() == PrimaryVariables::Sw_pg_Rv) {
+           rv_ = priVars.makeEvaluation(Indices::compositionSwitchIdx, timeIdx);
+           Evaluation rvsat = ExtboModule::rv(pvtRegionIdx, pbub_, zFraction_);
+           //TODO: Undersaturated compressibility ...
+           bg_ = ExtboModule::bg(pvtRegionIdx, pbub_, zFraction_) + 0.08*(rvsat-rv_);
+
```
totto82, line 810: same here.

opm/models/blackoil/blackoilextbomodules.hh
```diff
@@ -0,0 +1,1014 @@
...
+  Consult the COPYING file in the top-level source directory of this
+  module for the precise wording of the license and the list of
+  copyright holders.
+*/
+/*!
+ * \file
+ *
+ * \brief Contains the classes required to extend the black-oil model by solvents.
+ */
```
totto82, line 27: Add a reference to the ecmor paper

totto82 (changes requested):
Thanks for the contribution. Most of my comments are cosmetic. But the two TODOs I think needs to be addressed before merging.

totto82:
Thanks looks good. If jenkins is happy this can me merged.

opm/models/blackoil/blackoilextbomodules.hh
```diff
@@ -0,0 +1,996 @@
...
+
+    /*!
+     * \brief Re-compute face densities to account for zFraction dependency.
+     *
+     * At this point the pressures and saturations of the fluid state are correct.
+     */
+    void zPvtUpdate_(const ElementContext& elemCtx,
+                           unsigned scvIdx,
+                           unsigned timeIdx)
+    {
```
totto82, line 815: These input variables are not used. I suggest you just removed them and change the call site

opm/models/blackoil/blackoilextbomodules.hh
```diff
@@ -0,0 +1,992 @@
...
+
+        zFraction_ = priVars.makeEvaluation(zFractionIdx, timeIdx);
+
+        oilViscosity_ = ExtboModule::oilViscosity(pvtRegionIdx, fs.pressure(oilPhaseIdx), zFraction_);
+        gasViscosity_ = ExtboModule::gasViscosity(pvtRegionIdx, fs.pressure(gasPhaseIdx), zFraction_);
+
+        bo_ = ExtboModule::bo(pvtRegionIdx, fs.pressure(oilPhaseIdx), zFraction_);
+        bg_ = ExtboModule::bg(pvtRegionIdx, fs.pressure(gasPhaseIdx), zFraction_);
+
+        bz_ = ExtboModule::bg(pvtRegionIdx, fs.pressure(oilPhaseIdx), 0.99);
```
bska, line 758: @osae , @totto82 : Is this call supposed to go to the `Scalar` overload or to the `Evaluation` overload?

akva2:
This broke the build, https://ci.opm-project.org/job/opm-models/83/console

bska:
> This broke the build

For what it's worth, I have a local fix but it depends on the answer to my earlier [question](https://github.com/OPM/opm-models/pull/628#pullrequestreview-533746691).

opm/models/blackoil/blackoilextbomodules.hh
totto82: It should be Evaluation.

totto82:
Why wasn't this cached by Jenkins in the first place. I can provide a fix.

akva2:
ou only built from the downstream, thus the tests in opm-model are not built.

totto82:
I didn't know. Lucky I haven't broke more. So it is safer to build it from opm-common

akva2:
as a general rule you trigger from the first repo having a change.

bska:
> I can provide a fix.

Like I said, I already have a fix in place that's ready to go once I know the answer to which overload the call to `bg()` with a constant `0.99` is supposed to go to&mdash;the `Scalar` one or the `Evaluation` one.

totto82:
@bska It should  be Evaluation.  Then I will wait for your fix.

bska:
> It should be Evaluation.

Very good.  I've submitted PR #636 to address this issue.

PR #629, opened 2020-10-16
Add --help-all option that also prints hidden options.

atgeirr:
Requested by some developers to more easily see the full set of hidden options.

### DIFF
--- a/opm/models/utils/parametersystem.hh
+++ b/opm/models/utils/parametersystem.hh
@@ -513,7 +513,8 @@ void printParamList_(std::ostream& os, const std::list<std::string>& keyList, bo
 template <class TypeTag>
 void printUsage(const std::string& helpPreamble,
                 const std::string& errorMsg = "",
-                std::ostream& os = std::cerr)
+                std::ostream& os = std::cerr,
+                const bool showAll = false)
 {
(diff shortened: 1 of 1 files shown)


atgeirr:
jenkins build this please

opm/models/utils/parametersystem.hh
```diff
@@ -532,12 +533,15 @@ void printUsage(const std::string& helpPreamble,
         pInfo.paramName = "h,--help";
         pInfo.usageString = "Print this help message and exit";
         printParamUsage_(os, pInfo);
+        pInfo.paramName = "-help-all";
```
bska, line 536: Is this really supposed to have a single leading `-` character?

bska (approved):
This is indeed useful and i look forward to getting this functionality.  I just have one small question  about the `.paramName` of this new option, mostly because I'm not really familiar with the semantics of the `ParamInfo` type.

Other than that this looks good to me.

opm/models/utils/parametersystem.hh
atgeirr: Yes, because another dash will be added by printParamUsage_(). A little strange, but you see the same on the "h, --help" string above.

Merging since it has been approved.

PR #630, opened 2020-10-19
Disable property macros.

atgeirr:
Still available by manually overriding OPM_ENABLE_OLD_PROPERTY_MACROS, but not by default. This change has been communicated for a long time, as due to happen after the 2020.10 release.

Do not backport to the release.


atgeirr:
jenkins build this please

PR #633, opened 2020-11-02
add scotch(par)metis to debian packaging

bska:


bska:
Merging in my capacity as release manager for the 2020.10 release of OPM.

PR #636, opened 2020-11-19
ExtBoModule: Don't Overload on Scalar and Evaluation

bska:
The two types may be the same, in which case this is a build failure.

We switch to using templates for the property evaluation functions to allow the compiler to specialise it for all pertinent types.  In so doing we also need to explicitly form an `Evaluation` instance that was only implicitly constructed before.


bska:
jenkins build this please

totto82:
Thanks for both the fix and the cleanup. Can be merged when Jenkins is green.

bska:
> Can be merged when Jenkins is green.

Build check is green now.  Merging per maintainer approval.

PR #638, opened 2021-01-14
fix incrementer in boundarycontext

totto82:
Make sure that the iterator continues beyond the first boundary

### DIFF
--- a/opm/models/discretization/common/fvbaseboundarycontext.hh
+++ b/opm/models/discretization/common/fvbaseboundarycontext.hh
@@ -77,8 +77,15 @@ public:
     void increment()
     {
         const auto& iend = gridView().iend(element());
-        while (intersectionIt_ != iend && !intersectionIt_->boundary())
-            ++ intersectionIt_;
+
+        if(intersectionIt_ == iend)
+          return;
+
+        ++intersectionIt_;
+        // iterate to the next boundary intersection
+        while (intersectionIt_ != iend && !intersectionIt_->boundary()) {
+            ++intersectionIt_;
+        }
     }
 
     /*!
--- a/opm/models/discretization/common/fvbaseboundarycontext.hh
+++ b/opm/models/discretization/common/fvbaseboundarycontext.hh
@@ -252,7 +252,7 @@ public:
      *
      * \param boundaryFaceIdx The local index of the boundary segment
      */
-    const Intersection& intersection(unsigned boundaryFaceIdx OPM_UNUSED) const
+    const Intersection intersection(unsigned boundaryFaceIdx OPM_UNUSED) const
     { return *intersectionIt_; }
 
     /*!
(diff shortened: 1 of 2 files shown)


totto82:
jenkins build this please

totto82:
The failure is expected. The results looks more reasonable with this fix. 

With PR

image

current reference solution

image

totto82:
I think this fix is good, but there still seems to be some issues with the boundary conditions. I will need to investigate more. i.e. changed the status to WIP.

opm/models/discretization/common/fvbaseboundarycontext.hh
```diff
@@ -77,8 +77,12 @@ public:
     void increment()
     {
         const auto& iend = gridView().iend(element());
-        while (intersectionIt_ != iend && !intersectionIt_->boundary())
-            ++ intersectionIt_;
+        // iterate to the next boundary intersection
+        while (intersectionIt_ != iend) {
+            ++intersectionIt_;
+            if(intersectionIt_->boundary())
+                return;
```
blattms, line 84: I don't think that this is correct as you might be accessing the end iterator.

How about
```
++intersectionIt_; // asume we are not pass the last valid pos
while( intersectionIt_ != iend && !intersectionIt_->boundary())
  ++intersectionIt_;
```

totto82: It is used like this
`        for (unsigned faceIdx = 0; faceIdx < numBoundaryFaces; ++faceIdx, boundaryCtx.increment()) {`
so I guess it should never access the end iterator the way it is used now. I will add a guard to make sure it never access the end iterator in case it is used in elsewhere. I notices some strange boundary effects on another test case so I will look into that. Maybe there is more issues that needs to be addressed.

blattms: Assume that the iterator is at the last entry before calling increment. the your code does
```
...
++intersectionIt_; // we are at the end
intersectionIt_->boundary(); // boom, that is dereferencing the end iterator
```
But it will return unconditionally and not go beyond the end. There you are right.

totto82: Now I think the code is correct. 
First we need to find the first boundary intersection, then the next.

totto82:
jenkins build this please

opm/models/discretization/common/fvbaseboundarycontext.hh
blattms: Greate

blattms (approved):
Merge when green

totto82:
jenkins build this please

totto82:
Test test failure in BC_LAB is expected, see earlier conversation,  I will do an update data.

totto82:

totto82:
jenkins build this opm-tests=471 please

PR #639, opened 2021-01-14
Fix 2-phase blackoil boundary

totto82:
Add guard against unused phases and mapping to active index. 
Again many thanks to valgrind.


totto82:
jenkins build this please

totto82:
> Does this mean that the BlackoilRateVector is only used for the boundary? (Otherwise it should have hit us before)

Yes. 

Thanks for feedback. I will merge then.

PR #641, opened 2021-02-09
fix boundary face index in parallel

totto82:
The process boundaries are included in the vector of boundary intersections in the stencil i.e. numBoundaryFaces both includes number of real boundaries and process boundaries. This fix make sure that the iterator and the faceIdx are consistent.

### DIFF
--- a/opm/models/discretization/common/fvbaseboundarycontext.hh
+++ b/opm/models/discretization/common/fvbaseboundarycontext.hh
@@ -83,7 +83,7 @@ public:
 
         ++intersectionIt_;
         // iterate to the next boundary intersection
-        while (intersectionIt_ != iend && !intersectionIt_->boundary()) {
+        while (intersectionIt_ != iend && intersectionIt_->neighbor()) {
             ++intersectionIt_;
         }
     }
--- a/opm/models/discretization/common/fvbaselocalresidual.hh
+++ b/opm/models/discretization/common/fvbaselocalresidual.hh
@@ -426,7 +426,7 @@ protected:
 
         BoundaryContext boundaryCtx(elemCtx);
         // move the iterator to the first boundary
-        if(!boundaryCtx.intersection(0).boundary())
+        if(boundaryCtx.intersection(0).neighbor())
             boundaryCtx.increment();
 
         // evaluate the boundary for all boundary faces of the current context
@@ -434,6 +434,9 @@ protected:
         for (unsigned faceIdx = 0; faceIdx < numBoundaryFaces; ++faceIdx, boundaryCtx.increment()) {
             // add the residual of all vertices of the boundary
             // segment
+            if(!boundaryCtx.intersection(faceIdx).boundary())
+                continue;
+
             evalBoundarySegment_(residual,
                                  boundaryCtx,
                                  faceIdx,


totto82:
jenkins build this opm-simulators=3053 please

opm/models/discretization/common/fvbaseboundarycontext.hh
```diff
@@ -83,7 +83,7 @@ public:
 
         ++intersectionIt_;
         // iterate to the next boundary intersection
-        while (intersectionIt_ != iend && !intersectionIt_->boundary()) {
+        while (intersectionIt_ != iend && intersectionIt_->neighbor()) {
```
blattms, line 86: Not sure what we are aiming here for. Can you explain why we did end up on a non-boundary?

The thing is that in general one cannot assume that we are on the domain boundary if  `IntersectionIterator::neighbor()` returns `false`. It should return false also at the process boundary and on a level grid view the neighbor might simply be on another level (and we get false).

Is there maybe a bug in CpGrid or another grid we are using?

blattms: https://www.dune-project.org/doxygen/2.7.1/classDune_1_1Intersection.html#details

blattms (commented):
I am not sure whether this is the correct fix.

Could you explain the problem we are fixing?

To me it seems we are assuming that all intersection without a neigbor are boundaries. In a parallel grid this will include the processor boundary (which does not need to be a domain boundary).

opm/models/discretization/common/fvbaselocalresidual.hh
```diff
@@ -426,14 +426,17 @@ protected:
 
         BoundaryContext boundaryCtx(elemCtx);
         // move the iterator to the first boundary
-        if(!boundaryCtx.intersection(0).boundary())
+        if(boundaryCtx.intersection(0).neighbor())
```
blattms, line 429: :joy: BoundaryContext is really something. Upon on construction it might be on any intersection, any increment will always be on a boundary interface. (I spare the comment about BoundaryContext::intersection(0)).

I actually think the original code should move in the construct of BoundaryContext.

opm/models/discretization/common/fvbaselocalresidual.hh
```diff
@@ -426,14 +426,17 @@ protected:
...
             boundaryCtx.increment();
 
         // evaluate the boundary for all boundary faces of the current context
         size_t numBoundaryFaces = boundaryCtx.numBoundaryFaces(/*timeIdx=*/0);
         for (unsigned faceIdx = 0; faceIdx < numBoundaryFaces; ++faceIdx, boundaryCtx.increment()) {
             // add the residual of all vertices of the boundary
             // segment
+            if(!boundaryCtx.intersection(faceIdx).boundary())
+                continue;
```
blattms, line 438: This seems to be needed because of the above change.

blattms: We might even dereference the end pointer here. I think the correct opm.models way would be 
```
if( boundaryCtx.intersectionIt() == boundaryCtx.gridView().iend(boundaryCtx.element()) 
   break;
```
`boundaryCtx.increment()` will move to the next boundary or the end iterator `boundaryCtx.gridView().iend(boundaryCtx.element()` if `numBoundaryFaces` is to high.

blattms:
At least I finally got the problem.
- `FvBaseBoundaryContext::numBoundaryFaces()` might return a number higher than the actual number of boundaryFaces.
- `FvBaseBoundaryContext::increment will only stop on real boundaries.
- Hence we might dereference the end iterator.

But the change from `boundary()` to `!neighbor()` seems wrong. I might be missing something, though.

totto82:
Thanks for looking into this. I will try to explain a bit more. 

Without this fix, Flow + MPI + keyword BC segfaults. 

First the size of boundaryFaces_ is number of intersection - number of intersection where neighbor is true. 

https://github.com/OPM/opm-models/blob/04af39a1459859cc4c4d7fcc7ff806d13431f251/opm/models/discretization/ecfv/ecfvstencil.hh#L241-L254

So when we iterate over the boundary. We iterate over both real boundaries and partition boundaries 

https://github.com/OPM/opm-models/blob/04af39a1459859cc4c4d7fcc7ff806d13431f251/opm/models/discretization/common/fvbaselocalresidual.hh#L433-L434

Since we only want the real boundaries I added

```
            if(!boundaryCtx.intersection(faceIdx).boundary())
                continue;
```
I also needed to change `boundaryCtx.increment()` to move to next intersection where neighbor is false to make it consistent with the loop over both real and partitioned boundaries.

The fix in opm-simulators is needed since the transmissibility mapping takes the face index as an argument. The face index now gives the index into the combined real and partitioned boundaries list in the stencil. 

An alternative fix could be to change the loop in 

https://github.com/OPM/opm-models/blob/04af39a1459859cc4c4d7fcc7ff806d13431f251/opm/models/discretization/common/fvbaselocalresidual.hh#L433-L434

to loop over all intersections and avoid the usage of boundaryCtx as done in 

https://github.com/OPM/opm-simulators/blob/7862f7d56d7266f20858a2a2498148709c40107f/ebos/ecltransmissibility.hh#L205

but that will require changes in many of the models/problems in opm-models that I really would like to avoid. 

If my explanation is still insufficient I suggest a short jitsi. (before 11 today?)

blattms:
Sorry, I missed the time slot. Thanks for the explanation, I think I understood the problem.

> Without this fix, Flow + MPI + keyword BC segfaults.

My guess is that this happens in the loop (until numBoundaryFaces), because we there is no detection for the intersection iterator 
being at the end. lines like this
```
for (unsigned faceIdx = 0; faceIdx < numBoundaryFaces; ++faceIdx, boundaryCtx.increment()) {
  ....
}
```
will result in the boundaryCtxt (which wraps an intersection iterator) being positioned at the end iterator. Calling methods on it will dereference it.


maybe we can add a method `bool isAtEnd()` and `void .moveToFirstBoundary()` to the boundary context, and change such loops to:
```
for(boundaryCtxt.moveToFirstBoundary(); !boundaryCtxt.isAtEnd(), boundaryCtxt.increment())
```

I admit that this is not really elegant, but at seems like a slight improvement and should fix the bug.

I can prepare an alternative patch this afternoon if that helps. Just send me a small test case.

totto82:
> I can prepare an alternative patch this afternoon if that helps. Just send me a small test case.

That would be great. More readable code here for boundaries would be great. For test case I think you can just use your spe1 (or your favorite model) and just try to add

BC
 1 1 1 1 1 1 FREE X-
/

and run it with MPI.

blattms (approved):
I have now checked the boundary stuff in opm-models in more detail and conclude that this PR makes it a lot more consistent. 

It seems that opm-models tries to capture interior boundaries (not marked by the grid as boundaries). Internally this was consistent, but not in the interface of BoundaryContext (this is fixed with this PR).

While I am unsure whether flow even uses them, others might, Hence my suggested changes (disallowing interior borders) might break code.

I am suggesting to merge this PR and downstream. Thanks a lot and sorry for the delay.

totto82:
> @totto82 wouldyou try my proposal

You mean extending the iterator in boundaryCtxt to allow for something like

`for(boundaryCtxt.moveToFirstBoundary(); !boundaryCtxt.isAtEnd(), boundaryCtxt.increment())`

We may need to track the faceIdx as well. Would about adding a 
boundaryCtxt.intersectionIdx() and keep track of the index

A less performant solution, but maybe less error prone is to remove the iterator in boundaryCtxt
And instead add a local iterator in `intersection(faceIdx)` that actually use the faceIdx
```
intersectionIt = gridView().ibegin(element())
i = 0
while (intersectionIt_ != iend && i < faceIdx) 
      ++intersectionIt_
      ++i

return *intersectionIt_
```

If performance is important one could always use the intersection iterator directly and skip the boundaryCtxt

blattms:
Sorry for not being clear. Here come the complete explanation:

The problem is that opm-models seems to expect the boundary is evaluated whenever the intersection only has one neighbor. My guess is that the failing lense example relies on this.

Hence the check for `boundary()` should vanish from the commit f8bedee 
```diff
 for (unsigned faceIdx = 0; faceIdx < numBoundaryFaces; ++faceIdx, boundaryCtx.increment()) {
             // add the residual of all vertices of the boundary
             // segment
-            if(!boundaryCtx.intersection(faceIdx).boundary())
-                continue;
-
             evalBoundarySegment_(residual,
                                  boundaryCtx,
                                  faceIdx,
```

but that means that the model used in opm-simulators needs to make the check whether this is a boundary in evalBoundarySegment:
```diff
 void evalBoundarySegement_(auto& res, auto& boundaryCtx, auto faceIdx, ...){
+    if(!boundaryCtx.intersection(faceIdx).boundary())
+     return;
   ...
 }
```

I think this should  work here and in opm-simulators. 
Whether that (or the boundary treatment in opm-models) is a good idea, I can't say. This kind of boundary treatment was introduced after the fork from dumux. In dumux existed a somewhat similar treatment in the past (after the fork), but it was abandoned or changed later (maybe  for good reason). You could ask @berndflemisch what he thinks.

PR #643, opened 2021-03-09
move the boundary check downstream

totto82:
Should fix test failure in opm-models. 
Solution suggested by @blattms


totto82:
jenkins build this opm-simulators=3107 please

akva2:
the error is only visible when building without mpi so better include serial in that trigger.

akva2:
jenkins build this opm-simulators=3107 serial please

akva2:
fixes the issue, suggested by one dev and submitted by another. i'll merge this now to restore full CI.

PR #649, opened 2021-04-22
debian/control: Improved Debian package descriptions.

blattms:
For the proposal of OPM as a debian package (see issue #648). Comments/further improvements welcome.


blattms:
Closing as the plan should be to use the effort on https://salsa.debian.org/science-team/opm-models in the long run

PR #651, opened 2021-04-26
blackoilpolymermodules: include OpmLog.h

akva2:
Upstream changes will make this change necessary, however this can be merged separately as such.


akva2:
jenkins build this please

bska:
I don't understand the regression failures so I'll rerun the build check.  I **suspect** the failures were due to the reference solutions and/or opm-simulators being updated while the build check was running.

bska:
jenkins build this please

bska:
The earlier failure was a race condition and the build check is green now.  I'll merge into master.

PR #658, opened 2021-05-18
Remove explicit assignment-operator from class

akva2:
since it already has an assigment operator, the implicit
copy-constructor is deprecated. clang likes to
point this out.

### DIFF
--- a/opm/models/blackoil/blackoilratevector.hh
+++ b/opm/models/blackoil/blackoilratevector.hh
@@ -190,16 +190,6 @@ public:
             (*this)[i] = value;
         return *this;
     }
-
-    /*!
-     * \brief Assignment operator from another rate vector
-     */
-    BlackOilRateVector& operator=(const BlackOilRateVector& other)
-    {
-        for (unsigned i=0; i < this->size(); ++i)
-            (*this)[i] = other[i];
-        return *this;
-    }
 };
 
 } // namespace Opm


akva2:
jenkins build this please

akva2:
jenkins build this please

atgeirr:
Both here and in https://github.com/OPM/opm-grid/pull/536 we have classes that inherit from base classes without adding any new members, and do not have any virtual functions. Would it not be correct to just use the default copy constructors and assignment operators?

opm/models/blackoil/blackoilratevector.hh
```diff
@@ -84,6 +84,16 @@ public:
     BlackOilRateVector(Scalar value) : ParentType(Toolbox::createConstant(value))
     {}
 
+    /*!
+     * \brief Copy constructor
+     */
+    BlackOilRateVector(const BlackOilRateVector& other)
+      : Dune::FieldVector<GetPropType<TypeTag, Properties::Evaluation>,
```
atgeirr, line 91: In case I am wrong, and this is really needed, use ParentType to simplify here.

akva2:
it would probably be correct, i just chose the don't remove just add approach. will double check and possibly change approach tmrw.

akva2:
seems to work fine in removal mode.

akva2:
jenkins build this please

PR #662, opened 2021-06-23
Unstructured grid vanguard

rbe051:
This pull request adds a vanguard for the unstructured grid.


akva2:
rebase.

CMakeLists_files.cmake
```diff
@@ -169,6 +169,7 @@ list (APPEND PUBLIC_HEADER_FILES
              opm/models/io/vtktemperaturemodule.hh
              opm/models/io/vtkcompositionmodule.hh
              opm/models/io/structuredgridvanguard.hh
+						 opm/models/io/unstructuredgridvanguard.hh
```
akva2, line 173: kill the tabs.

rbe051: done

blattms:
opm-grid does not seem to be required for opm-models (according to opm-models-prereq.cmake) and does not include any headers of it. Hence you need to check whether opm-grid was actually found and guard against the case when it is not found (add HAVE_OPM_GRID to opm-models_CONFIG_VARS in opm-models-prereq.cmake and use `#ifdef HAVE_OPM_GRID` in the source file). Otherwise you might break other people's code.

All the other grid specific vanguards seem to be in opm-simulators (below ebos). Any reason why this one should/needs to be in opm-models? That might be easier if there is no compelling reason to add it here.

PR #666, opened 2021-08-03
changed: get rid of OPM_UNUSED macro usage

akva2:
prefer anonymous parameters and c++17 [[maybe_unused]]


akva2:
jenkins build this please

blattms:
Thanks. Always a bit hesitant of removing parameter names when I see doxygen's @copydoc, but it seems like parameters are NEVER documented in this part of OPM, and hence I will trigger the merge button

akva2:
just admit it, your real concern was the number of the beast ;)

PR #667, opened 2021-08-05
extbomodule: enable extrapolation for function evaluations

akva2:
To avoid assertions, ref https://github.com/OPM/opm-common/pull/2594


totto82:
Thanks! Looks good. Please merge if OPM/opm-common#2594 is green.

akva2:
it's not green but it fixed the issue this addresses -> merging.

PR #668, opened 2021-08-26
Fixes compilation and deprecation warnings for Dune 2.8

blattms:


blattms:
jenkins build this please

blattms:
jenkins build this opm-simulators=3492 please

blattms:
2.8 has been released about a week ago. I think we should merge this and OPM/opm-simulators#3492 (both depend on eacg other) soon, to prevent unnecessary bug reports from users.

opm/models/parallel/gridcommhandles.hh
```diff
@@ -58,7 +58,11 @@ public:
         return codim == commCodim;
     }
 
+#if DUNE_VERSION_LT(DUNE_GRID, 2, 8)
```
akva2, line 61: slight nitpick: GT elsewhere, here LT. consider unifying.

blattms: I must admit that I am reluctant to change. The whole patch uses LT. I have this patch already in the Debian packages and uploaded these to mentors.debian.net. Hence this would trigger quite some work for very gain, don't you agree? (Also would need to test the build for two DUNE versions again...)

akva2: it's okay, if it is a major hassle it certainly is not worth it.

bska (approved):

PR #671, opened 2021-10-08
micp

daavid00:
It depends on: OPM/opm-common#2743 OPM/opm-material#475

### DIFF
--- a/opm/models/blackoil/blackoilmicpmodules.hh
+++ b/opm/models/blackoil/blackoilmicpmodules.hh
@@ -1,0 +1,675 @@
+// -*- mode: C++; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*-
+// vi: set et ts=4 sw=4 sts=4:
+/*
+  This file is part of the Open Porous Media project (OPM).
+
+  OPM is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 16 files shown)


opm/models/blackoil/blackoilnewtonmethod.hh
```diff
@@ -259,7 +262,7 @@ protected:
         Scalar deltaSg = 0.0;
         Scalar deltaSs = 0.0;
 
-        if (Indices::waterEnabled) {
+        if (Indices::waterEnabled && !enableMICP) {
             deltaSw = update[Indices::waterSaturationIdx];
```
totto82, line 266: What is the logic behind this change.

opm/models/blackoil/blackoilmicpmodules.hh
```diff
@@ -0,0 +1,918 @@
+// -*- mode: C++; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*-
```
totto82, line 1: New files should be added to the CmakeLists_files.cmake in opm-models

opm/models/blackoil/blackoilnewtonmethod.hh
daavid00: This is related to issues with running one-phase only with water. Then one solution is to add 'FluidSystem::numActivePhases() > 1' to the 'if'.

PR #678, opened 2021-11-29
update parameters BlackoilFluidSystem

goncalvesmachadoc:
Extension of PR https://github.com/OPM/opm-material/pull/429


totto82:
You need to add EnableEvaporation and enableSaltPrecipitation  to Properties. See for instance what is done for enableBrine: https://github.com/OPM/opm-models/blob/9c5cbc53a302cd104777548dc4531f062171a7f7/opm/models/blackoil/blackoilproperties.hh#L60

PR #680, opened 2022-01-06
Extrapolate pvt properties in the extended blackoil model

totto82:
Adding some more explicit extrapolation to avoid hard-crash when values are out of range.


totto82:
jenkins build this please

blattms:
Looks good to me. Let's trust jenkins and merge.

PR #681, opened 2022-01-07
Enable salt precipitation/dissolution

plgbrts:
Main contents of this PR is 

1. The extension of the brine module to deal with salt precipitation/dissolution. 
2. Allowing primary variable switching between salt concentration and solid( precipitated) salt

This PR depends on https://github.com/OPM/opm-material/pull/490

### DIFF
--- a/opm/models/blackoil/blackoilprimaryvariables.hh
+++ b/opm/models/blackoil/blackoilprimaryvariables.hh
@@ -103,7 +103,6 @@ class BlackOilPrimaryVariables : public FvBasePrimaryVariables<TypeTag>
     enum { enableFoam = getPropValue<TypeTag, Properties::EnableFoam>() };
     enum { enableBrine = getPropValue<TypeTag, Properties::EnableBrine>() };
     enum { enableSaltPrecipitation = getPropValue<TypeTag, Properties::EnableSaltPrecipitation>() };
-
     enum { enableEnergy = getPropValue<TypeTag, Properties::EnableEnergy>() };
     enum { enableMICP = getPropValue<TypeTag, Properties::EnableMICP>() };
     enum { gasCompIdx = FluidSystem::gasCompIdx };
--- a/opm/models/blackoil/blackoilprimaryvariables.hh
+++ b/opm/models/blackoil/blackoilprimaryvariables.hh
@@ -291,7 +291,8 @@ public:
         bool oilPresent = FluidSystem::phaseIsActive(oilPhaseIdx)?(fluidState.saturation(oilPhaseIdx) > 0.0):false;
         static const Scalar thresholdWaterFilledCell = 1.0 - 1e-6;
         bool onlyWater = FluidSystem::phaseIsActive(waterPhaseIdx)?(fluidState.saturation(waterPhaseIdx) > thresholdWaterFilledCell):false;
-        bool precipitatedSaltPresent = enableSaltPrecipitation?(fluidState.saltSaturation() > 0.0):false;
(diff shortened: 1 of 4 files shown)


totto82:
Thanks for the contribution. I have look at the code change and it looks good. Have you done any comparison with other implementations? I will just jenkins it to double check that it does not affect the standard brine module.

totto82:
jenkins build this opm-material=490 please

totto82:
jenkins build this opm-material=490 opm-simulators=3759 please

plgbrts:
> Have you done any comparison with other implementations?
We are planning to set up a test with a radial grid and twophase gas-water which we can  compare with earlier work done with DumuX

I see that Jenkins breaks when compiling the test case `opm-models/tests/reservoir_blackoil_vcfv.cc`.  Members `saltSaturation()` and` saltConcentration()` are not defined for a fluidstate coming from a `CompositionalFluidstate`. (This issue occurs if a`ssignNaive `is called with such fluidstate). Adding these members to the class  `FluidStateExplicitSaturationModule` resolves the compilation issue but  I am not sure if this is the right way. Do you have any advice? If you want to inspect this fix attempt  I can update the PR.

opm/models/blackoil/blackoilprimaryvariables.hh
```diff
@@ -273,6 +291,7 @@ public:
         bool oilPresent = FluidSystem::phaseIsActive(oilPhaseIdx)?(fluidState.saturation(oilPhaseIdx) > 0.0):false;
         static const Scalar thresholdWaterFilledCell = 1.0 - 1e-6;
         bool onlyWater = FluidSystem::phaseIsActive(waterPhaseIdx)?(fluidState.saturation(waterPhaseIdx) > thresholdWaterFilledCell):false;
+        bool precipitatedSaltPresent = enableSaltPrecipitation?(fluidState.saltSaturation() > 0.0):false;
```
totto82, line 295: You can not call saltSaturation from fluidState here. 
You must instead use something like 
const auto& SaltSaturation = BlackOil::getSaltSaturation_<....> ;
(see whats done for RS/RV and the stuff you added to the opm-material PR)

opm/models/blackoil/blackoilprimaryvariables.hh
```diff
@@ -347,6 +372,15 @@ public:
...
         }
 
+        if (enableSaltPrecipitation) {
+            if (primaryVarsMeaningBrine() == Sp) {
+                (*this)[saltConcentrationIdx] = FsToolbox::value(fluidState.saltSaturation());
+            }
+            else {
+                (*this)[saltConcentrationIdx] = FsToolbox::value(fluidState.saltConcentration());
+            }
+        }
```
totto82, line 382: The same comment as above

totto82:
@plgbrts I added some comments above. Can you check if this fixes the compilation issues? If my comments where unclear please give me a hint and I can try to explain it better.

plgbrts:
>  Can you check if this fixes the compilation issues? If my comments where unclear please give me a hint and I can try to explain it better.

I followed your comments   and it did resolve the compilation issue. Many thanks!

totto82:
jenkins build this opm-material=490 opm-simulators=3759 please

plgbrts:
@totto82 Only the water+gas and brine case failed. It should be fixed now.

plgbrts:
@totto82 could you re-run jenkins?

totto82:
jenkins build this opm-material=490 opm-simulators=3759 please

totto82:
Thanks for a very interesting extension of Flow. This is merge ready now. I suggest you add the test case to the ctest list to make sure the extension keep working.

PR #683, opened 2022-02-21
Add transflux module and a test that uses it

totto82:
Compute flux based on transmissibilites. The permeability is assumed to be diagonal and aligned with the local cell.

### DIFF
--- a/opm/models/common/transfluxmodule.hh
+++ b/opm/models/common/transfluxmodule.hh
@@ -1,0 +1,511 @@
+// -*- mode: C++; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*-
+// vi: set et ts=4 sw=4 sts=4:
+/*
+  This file is part of the Open Porous Media project (OPM).
+
+  OPM is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation, either version 2 of the License, or
+  (at your option) any later version.
+
+  OPM is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License
+  along with OPM.  If not, see <http://www.gnu.org/licenses/>.
+
+  Consult the COPYING file in the top-level source directory of this
+  module for the precise wording of the license and the list of
+  copyright holders.
+*/
+/*!
+ * \file
+ *
+ * \brief This file contains the flux module that uses transmissibilities
+ *
+ * The transmissibility approach to fluxes used here is limited
+ * to the two-point flux approximation
+ */
+#ifndef EWOMS_TRANS_FLUX_MODULE_HH
+#define EWOMS_TRANS_FLUX_MODULE_HH
+
+#include "multiphasebaseproperties.hh"
+#include <opm/models/utils/signum.hh>
+#include <opm/material/common/Valgrind.hpp>
+#include <dune/common/fvector.hh>
+#include <dune/common/fmatrix.hh>
+
+namespace Opm {
+
+template <class TypeTag>
+class TransIntensiveQuantities;
+
+template <class TypeTag>
+class TransExtensiveQuantities;
+
+template <class TypeTag>
+class TransBaseProblem;
+
+/*!
+ * \brief Specifies a flux module which uses transmissibilities.
+ */
+template <class TypeTag>
+struct TransFluxModule
+{
+    using FluxIntensiveQuantities = TransIntensiveQuantities<TypeTag>;
+    using FluxExtensiveQuantities = TransExtensiveQuantities<TypeTag>;
+    using FluxBaseProblem = TransBaseProblem<TypeTag>;
+    /*!
+     * \brief Register all run-time parameters for the flux module.
+     */
+    static void registerParameters()
+    { }
+};
+
+/*!
+ * \brief Provides the defaults for the parameters required by the
+ *        transmissibility based volume flux calculation.
+ */
+template <class TypeTag>
+class TransBaseProblem
+{ };
+
+/*!
+ * \brief Provides the intensive quantities for the transmissibility based flux module
+ */
+template <class TypeTag>
+class TransIntensiveQuantities
+{
+    using ElementContext = GetPropType<TypeTag, Properties::ElementContext>;
+protected:
+    void update_(const ElementContext&, unsigned, unsigned)
+    { }
+};
+
+/*!
+ * \brief Provides the transmissibility based flux module
+ */
+template <class TypeTag>
+class TransExtensiveQuantities
+{
+    using Implementation = GetPropType<TypeTag, Properties::ExtensiveQuantities>;
+
+    using FluidSystem = GetPropType<TypeTag, Properties::FluidSystem>;
+    using ElementContext = GetPropType<TypeTag, Properties::ElementContext>;
+    using Scalar = GetPropType<TypeTag, Properties::Scalar>;
+    using Evaluation = GetPropType<TypeTag, Properties::Evaluation>;
+    using GridView = GetPropType<TypeTag, Properties::GridView>;
+    using MaterialLaw = GetPropType<TypeTag, Properties::MaterialLaw>;
+    using Discretization = GetPropType<TypeTag, Properties::Discretization>;
+
+    enum { dimWorld = GridView::dimensionworld };
+    enum { numPhases = FluidSystem::numPhases };
+
+    typedef MathToolbox<Evaluation> Toolbox;
+    typedef Dune::FieldVector<Scalar, dimWorld> DimVector;
+    typedef Dune::FieldVector<Evaluation, dimWorld> EvalDimVector;
+    typedef Dune::FieldMatrix<Scalar, dimWorld, dimWorld> DimMatrix;
+
+public:
+    /*!
+     * \brief Return the intrinsic permeability tensor at a face [m^2]
+     */
+    const DimMatrix& intrinsicPermeability() const
+    {
+        throw std::logic_error("The ECL transmissibility module does not provide an explicit intrinsic permeability");
+    }
+
+    /*!
+     * \brief Return the pressure potential gradient of a fluid phase at the
+     *        face's integration point [Pa/m]
+     *
+     * \param phaseIdx The index of the fluid phase
+     */
+    const EvalDimVector& potentialGrad(unsigned) const
+    {
+        throw std::logic_error("The ECL transmissibility module does not provide explicit potential gradients");
+    }
+
+    /*!
+     * \brief Return the gravity corrected pressure difference between the interior and
+     *        the exterior of a face.
+     *
+     * \param phaseIdx The index of the fluid phase
+     */
+    const Evaluation& pressureDifference(unsigned phaseIdx) const
+    { return pressureDifference_[phaseIdx]; }
+
+    /*!
+     * \brief Return the filter velocity of a fluid phase at the face's integration point
+     *        [m/s]
+     *
+     * \param phaseIdx The index of the fluid phase
+     */
+    const EvalDimVector& filterVelocity(unsigned) const
+    {
+        throw std::logic_error("The ECL transmissibility module does not provide explicit filter velocities");
+    }
+
+    /*!
+     * \brief Return the volume flux of a fluid phase at the face's integration point
+     *        \f$[m^3/s / m^2]\f$
+     *
+     * This is the fluid volume of a phase per second and per square meter of face
+     * area.
+     *
+     * \param phaseIdx The index of the fluid phase
+     */
+    const Evaluation& volumeFlux(unsigned phaseIdx) const
+    { return volumeFlux_[phaseIdx]; }
+
+protected:
+    /*!
+     * \brief Returns the local index of the degree of freedom in which is
+     *        in upstream direction.
+     *
+     * i.e., the DOF which exhibits a higher effective pressure for
+     * the given phase.
+     */
+    unsigned upstreamIndex_(unsigned phaseIdx) const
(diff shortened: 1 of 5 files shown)


totto82:
jenkins build this please

opm/models/common/transfluxmodule.hh
```diff
@@ -0,0 +1,503 @@
...
+    typedef Dune::FieldVector<Evaluation, dimWorld> EvalDimVector;
+    typedef Dune::FieldMatrix<Scalar, dimWorld, dimWorld> DimMatrix;
+
+public:
+    /*!
+     * \brief Return the intrinsic permeability tensor at a face [m^2]
+     */
+    const DimMatrix& intrinsicPermeability() const
+    {
+        throw std::invalid_argument("The ECL transmissibility module does not provide an explicit intrinsic permeability");
```
blattms, line 116: Use std::logic_error? There is no argument here.

opm/models/common/transfluxmodule.hh
```diff
@@ -0,0 +1,503 @@
...
+
+    /*!
+     * \brief Return the pressure potential gradient of a fluid phase at the
+     *        face's integration point [Pa/m]
+     *
+     * \param phaseIdx The index of the fluid phase
+     */
+    const EvalDimVector& potentialGrad(unsigned) const
+    {
+        throw std::invalid_argument("The ECL transmissibility module does not provide explicit potential gradients");
```
blattms, line 127: Same

opm/models/common/transfluxmodule.hh
```diff
@@ -0,0 +1,503 @@
...
+
+    /*!
+     * \brief Return the filter velocity of a fluid phase at the face's integration point
+     *        [m/s]
+     *
+     * \param phaseIdx The index of the fluid phase
+     */
+    const EvalDimVector& filterVelocity(unsigned) const
+    {
+        throw std::invalid_argument("The ECL transmissibility module does not provide explicit filter velocities");
```
blattms, line 147: logic_error

opm/models/common/transfluxmodule.hh
```diff
@@ -0,0 +1,503 @@
...
+                pressureDifference_[phaseIdx] = 0.0;
+                volumeFlux_[phaseIdx] = 0.0;
+                continue;
+            }
+
+            // do the gravity correction: compute the hydrostatic pressure for the
+            // external at the depth of the internal one
+            const Evaluation& rhoIn = intQuantsIn.fluidState().density(phaseIdx);
+            Scalar rhoEx = Toolbox::value(intQuantsEx.fluidState().density(phaseIdx));
+            Evaluation rhoAvg = (rhoIn + rhoEx)/2;
```
blattms, line 250: Maybe this suffices but it might be a rough estimation if the cells are very different in size. Maybe use a (volume?) weighted harmonic mean or similar?

opm/models/common/transfluxmodule.hh
```diff
@@ -0,0 +1,503 @@
...
+            // is always the same: if the pressure is equal, the DOF with the lower
+            // global index is regarded to be the upstream one.
+            if (pressureDifference_[phaseIdx] > 0.0) {
+                upIdx_[phaseIdx] = exteriorDofIdx_;
+                dnIdx_[phaseIdx] = interiorDofIdx_;
+            }
+            else if (pressureDifference_[phaseIdx] < 0.0) {
+                upIdx_[phaseIdx] = interiorDofIdx_;
+                dnIdx_[phaseIdx] = exteriorDofIdx_;
+            }
```
blattms, line 270: Floating point comparison with `==`. Should we rather check that we are near to zero with an epsilon?

On the other hand: Does it really matter what we choose if it is nearly zero? If not we could also use
```
if (... > 0.0) ... else ...
```
and skip some computations.

opm/models/common/transfluxmodule.hh
```diff
@@ -0,0 +1,503 @@
...
+            else {
+                // if the pressure difference is zero, we chose the DOF which has the
+                // larger volume associated to it as upstream DOF
```
blattms, line 273: Out of curiosity: Why?

opm/models/common/transfluxmodule.hh
```diff
@@ -0,0 +1,503 @@
...
+                    dnIdx_[phaseIdx] = exteriorDofIdx_;
+                }
+                else if (Vin < Vex) {
+                    upIdx_[phaseIdx] = exteriorDofIdx_;
+                    dnIdx_[phaseIdx] = interiorDofIdx_;
+                }
+                else {
+                    assert(Vin == Vex);
+                    // if the volumes are also equal, we pick the DOF which exhibits the
+                    // smaller global index
```
blattms, line 287: This seems really random and only makes the code more complicated. Please use a simple if(Vin > Vex) ... else` if there is no good reason.

opm/models/common/transfluxmodule.hh
```diff
@@ -0,0 +1,503 @@
...
+            // does not matter, though.
+            unsigned upstreamIdx = upstreamIndex_(phaseIdx);
+            const auto& up = elemCtx.intensiveQuantities(upstreamIdx, timeIdx);
+
+            if (upstreamIdx == interiorDofIdx_)
+                volumeFlux_[phaseIdx] =
+                    pressureDifference_[phaseIdx]*up.mobility(phaseIdx)*(-trans);
+            else
+                volumeFlux_[phaseIdx] =
+                    pressureDifference_[phaseIdx]*(Toolbox::value(up.mobility(phaseIdx))*(-trans));
```
blattms, line 310: Unclear to me. Both branches use `up` and the negative transmissibility. Only difference is that else uses `Toolbox::value`. Somehow I expected something different. Please explain.

opm/models/common/transfluxmodule.hh
```diff
@@ -0,0 +1,503 @@
...
+                dnIdx_[phaseIdx] = -1;
+            }
+
+            short upstreamIdx = upstreamIndex_(phaseIdx);
+            if (upstreamIdx == interiorDofIdx_) {
+
+                // this is slightly hacky because in the automatic differentiation case, it
+                // only works for the element centered finite volume method. for ebos this
+                // does not matter, though.
+                const auto& up = elemCtx.intensiveQuantities(upstreamIdx, timeIdx);
```
blattms, line 384: Maybe we can somehow check that this is cell-centered?

opm/models/common/transfluxmodule.hh
```diff
@@ -0,0 +1,503 @@
...
+
+                volumeFlux_[phaseIdx] =
+                    pressureDifference_[phaseIdx]*up.mobility(phaseIdx)*(-trans);
+
+            }
+            else {
+                // compute the phase mobility using the material law parameters of the
+                // interior element. TODO: this could probably be done more efficiently
```
blattms, line 392: Use doxygen \todo?

opm/models/common/transfluxmodule.hh
```diff
@@ -0,0 +1,503 @@
...
+            short upstreamIdx = upstreamIndex_(phaseIdx);
+            if (upstreamIdx == interiorDofIdx_) {
+
+                // this is slightly hacky because in the automatic differentiation case, it
+                // only works for the element centered finite volume method. for ebos this
+                // does not matter, though.
+                const auto& up = elemCtx.intensiveQuantities(upstreamIdx, timeIdx);
+
+                volumeFlux_[phaseIdx] =
+                    pressureDifference_[phaseIdx]*up.mobility(phaseIdx)*(-trans);
```
blattms, line 387: For my understanding: Why always the negative transmissibility?

opm/models/common/transfluxmodule.hh
```diff
@@ -0,0 +1,503 @@
...
+        const auto& stencil = elemCtx.stencil(timeIdx);
+        const auto& face = stencil.interiorFace(scvfIdx);
+        const auto& interiorPos = stencil.subControlVolume(face.interiorIndex()).globalPos();
+        auto distVec0 = face.integrationPos() - interiorPos;
+        Scalar ndotDistIn = face.normal() * distVec0;
+        Scalar distSquaredIn = distVec0 * distVec0;
+        const auto& K0mat = elemCtx.problem().intrinsicPermeability(elemCtx, face.interiorIndex(), timeIdx);
+        // assume the permeability aligns with the grid
+        // and use only the permeability in the principle direction
+        int idx = 0;
```
blattms, line 461: same.

opm/models/common/transfluxmodule.hh
```diff
@@ -0,0 +1,503 @@
...
+        const Scalar& K0 = K0mat[idx][idx];
+        const Scalar T0 = K0 * ndotDistIn / distSquaredIn;
+        return T0;
+    }
+
+    template <class Context>
+    Scalar dofCenterDepth_(const Context& context, unsigned spaceIdx, unsigned timeIdx) const
+    {
+        const auto& pos = context.pos(spaceIdx, timeIdx);
+        return pos[dimWorld-1];
```
blattms, line 478: When will this assumption hold?

atgeirr:
> Compute flux based on transmissibilites.

What it the essential difference between this flux module and the eclfluxmodule.hh in opm-simulators?

totto82:
> What it the essential difference between this flux module and the eclfluxmodule.hh in opm-simulators?

It is the same. So all @blattms comments also holds for the current code in eclfluxmodule.hh that is used by Flow.

totto82:
@blattms Thanks for the comments. I would argue that we should keep this file the same as the one in opm-simulators in terms of logic and implementation. If you agree can this be merged?

blattms:
To me it seems that the file here is not the same as in opm-simulators. The latter has subtle differences (e.g. in the pressure difference computation), more bells and whistles. There is a lot of copied code and that does not help with maintenance.

If we can live with one version, than there should be just one. If we need both, then we should probably try to get rid off code
duplication be e.g. using ineritance or something else.

Conceptionally, opm-models is more general (supposed to support different discretization methods), while opm-simulators is still purely cell centered finite volume. Hence my arguments do have a point.
To the very least there should be an informative error if somebody tries this with the wrong discretization method.

totto82:
>Hence my arguments do have a point.

I agree. I will update the PR based on your comments.

opm/models/common/transfluxmodule.hh
totto82: Seems like a good idea. Is this something for Flow as well?

https://github.com/OPM/opm-simulators/blob/0c9bb3990e8f226ddba422604456fdda53472b4c/ebos/eclfluxmodule.hh#L271

opm/models/common/transfluxmodule.hh
totto82: Again, this is similar to the code used by Flow. I don't know why we have this careful code to handle the zero case.

opm/models/common/transfluxmodule.hh
totto82: Good question. It is what Flow does.

opm/models/common/transfluxmodule.hh
totto82: Same as flow

opm/models/common/transfluxmodule.hh
totto82: We compute fluxes from both sides. They should give the same values but the derivatives are wrt to different cells. Hence the logic of using Toolbox::value to make sure we don't  combine derivatives wrt to different cells.

opm/models/common/transfluxmodule.hh
totto82: Good point. Will fix.

opm/models/common/transfluxmodule.hh
totto82: I guess it is more by definition than by assumption.

opm/models/common/transfluxmodule.hh
totto82: Isn't it per definition? i.e. flux = - mob * T \delta P I have checked that it gives the same results as the standard flux module for Cartesian grids with constant permeability and also compared this with Flow on the Smeaheia model. So I guess it is correct, but "signs" are easy to get wrong and would depend on how pressureDifference is computed.

opm/models/common/transfluxmodule.hh
blattms: I would think so. But let's postpone this to another PR, as it might change results.

blattms:
You did not really comment on the differences. Here is what is missing here compared with eclfluxmodule.hh:
- additional support for solvent [here](/OPM/opm-simulators/blob/78c5675daf742e106f9453a6ba12b88a4f37e72b/ebos/eclfluxmodule.hh#L446) and [here](/OPM/opm-simulators/blob/78c5675daf742e106f9453a6ba12b88a4f37e72b/ebos/eclfluxmodule.hh#L284)
- usage of modified transmissibilities [here](/OPM/opm-simulators/blob/78c5675daf742e106f9453a6ba12b88a4f37e72b/ebos/eclfluxmodule.hh#L441) and [here](/OPM/opm-simulators/blob/78c5675daf742e106f9453a6ba12b88a4f37e72b/ebos/eclfluxmodule.hh#L347), but maybe that is handled here via `transmissibility_` and `transmissibilityBoundary_`
- apply [threshold pressure](/OPM/opm-simulators/blob/78c5675daf742e106f9453a6ba12b88a4f37e72b/ebos/eclfluxmodule.hh#L327)

What is the plan to reduce code duplication? Will eclfluxmodule.hh eventually vanish or based on this one?

opm/models/common/transfluxmodule.hh
totto82: Done

totto82:
The purpose of this is to be able to run simulations using simulators in opm-models that do not use ecl type decks but where the cell wise permeability is defined to follow the cell-wise grid orientation (Like we do for Flow). 
i.e. no plan for solvent, threshold pressure or modified permeability etc. There is no plans to remove or base eclfluxmodule on this. We may however want to remove the transfluxmodule.hh in the future when we come to the point where we can run compositional simulations using Flow.

PR #684, opened 2022-03-06
Enabling water evaporation into the gas phase

plgbrts:
Main items
1) adding gas-water ratio (Rvw)  to the black-oil equations
2) variable switching Sw<->Rvw ( typically water may disappear due to evaporation)
3) flux correction in case of mobility reduction  due to salt precipitation  (opm/models/blackoil/blackoilbrinemodules.hh)

This PR depends on https://github.com/OPM/opm-material/pull/491

### DIFF
--- a/opm/models/blackoil/blackoilnewtonmethod.hh
+++ b/opm/models/blackoil/blackoilnewtonmethod.hh
@@ -264,7 +264,7 @@ protected:
         Scalar deltaSg = 0.0;
         Scalar deltaSs = 0.0;
 
-        if (Indices::waterEnabled && FluidSystem::numActivePhases() > 1) {
+        if (Indices::waterEnabled && FluidSystem::numActivePhases() > 1 && currentValue.primaryVarsMeaning() != PrimaryVariables::Rvw_po_Sg) {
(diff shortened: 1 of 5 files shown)


bska:
jenkins build this opm-material=491 please

opm/models/blackoil/blackoilnewtonmethod.hh
```diff
@@ -342,6 +348,11 @@ protected:
                 const double sign = delta >= 0. ? 1. : -1.;
                 delta = sign * std::min(std::abs(delta), maxTempChange_);
             }
+            else if (enableBrine && pvIdx == Indices::saltConcentrationIdx && currentValue.primaryVarsMeaningBrine() == PrimaryVariables::Sp) {
```
blattms, line 351: Hey @plgbrts, could there be a bug here? 
Valgrind is reporting possible usage of uninitialized values in the last term (when running SPE1CASE1_BRINE. To me it seems that the value returned by `currentValue.primaryVarsMeaningBrine()` might not be initialized if `enableSaltPrecipitation==true` holds.

PR #688, opened 2022-04-11
adding primary variable combination/switching 

plgbrts:
This PR is part of a set of PRs to allow for water evaporation into gas in combination with oil vaporization.
This PR adds new primary variable combination and switches relevant for combined vaporization of water and oil into the gas phase.

### DIFF
--- a/opm/models/blackoil/blackoilprimaryvariables.hh
+++ b/opm/models/blackoil/blackoilprimaryvariables.hh
@@ -470,7 +470,12 @@ public:
                 }
             }
          }
-
+        //  Primary variable switches possibilities for black-oil (from -> to)
+        //  Sw_po_Sg (3-phase)    -> Rvw_po_Sg, Sw_pg_Rv, Rvw_pg_Rv, Sw_po_Rs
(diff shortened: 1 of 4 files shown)


opm/models/blackoil/blackoilprimaryvariables.hh
```diff
@@ -129,7 +129,7 @@ public:
         Sw_po_Rs, // water + oil case
         Sw_pg_Rv, // water + gas case
         Rvw_po_Sg, // gas + oil  case
-
+        Rvw_pg_Rv, //gas only
         OnePhase_p, // onephase case
```
totto82, line 133: As discussed. Please add an overview of the different options for primary-variable meaning that will help the other developers reading the code IMO.

totto82 (commented):
From code points of view this is ok. The implementations follows what is done for the oil/gas system. The code naturally becomes more complex when water is allowed to vaporize, but I guess that is unavoidable as long as the blackoil/variable switching logic is followed.

bska:
Merging this now since the upstream is in the build queue.

PR #690, opened 2022-04-20
[bugfix] Prevent accessing undefined variables for brine without salt precipitation

### LINKED ISSUE #689: Uninitialized values for brine without saltperticipation
See [comment](https://github.com/OPM/opm-models/pull/684/files#r854000348) on PR #684.

Should we maybe always initialize to Sp as default in [blackoilprimaryvariables.hh#L333-L338](/OPM/opm-models/blob/28d931f39d9440425bc20133b9e925bf2d1b938e/opm/models/blackoil/blackoilprimaryvariables.hh#L333-L338) or when constructing, @plgbrts?

blattms:
It seems like `currentValue.primaryVarsMeaningBrine()` will return an uninitialized variable if `enableSaltPrecipitation` is false. This can lead to undefined behavior especially in a parallel run. Hence we also check whether salt precipitation is is enabled and do nothing otherwise.

This change seems to be in line with the other code.

Closes #689

### DIFF
--- a/opm/models/blackoil/blackoilnewtonmethod.hh
+++ b/opm/models/blackoil/blackoilnewtonmethod.hh
@@ -348,7 +348,9 @@ protected:
                 const double sign = delta >= 0. ? 1. : -1.;
                 delta = sign * std::min(std::abs(delta), maxTempChange_);
             }
-            else if (enableBrine && pvIdx == Indices::saltConcentrationIdx && currentValue.primaryVarsMeaningBrine() == PrimaryVariables::Sp) { 
+            else if (enableBrine && pvIdx == Indices::saltConcentrationIdx &&
+                     enableSaltPrecipitation &&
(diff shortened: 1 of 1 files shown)


blattms:
@plgbrts Please double check.

blattms:
jenkins build this please

plgbrts:
> @plgbrts Please double check.

Thanks for fixing this! This fix looks good to me, but I think the flag enableBrine is not necessary.

blattms:
thanks a lot. I will check the enableBrine, then.

blattms:
On second thought, leaving it there seems consistent with the rest of the code.

It might mean that the enable variable in `if (enableXXX && pvIdx == Indices:::particularXXXIndex)` is always superfluous. Seems like the enableXXX variable is an indicator that `Indices:::particularXXXIndex)>=0` but I would assume that `pvIdx` is never negative.

Anyway, playing it safe here and merging as is.

PR #695, opened 2022-06-10
fix blackoil diffusion module

totto82:
This fixes two issue in the diffusion module. The first is related to conversion from molar to surface rates, the second is related to convergence and derivatives of the gradient.


totto82:
jenkins build this please

opm/models/blackoil/blackoildiffusionmodule.hh
```diff
@@ -135,23 +135,48 @@ public:
             // phase not present, skip
             if(bSAvg < 1.0e-6)
                 continue;
+            Evaluation convFactor = 1.0;
+            Evaluation diffR = 0.0;
+            if (FluidSystem::enableDissolvedGas() && phaseIdx == FluidSystem::oilPhaseIdx) {
+                Evaluation rsAvg = (fluidStateI.Rs() + Toolbox::value(fluidStateJ.Rs())) / 2;
+                convFactor = 1.0 / (fraction(0) + rsAvg);
```
atgeirr, line 142: Should the proper PVT region be used here, not 0? (Also below.)

atgeirr (commented):
Other than the comment about region index, I think this can be merged.

totto82:
jenkins build this please

opm/models/blackoil/blackoildiffusionmodule.hh
totto82: fixed

fgasa:
Commit solved issues with the oscillations and low diffusion flux. User-defined tortuosity for the long run might be a good feature of OPM code

totto82:
> User-defined tortuosity for the long run might be a good feature of OPM code

For now I just disabled tortuosity.  The current implemented model is based on Millington, R. J., & Quirk, J. P. (1961), but may not always be the best. A more general (A /phi^(1-m))n where A,m and n are user input and A is default to zero (i.e. no effect) could be a good solution.

totto82:
> Other than the comment about region index, I think this can be merged.

Since the region index is fixed, I will merge this now.

PR #702, opened 2022-08-04
Fix Build After "Unused.hpp" Removal

bska:
Needed for configurations that don't have Dune-localfunctions or which **do** have ALUGrid.


bska:
jenkins build this please

bska:
Instant self-merging to restore build.  I take full responsibility for any adverse effects.

PR #704, opened 2022-08-09
Modernization in blackoil model

akva2:
Use if constexpr to avoid building code when appropriate.


akva2:
jenkins build this please

blattms:
Thanks a lot. Looked good

PR #705, opened 2022-08-11
fixed: avoid vector<bool> in fvbasediscretization

akva2:
writes to `vector<bool>` is not thread safe.

closes https://github.com/OPM/opm-simulators/issues/3993

### DIFF
--- a/opm/models/discretization/common/fvbasediscretization.hh
+++ b/opm/models/discretization/common/fvbasediscretization.hh
@@ -730,7 +730,7 @@ public:
             return;
 
         intensiveQuantityCache_[timeIdx][globalIdx] = intQuants;
-        intensiveQuantityCacheUpToDate_[timeIdx][globalIdx] = true;
+        intensiveQuantityCacheUpToDate_[timeIdx][globalIdx] = 1;
     }
 
     /*!
@@ -747,7 +747,7 @@ public:
         if (!storeIntensiveQuantities())
             return;
 
-        intensiveQuantityCacheUpToDate_[timeIdx][globalIdx] = newValue;
+        intensiveQuantityCacheUpToDate_[timeIdx][globalIdx] = newValue ? 1 : 0;
     }
 
     /*!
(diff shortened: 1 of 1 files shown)


akva2:
jenkins build this please

opm/models/discretization/common/fvbasediscretization.hh
```diff
@@ -1960,7 +1960,8 @@ protected:
     // cur is the current iterative solution, prev the converged
     // solution of the previous time step
     mutable IntensiveQuantitiesVector intensiveQuantityCache_[historySize];
-    mutable std::vector<bool> intensiveQuantityCacheUpToDate_[historySize];
+    // while these are logically bools, concurrent writes to vector<bool> is not thread safe.
```
bska, line 1963: Nit: "concurrent writes ... **are** not"

opm/models/discretization/common/fvbasediscretization.hh
```diff
@@ -1960,7 +1960,8 @@ protected:
...
+    mutable std::vector<char> intensiveQuantityCacheUpToDate_[historySize];
```
bska, line 1964: Did you consider using `unsigned char` here?

bska (commented):
Good catch.  You don't really need to alter the assignments as `bool` will implicitly convert to (`unsigned`) `char`.  Other than that this looks good to me.

opm/models/discretization/common/fvbasediscretization.hh
akva2: fixed.

opm/models/discretization/common/fvbasediscretization.hh
akva2: uhm, not really since it was more typing ;) changed.

akva2:
and yes, i know bools are implicitly convertible but i personally prefer to be explicit about it. i can change it back if you insist.

akva2:
jenkins build this please

bska:
> i personally prefer to be explicit about it

I appreciate that, although you should arguably write `static_cast<unsigned char>(1u)` in that case since just the number `1` is an implicit, narrowing conversion from `int` to `unsigned char`.

> i can change it back if you insist.

No need.  I'll merge this as is.

PR #707, opened 2022-08-22
fix for gaswater system with water evaporation

plgbrts:
This PR fixes the negative water saturations observed in  https://github.com/OPM/opm-simulators/pull/4002.

### DIFF
--- a/opm/models/blackoil/blackoilprimaryvariables.hh
+++ b/opm/models/blackoil/blackoilprimaryvariables.hh
@@ -503,17 +503,19 @@ public:
             if(Sw < -eps && FluidSystem::enableVaporizedWater()) {
                 Scalar pg = 0.0;
                 Scalar T = asImp_().temperature_();
-                if (waterEnabled && gasEnabled && !oilEnabled) {
+                if constexpr (waterEnabled && gasEnabled && !oilEnabled) {
                     // twophase water-gas system
                     pg = (*this)[Indices::pressureSwitchIdx];
                 }
-                else if (So3 > 0.0 && Sg > 0.0) {
-                    // threephase case
-                    std::array<Scalar, numPhases> pC = { 0.0 };
-                    const MaterialLawParams& matParams = problem.materialLawParams(globalDofIdx);
-                    computeCapillaryPressures_(pC, So3, Sg + solventSaturation_(), /*Sw=*/ 0.0, matParams);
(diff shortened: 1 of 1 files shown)


plgbrts:
jenkins build this please

plgbrts:
jenkins build this please

opm/models/blackoil/blackoilprimaryvariables.hh
```diff
@@ -498,14 +498,23 @@ public:
...
-            if(Sw < -eps && So3 > 0.0 && Sg > 0.0  && FluidSystem::enableVaporizedWater()) {
-                Scalar po = (*this)[Indices::pressureSwitchIdx];
+            if(Sw < -eps && FluidSystem::enableVaporizedWater()) {
+                Scalar pg = 0.0;
                 Scalar T = asImp_().temperature_();
-                std::array<Scalar, numPhases> pC = { 0.0 };
-                const MaterialLawParams& matParams = problem.materialLawParams(globalDofIdx);
-                computeCapillaryPressures_(pC, So3, Sg + solventSaturation_(), /*Sw=*/ 0.0, matParams);
-                Scalar pg = po + (pC[gasPhaseIdx] - pC[oilPhaseIdx]);
+                if constexpr (waterEnabled && gasEnabled && !oilEnabled) {
```
akva2, line 506: can't if constexpr followed by a runtime else (ie the constexpr here is a nop). you either drop the constexpr or you have  to change to 
```
else {
  if (So3 > 0.0 && Sg > 0) {
     ...
  }
}
```

plgbrts: Thanks for pointing this out. I removed the `constexpr`.

plgbrts:
jenkins build this please

opm/models/blackoil/blackoilprimaryvariables.hh
atgeirr: Would it not be better to keep the constexpr (and nest the if, as suggested above), to let the compiler optimize more?

plgbrts: > Would it not be better to keep the constexpr (and nest the if, as suggested above), to let the compiler optimize more?
That makes sense. Changed the code using @avka second option.

plgbrts:
jenkins build this please

PR #710, opened 2022-08-25
changed: unify matrixblock.hh with downstream MatrixBlock.hpp

akva2:
Let's try this again.

The issue was that previously a generic utility function was returning the identity matrix for a failed invert.
I changed this and did not realize something was relying on this rather funky behavior (ie the Norne model).
I have now changed it so the generic utility throws, and then the exceptional behavior is applied locally in the well
code instead.

If you do not like this approach and rather want the old one I can do that, but I really do not think that is the proper way to do it.

### DIFF
--- a/opm/models/discretization/common/fvbasediscretization.hh
+++ b/opm/models/discretization/common/fvbasediscretization.hh
@@ -55,7 +55,6 @@
 #include <opm/models/utils/timer.hh>
 #include <opm/models/utils/timerguard.hh>
 #include <opm/models/io/vtkprimaryvarsmodule.hh>
-#include <opm/simulators/linalg/matrixblock.hh>
 
 #include <opm/material/common/MathToolbox.hpp>
 #include <opm/material/common/Valgrind.hpp>
--- a/opm/simulators/linalg/matrixblock.hh
+++ b/opm/simulators/linalg/matrixblock.hh
@@ -23,190 +23,208 @@
 #ifndef EWOMS_MATRIX_BLOCK_HH
 #define EWOMS_MATRIX_BLOCK_HH
 
-#include <dune/istl/scalarproducts.hh>
-#include <dune/istl/operators.hh>
-#include <dune/istl/solvers.hh>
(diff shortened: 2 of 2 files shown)


akva2:
everything passed fine, and the curves seems fine to me. i am pushing an update that uses a regular throw rather than OPM_THROW to avoid the terminal output.

akva2:
jenkins run norne norne_parallel opm-simulators=4028 please

atgeirr:
benchmark please opm-simulators=4028

blattms:
I don't know introduced this behavior, but I recall that in ancient times there were problems because of failed inversions. I admit that my memory about this is very faint.

blattms:
Does the introducing commit have a hint about the reason?

akva2:
reason is that inversion fails during linear iterations due to invalid temporary well state, and we want to continue the simulation as eventually the state is restored to valid again after some newton updates. that's not really the issue at hand, it's more about where to apply the hack.

ytelses:
Benchmark result overview: 




| Test    | Configuration                                                                                 | Relative |
| ------- | --------------------------------------------------------------------------------------------- | -------- |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 1                                                    | 0.995    |
| opm-git | OPM Benchmark: flow_mpi_extra - Threads: 8                                                    | 0.989    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 1                                                    | 1.001    |
| opm-git | OPM Benchmark: flow_mpi_norne - Threads: 8                                                    | 0.991    |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 1 - FOPT (Total Oil Production At End Of Run) | 1        |
| opm-git | OPM Benchmark: flow_mpi_norne_4c_msw - Threads: 8 - FOPT (Total Oil Production At End Of Run) | 1        |
* Speed-up = Total time master / Total time pull request. Above 1.0 is an improvement. *



View result details @ https://www.ytelses.com/opm/?page=result&id=1759

PR #712, opened 2022-09-01
change default max/min temp in newton clamping

totto82:
This removes the default temperature clamping limits for the blackoil thermal model.

### DIFF
--- a/opm/models/blackoil/blackoilnewtonmethod.hh
+++ b/opm/models/blackoil/blackoilnewtonmethod.hh
@@ -83,13 +83,13 @@ template<class TypeTag>
 struct TemperatureMax<TypeTag, TTag::NewtonMethod>
 {
     using type = GetPropType<TypeTag, Scalar>;
-    static constexpr type value = 400; //Kelvin
+    static constexpr type value = 1e9; //Kelvin
 };
 template<class TypeTag>
 struct TemperatureMin<TypeTag, TTag::NewtonMethod>
 {
     using type = GetPropType<TypeTag, Scalar>;
-    static constexpr type value = 280; //Kelvin
+    static constexpr type value = 0.0; //Kelvin
 };
 } // namespace Opm::Properties
 


totto82:
@goncalvesmachadoc Can you just check if this works for you case. Alternatively you could just set 
--temperature-min=0 (in kelvin) or something smaller than wtemp)
--temperature-max=1000 (in kelvin) or  something larger than wtemp) 
in the flow version you have and check your case

totto82:
jenkins build this please

goncalvesmachadoc:
> --temperature-min=0

Hi Tor, it does help a lot. The results look qualitative correct:
image

But still have the issue with temperatures in the grid several degrees below (or above) injection temperature.

totto82:
> But still have the issue with temperatures in the grid several degrees below (or above) injection temperature.

Could you share your exact setup and I can take a look. Is it based on the SPE1 test case? Entalpy is a a function of pressure and temperature, so it may happen that the temperature is decreased/ increased due to changes in pressure.

goncalvesmachadoc:
> > But still have the issue with temperatures in the grid several degrees below (or above) injection temperature.
> 
> Could you share your exact setup and I can take a look. Is it based on the SPE1 test case? Enthalpy is a a function of pressure and temperature, so it may happen that the temperature is decreased/ increased due to changes in pressure.

Yes, but 6-8 degrees change for water seems a bit too much. Let me double check if didn't mess up when converting from field to metric and will send you the case.

totto82:
>Yes, but 6-8 degrees change for water seems a bit too much. Let me double check if didn't mess up when converting from field to metric and will send you the case.

@goncalvesmachadoc did you double check. I don't remember getting any case. It would be very good if we could fix this before the release.

goncalvesmachadoc:
Hi @totto82 , considering this is already a great improvement in the solution, think should go ahead and get this fixed for the release.

atgeirr:
> should go ahead and get this fixed for the release.

I will merge this, further fixes can follow if identified.

PR #713, opened 2022-09-01
Bugfix: make TPFA version of local linearizer work for two-phase runs.

atgeirr:
No test currently exercise this, that will change when eventually OPM/opm-simulators#4017 gets merged.


atgeirr:
jenkins build this please

PR #714, opened 2022-09-01
Add support for directional relative permeabilities

hakonhagland:
Depends on https://github.com/OPM/opm-material/pull/532 and https://github.com/OPM/opm-common/pull/3113

### DIFF
--- a/opm/models/blackoil/blackoilintensivequantities.hh
+++ b/opm/models/blackoil/blackoilintensivequantities.hh
@@ -120,6 +120,28 @@ class BlackOilIntensiveQuantities
     using FluxIntensiveQuantities = typename FluxModule::FluxIntensiveQuantities;
     using DiffusionIntensiveQuantities = BlackOilDiffusionIntensiveQuantities<TypeTag, enableDiffusion>;
 
+    struct DirectionalMobility {
+        using array_type = std::array<Evaluation,numPhases>;
+        DirectionalMobility(const array_type& mX, const array_type& mY, const array_type& mZ) 
+            : mobilityX_{mX}, mobilityY_{mY}, mobilityZ_{mZ} {}
+        DirectionalMobility() : mobilityX_{}, mobilityY_{}, mobilityZ_{} {}
+        array_type& getArray(int index) {
+            switch(index) {
+                case 0:
+                    return mobilityX_;
+                case 1:
+                    return mobilityY_;
+                case 2:
+                    return mobilityZ_;
+                default:
+                    throw std::runtime_error("Unexpected mobility array index00");
+            }
+        }
+        array_type mobilityX_;
+        array_type mobilityY_;
+        array_type mobilityZ_;
+    };
+
 public:
     using FluidState = BlackOilFluidState<Evaluation,
                                           FluidSystem,
@@ -399,9 +421,14 @@ public:
         paramCache.updateAll(fluidState_);
 
         // compute the phase densities and transform the phase permeabilities into mobilities
-        constexpr int nmobilities = 4;
-        std::array<std::array<Evaluation,numPhases>*,nmobilities> mobilities
-           = {&mobility_, &mobilityX_, &mobilityY_, &mobilityZ_};
+        int nmobilities = 1;
+        std::vector<std::array<Evaluation,numPhases>*> mobilities = {&mobility_};
+        if (dirMob_) {
+            for (int i=0; i<3; i++) {
+                nmobilities += 1;
+                mobilities.push_back(&(dirMob_->getArray(i)));
+            }
+        }
         for (unsigned phaseIdx = 0; phaseIdx < numPhases; ++phaseIdx) {
             if (!FluidSystem::phaseIsActive(phaseIdx))
                 continue;
@@ -545,16 +572,22 @@ public:
     const Evaluation& mobility(unsigned phaseIdx, FaceDir::DirEnum facedir) const
     {
         using Dir = FaceDir::DirEnum;
-        switch(facedir) {
-            case Dir::XPlus:
-                return mobilityX_[phaseIdx];
-            case Dir::YPlus:
-                return mobilityY_[phaseIdx];
-            case Dir::ZPlus:
-                return mobilityZ_[phaseIdx];
-            default:
-                throw std::runtime_error("Unexpected face direction");
+        if (dirMob_) {
+            switch(facedir) {
+                case Dir::XPlus:
+                    return dirMob_->mobilityX_[phaseIdx];
+                case Dir::YPlus:
+                    return dirMob_->mobilityY_[phaseIdx];
+                case Dir::ZPlus:
+                    return dirMob_->mobilityZ_[phaseIdx];
+                default:
+                    throw std::runtime_error("Unexpected face direction");
+            }
         }
+        else {
+            return mobility_[phaseIdx];
+        }
+
     }
 
     /*!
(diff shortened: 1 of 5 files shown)


hakonhagland:
jenkins build this opm-material=532 opm-common=3113 please

opm/models/blackoil/blackoilintensivequantities.hh
```diff
@@ -588,6 +646,39 @@ private:
     friend BlackOilBrineIntensiveQuantities<TypeTag>;
     friend BlackOilMICPIntensiveQuantities<TypeTag>;
 
+    void updateRelperms(const Problem& problem, const MaterialLawParams& materialParams, unsigned globalSpaceIdx)
```
atgeirr, line 649: When it is spread across two functions, it is not easy to see that mobility_ and dirMob_ gets updated here. What about making this a static function and passing mobility_ and dirMob_ to it, that way it is obvious at the call site that those variables get updated?

opm/models/discretization/ecfv/ecfvstencil.hh
```diff
@@ -205,10 +209,34 @@ public:
         Scalar area() const
         { return area_; }
 
+        /*!
+         * \brief Returns the direction of the face
+         */
+        FaceDir::DirEnum dirId() const
```
atgeirr, line 215: For my current work on boundary conditions, it would be very useful if dirId() returned the underlying int instead, as I need to distinguish between the XMin and XPlus cases. Could we make that change, and add a free function faceDirFromDirId() with the implementation below?

opm/models/blackoil/blackoilintensivequantities.hh
```diff
@@ -597,6 +688,7 @@ private:
     Evaluation porosity_;
     Evaluation rockCompTransMultiplier_;
     std::array<Evaluation,numPhases> mobility_;
+    std::shared_ptr<DirectionalMobility> dirMob_;
```
atgeirr, line 691: I assume this is a shared_ptr and not a unique_ptr to keep the class copyable. But I think that changes the semantics of the class in a nontrivial way, as we will then not get a fully independent copy anymore. I am not set against it, but have you looked at the places where IQs are copied and checked that this is ok?

If we want to preserve deep copy semantics with a smart pointer it is possible, the Thor library has (or at least used to have) a CopiedPtr template that implemented it: https://bromeon.ch/libraries/thor/tutorials/v1.1/smartptr.html

Not sure that is necessary though.

atgeirr (changes requested):
What I looked at seems good, I just have a few questions and minor requests.

opm/models/blackoil/blackoilintensivequantities.hh
hakonhagland: Good idea! See update.

opm/models/discretization/ecfv/ecfvstencil.hh
hakonhagland: Yes, see update.

opm/models/blackoil/blackoilintensivequantities.hh
hakonhagland: Yes! That was to make the IQs copyable. I am not sure if there is an easy way to find all the places where the copy constructor and assignment operator for the IQ is used. I checked some files and looked at the compiler error messages and found one place in fvbaseelmentcontext.hh where they are copied: see updateIntensitiveQuantities_() at [line 562](https://github.com/OPM/opm-models/blob/331a3de9602691601916906f7cd19ccd13cce772/opm/models/discretization/common/fvbaseelementcontext.hh#L562)  and one place where the IQs  are reassigned, see updateStencil() at [line 140](https://github.com/OPM/opm-models/blob/331a3de9602691601916906f7cd19ccd13cce772/opm/models/discretization/common/fvbaseelementcontext.hh#L140). There are probably other places? 

In the updated PR, I changed the shared_ptr to unique_ptr and added a copy constructor and an assignment operator to the BlackOilIntensiveQuantities class.

atgeirr: I think it is not a good solution to add explicitly coded copy constructors and assignment operators. It adds to the maintenance overhead significantly. I am also not sure it is actually correct as implemented here in the copy constructor, it fails to copy most members, this just demonstrates the point that it is hard to maintain.

I think using a copyable smart pointer is the best solution, keeping the copy construction and assigment defaulted. I may have come across as skeptical of the shared_ptr, but I think it can be an acceptable solution, as long as we have considered the consequences. A smart pointer with copy semantics is also a fine solution I think, but requires us to create or adopt such a template class.

opm/models/discretization/ecfv/ecfvstencil.hh
atgeirr: Thanks, I have added the code identically to my own branch too, so it should be easy to rebase one once the other is merged.

opm/models/blackoil/blackoilintensivequantities.hh
hakonhagland: >  It adds to the maintenance overhead significantly

@atgeirr Yes, I agree.

> A smart pointer with copy semantics is also a fine solution I think, but requires us to create or adopt such a template class.

I'll give the "copyable pointer" a try first. In the added commit I try to implement a minimal functionality for a copyable pointer.

opm/models/blackoil/blackoilintensivequantities.hh
```diff
@@ -142,6 +144,59 @@ class BlackOilIntensiveQuantities
...
+            return *this;
+        }
+        // member access operator
+        T* operator->() const {return ptr_.get(); }
+        // boolean context operator
+        explicit operator bool() const noexcept {
+            return ptr_ ? true : false;
+        }
+        // get a reference to the stored value
+        T& get() const {return *ptr_;}
```
atgeirr, line 194: To be consistent with the unique_ptr interface, it should return the raw pointer instead (`return ptr_.get()`).

opm/models/blackoil/blackoilintensivequantities.hh
```diff
@@ -142,6 +144,59 @@ class BlackOilIntensiveQuantities
...
+    // of the unique_ptr each time we copy construct or assign to it from another BlackOilIntensiveQuantites.
+    // (On the other hand, if a copy could share the ptr with the original, a shared_ptr could be used instead and the
+    // wrapper would not be needed)
+    template <class T>
+    class CopyablePtr {
+    public:
+        CopyablePtr() : ptr_(nullptr) {}
+        CopyablePtr(const CopyablePtr& other) {
+            if (other) { // other does not contain a nullptr
+                ptr_ = std::make_unique<T>(other.get());
```
atgeirr, line 168: With get() returning a pointer (see below), this should change to take `*this`.

opm/models/blackoil/blackoilintensivequantities.hh
```diff
@@ -142,6 +144,59 @@ class BlackOilIntensiveQuantities
...
+            }
+            // else {ptr_ = nullptr;} // this is the default construction value
```
atgeirr, line 170: The else branch must be implemented to kill the current pointer. Copying from an empty smart pointer should leave the current object as a similarly empty one, just as in the assignment operator.

opm/models/blackoil/blackoilintensivequantities.hh
```diff
@@ -142,6 +144,59 @@ class BlackOilIntensiveQuantities
...
+    // assignment operators explicitly (which is needed when writing the custom copy constructor and assignment
+    // operators) which could become maintenance burden. For example, when adding a new base class (if that should
+    // be needed sometime in the future) to BlackOilIntensiveQuantites we could forget to update the copy
+    // constructor and assignment operators.
+    //
+    // We want each copy of the BlackOilIntensiveQuantites to be unique, (TODO: why?) so we have to make a copy
+    // of the unique_ptr each time we copy construct or assign to it from another BlackOilIntensiveQuantites.
+    // (On the other hand, if a copy could share the ptr with the original, a shared_ptr could be used instead and the
+    // wrapper would not be needed)
+    template <class T>
```
atgeirr, line 162: I would add one more warning to this class template comment: You should never use it polymorphically. (That would require a virtual `clone()` method etc.) It will only ever copy the static class type of the pointed to class.

Also, it should be put in a separate file, and have some unit tests.

opm/models/blackoil/blackoilintensivequantities.hh
hakonhagland: Good idea

opm/models/blackoil/blackoilintensivequantities.hh
hakonhagland: I thought since we are constructing a new object there will be no current object? Then, the default constructor will apply to the unique_ptr which will initialize it with a nullptr. Anyway, for clarity I have updated the code to use the else branch.

opm/models/blackoil/blackoilintensivequantities.hh
hakonhagland: Good idea, I have moved the class to opm-common, see https://github.com/OPM/opm-common/pull/3113 and added some tests..

opm/models/blackoil/blackoilintensivequantities.hh
atgeirr: You are correct in that there is no current object, my mistake! Maybe I thought it was a move constructor or something, not sure. But the default constructor is not invoked automatically, so you should always ensure all members are initialized.

bska:
This PR appears to [break](https://ci.opm-project.org/job/opm-models/155/console) the post-commit build.  Please fix or revert.

totto82:
Strange, it was green. Let me first double-check locally. Could there be an issue with the ordering and timing of when this and the PR in opm-material was merged?

hakonhagland:
> This PR appears to break the post-commit build

I see in the jenkins log:

> [5/119] Building CXX object CMakeFiles/reservoir_blackoil_vcfv.dir/tests/reservoir_blackoil_vcfv.cc.o
FAILED: CMakeFiles/reservoir_blackoil_vcfv.dir/tests/reservoir_blackoil_vcfv.cc.o 

But locally on my machine the test `reservoir_blackoil_vcfv` in opm-models seems not to be built. Any ideas how I can build and test this locally?

bska:
> locally on my machine the test `reservoir_blackoil_vcfv` in opm-models seems not to be built. Any ideas how I can build and test this locally?

Configure all your module builds with the
```sh
-DBUILD_TESTING:BOOL=True
```
cmake option.

bska:
> Could there be an issue with the ordering and timing of when this and the PR in opm-material was merged?

I suppose, but I get the same result locally when I include `BUILD_TESTING` in my configuration options.

hakonhagland:
>Configure all your module builds with the
>
>-DBUILD_TESTING:BOOL=True

@bska Thanks for the information!

> This PR appears to break the post-commit build. Please fix or revert.

This has been addressed in  https://github.com/OPM/opm-models/pull/727. Thanks to @totto82 for ideas on how to solve this.

PR #715, opened 2022-09-09
fixed: respect rule-of-three for fvbaseintensivequantities

akva2:
is no need to explicitly declare the cc here


akva2:
jenkins build this please

blattms:
Please help me: Why do we need the default constructor implementation if we don't even zero-initialize?

akva2:
precisely ;) read the diff again.

akva2:
jenkins build this please

PR #718, opened 2022-09-16
Boundary condition support for TpfaLinearizer

atgeirr:
Must be combined with OPM/opm-simulators#4119


atgeirr:
I believe this is ready for merging.

atgeirr:
jenkins build this opm-simulators=4119 please

opm/models/discretization/ecfv/ecfvstencil.hh
```diff
@@ -205,10 +210,45 @@ public:
...
+            switch(dirId_) {
+                case 0:
+                case 1:
+                    return FaceDir::XPlus;
+                case 2:
+                case 3:
+                    return FaceDir::YPlus;
+                case 4:
+                case 5:
+                    return FaceDir::ZPlus;
```
bska, line 241: Do the `*Minus` directions not make sense in this context?

atgeirr: This part is actually for the handling of anisotropic relative permeabilities, and overlaps with that PR. It's just that both had similar needs, so I duplicated that part. So I guess the question should be asked in #714

bska: > This part is actually for the handling of anisotropic relative permeabilities,

Okay, in that case it's definitely wrong to not include the `*Minus` directions.

atgeirr:
jenkins build this opm-simulators=4119 please

atgeirr:
Test failure due to clang accepting `const auto bool x = false` but gcc not. It was not intended in any case, so amended.

atgeirr:
jenkins build this opm-simulators=4119 please

opm/models/blackoil/blackoillocalresidualtpfa.hh
```diff
@@ -414,6 +415,117 @@ public:
...
+                                    const IntensiveQuantities& insideIntQuants,
+                                    unsigned globalSpaceIdx,
+                                    unsigned timeIdx)
+    {
+        if (bdyInfo.type == BCType::RATE) {
+            computeBoundaryFluxRate(bdyFlux, bdyInfo);
+        } else if (bdyInfo.type == BCType::FREE) {
+            computeBoundaryFluxFree(problem, bdyFlux, bdyInfo, insideIntQuants, globalSpaceIdx, timeIdx);
+        } else {
+            throw std::logic_error("Unknown boundary condition type in computeBoundaryFlux().");
```
bska, line 431: Could you add the actual type here too? E.g.,
```
... type " + std::to_string(static_cast<int>(bdyInfo.type)) + " in ...
```
or is it not that easy?

opm/models/blackoil/blackoillocalresidualtpfa.hh
```diff
@@ -414,6 +415,117 @@ public:
...
+            // mass conservation
+            if (pBoundary < pInside) {
+                // outflux
+                const auto& invB = getInvB_<FluidSystem, FluidState, Evaluation>(insideIntQuants.fluidState(), phaseIdx, pvtRegionIdx);
+                Evaluation surfaceVolumeFlux = invB * volumeFlux[phaseIdx];
+                evalPhaseFluxes_<Evaluation>(tmp,
+                                             phaseIdx,
+                                             insideIntQuants.pvtRegionIndex(),
+                                             surfaceVolumeFlux,
+                                             insideIntQuants.fluidState());
```
bska, line 490: Does template argument deduction not work here?  It would appear that `surfaceVolumeFlux` already is an `Evaluation` object.

bska (commented):
Looks good for the most part.  I still think that `faceDirFromDirId()` should handle the `*Minus` cases too, but that's a topic for another day.  Other than that I have only minor comments here.

bska:
We're getting [two new warnings](https://ci.opm-project.org/job/opm-models-PR-builder/233/gcc/new/folder.-1825958716/) from this PR.  Would you mind addressing those too?

opm/models/blackoil/blackoillocalresidualtpfa.hh
atgeirr: It does not, since the argument explicitly provided is the `UpEval` type, and the `surfaceVolumeFlux` is the `Eval` type. See the call below for the constrasting case, it also passes an Evaluation, but the UpEval is different.

atgeirr:
Have tried to address the review comments.

atgeirr:
jenkins build this please

atgeirr:
jenkins build this opm-simulators=4119 please

atgeirr:
jenkins build this opm-simulators=4119 please

atgeirr:
jenkins build this opm-simulators=4119 please

atgeirr:
jenkins build this opm-simulators=4119 please

bska (approved):
This looks good to me now and I'll merge this into the master branch.

PR #720, opened 2022-09-19
blackoilextbomodules: put parameters in separate class

akva2:
limits the amount of static variable declarations

### DIFF
--- a/opm/models/blackoil/blackoilextboparams.hh
+++ b/opm/models/blackoil/blackoilextboparams.hh
@@ -1,0 +1,68 @@
+// -*- mode: C++; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*-
+// vi: set et ts=4 sw=4 sts=4:
+/*
+  This file is part of the Open Porous Media project (OPM).
+
+  OPM is free software: you can redistribute it and/or modify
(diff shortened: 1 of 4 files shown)


akva2:
jenkins build this please

opm/models/blackoil/blackoilextboparams.hh
```diff
@@ -0,0 +1,67 @@
...
+#define EWOMS_BLACK_OIL_EXTBO_PARAMS_HH
+
+#include <opm/material/common/Tabulated1DFunction.hpp>
+#include <opm/material/common/UniformXTabulated2DFunction.hpp>
+
+#include <vector>
+
+namespace Opm {
+
+template<class Scalar>
```
blattms, line 42: Some doxygen documentation at least about the class might be handy for fellow developers.

blattms (approved):
Looks good, would prefer some documentiation for the new class though.

opm/models/blackoil/blackoilextboparams.hh
akva2: i totally agree. problem is i'm a bit in deep water as to what the various parameters are for :/ but at least i can doxy what the class is used for.

akva2:
jenkins build this please

PR #724, opened 2022-09-19
blackoilsolventmodules: put parameters in separate class

akva2:
limits the amount of static variable declarations


blattms:
Approved and merging

PR #726, opened 2022-09-25
Initialize Rvw

goncalvesmachadoc:


goncalvesmachadoc:
jenkins build this please

blattms (approved):
Yep,  this ConditionalStorage seems tricky (no default initialiaztion). I will happily merge this if you add a short explanation in the description what this fixes for future reference and the manual author.

goncalvesmachadoc:
It doesn't fix a problem per se, I just thought it should be there.

PR #727, opened 2022-09-26
Move updateRelperms() to the problem class

hakonhagland:
Background: This PR is trying to fix the build test problem in https://github.com/OPM/opm-models/pull/714 that was discovered after it was merged into master.

Moves the `updateRelperms()` method in `BlackOilIntensiveQuantities` to the `xxxProblem` class. This is a more natural place for this method and it avoids including `EclMaterialManager` into the `BlackOilIntensiveQuantities`. The `DirectionalMobility` struct is moved to a separate file such that it can be included from both the `xxxProblem` files and the `BlackOilIntensiveQuantities`.

See also the companion PR https://github.com/OPM/opm-simulators/pull/4132 in opm-simulators.

### DIFF
--- a/opm/models/blackoil/blackoilintensivequantities.hh
+++ b/opm/models/blackoil/blackoilintensivequantities.hh
@@ -234,7 +234,7 @@ public:
         std::array<Evaluation, numPhases> pC;
         const auto& materialParams = problem.materialLawParams(globalSpaceIdx);
         MaterialLaw::capillaryPressures(pC, materialParams, fluidState_);
-        problem.template updateRelperms<FluidState>(mobility_, dirMob_, fluidState_, globalSpaceIdx);
+        problem.updateRelperms(mobility_, dirMob_, fluidState_, globalSpaceIdx);
 
         // oil is the reference phase for pressure
         if (priVars.primaryVarsMeaning() == PrimaryVariables::Sw_pg_Rv || priVars.primaryVarsMeaning() == PrimaryVariables::Rvw_pg_Rv) {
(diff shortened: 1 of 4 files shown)


hakonhagland:
jenkins build this opm-simulators=4132 please

hakonhagland:
Seems like the jenkins build will not start?

@akva2 Could you add me to the white list for jenkins build triggering?

bska:
jenkins build this opm-simulators=4132 please

opm/models/blackoil/blackoilintensivequantities.hh
```diff
@@ -258,7 +234,7 @@ public:
         std::array<Evaluation, numPhases> pC;
         const auto& materialParams = problem.materialLawParams(globalSpaceIdx);
         MaterialLaw::capillaryPressures(pC, materialParams, fluidState_);
-        BlackOilIntensiveQuantities::updateRelperms(mobility_, dirMob_, fluidState_, problem, materialParams, globalSpaceIdx);
+        problem.template updateRelperms<FluidState>(mobility_, dirMob_, fluidState_, globalSpaceIdx);
```
bska, line 237: Does template argument deduction of `FluidState` not work here?  Why do you need an explicit template argument?

hakonhagland: Yes you are right, I have updated the PR with the simpler statement

hakonhagland:
I see the following error in the [jenkins build](https://ci.opm-project.org/job/opm-models-PR-builder/lastBuild/console):
```
/var/lib/jenkins/workspace/opm-models-PR-builder/mpi/install/include/opm/models/common/multiphasebaseproblem.hh:33:10: fatal error: opm/models/common/directionalmobility.hh: No such file or directory
   33 | #include <opm/models/common/directionalmobility.hh>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```
Maybe I need to add the new header file (`directionalmobility.hh`) to `CMakeLists_files.cmake` ?

bska:
jenkins build this opm-simulators=4132 please

totto82:
Thanks for the fix.

PR #730, opened 2022-10-10
Fixes compilation with DUNE 2.9 and 2.6

blattms:
Compilation of the tests has broken since month for DUNE 2.6.
Somewhat related to OPM/opm-grid#587, but should also be mergable independently.


atgeirr (approved):

GitPaean:
It looks like the PR breaks running with DUNE-2.6, at least from my testing. 

reverting this PR for testing purpose.

PR #732, opened 2022-10-11
add option to make well assembly based on only wells

hnil:
added option to do assembly on of well contributions to reservoir only on reservoir cells.

### DIFF
--- a/opm/models/discretization/common/tpfalinearizer.hh
+++ b/opm/models/discretization/common/tpfalinearizer.hh
@@ -48,7 +48,13 @@
 #include <set>
 #include <exception>   // current_exception, rethrow_exception
 #include <mutex>
-
+namespace Opm::Properties {
+    template<class TypeTag, class MyTypeTag>
+    struct AssembleWellLocal {
+        using type = bool;
+        static constexpr type value = true;
+    };
+}
 
 namespace Opm {
 // forward declarations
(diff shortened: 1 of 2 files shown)


opm/models/discretization/common/tpfalinearizer.hh
```diff
@@ -483,6 +492,9 @@ private:
             }
         } // end of loop for cell globI.
 
+        if( not(well_local_)){
```
atgeirr, line 495: Nitpicks: space after is, space before start brace.
Next line: typo, should be addReservoirSourceTerms(), space after comma.

atgeirr: I agree, we had a plan to start with opm-common, but there is a significant change-set that will come with it. I would happily go along with any reasonable approacy.

akva2: while I don't agree as I find clang-format causes more grief than it resolves  (but I accept that it will happen), the only  sane way to go about it imo is to require new code and/or code touched by diffs to be updated. Doing one big bulk update is going to be extremely annoying.

atgeirr: > require new code and/or code touched by diffs to be updated

Which is what I have been doing all the time...

PR #733, opened 2022-10-13
Use current dir for output by default.

atgeirr:
For OPM Flow, this only affects VTK output.

### DIFF
--- a/opm/models/discretization/common/fvbaseproblem.hh
+++ b/opm/models/discretization/common/fvbaseproblem.hh
@@ -206,7 +206,7 @@ public:
         std::string outputDir = EWOMS_GET_PARAM(TypeTag, std::string, OutputDir);
 
         if (outputDir == "")
-            throw std::runtime_error("No directory for output specified");
+            outputDir = ".";
 
         // TODO: replace this by std::filesystem once we require c++-2017
         struct stat st;


atgeirr:
jenkins build this please

bska (commented):
Looks good to me, but I'm not sure I have a full overview of all callers so I'm going to defer to someone else to approve/merge the PR.  Incidentally, with might consider switching this function over to using `std::filesystem` instead of direct Posix calls for portability.

blattms:
If this is flow, why would you want to output to the current directory? All other files end up somewhere else by default, don't they?

Hence shouldn't this be set by opm-simulators to the data file directory or whatever is specified there as the output directory, if it is not specified?

atgeirr:
> Hence shouldn't this be set by opm-simulators to the data file directory or whatever is specified there as the output directory, if it is not specified?

This code is opm-models code, and it only sees the OutputDir parameter. The defaults for that parameter also varies from one class to the next derived one, both "." and "" are used at different levels, for flow it ends up being "". That results in an exception here if you request vtk output. The only effect of this code on flow is for vtk output. For other opm-models simulators, it may result in writing output to the current directory instead of throwing, I think that is an improvement.

> Incidentally, with might consider switching this function over to using `std::filesystem` instead of direct Posix calls for portability.

We already have such code in opm-simulators, but wanted to keep the PR tiny...

atgeirr:
@blattms is my reasoning satisfactory, or do you want another solution?

blattms:
I'll back off then

PR #735, opened 2022-10-14
Register all (not just some) of the vtk parameters unconditionally.

atgeirr:
No reason to treat the diffusion parameters differently. This caused the FlowEarlyBird (TypeTag) set of parameters to differ from the one used for Flow blackoil, since that (using TpfaLinearizer) does not activate diffusion.

Closes OPM/opm-common#3168


atgeirr:
jenkins build this please

bska (approved):

PR #741, opened 2022-10-19
Put properties of Newton method into separate header for reuse.

blattms:
Now they can be reused in opm-simulators without any hassle.
There will be a downstream change.

### DIFF
--- a/CMakeLists_files.cmake
+++ b/CMakeLists_files.cmake
@@ -185,7 +185,7 @@ list (APPEND PUBLIC_HEADER_FILES
              opm/models/io/vtkblackoilenergymodule.hh
              opm/models/io/baseoutputmodule.hh
              opm/models/io/vtkblackoilpolymermodule.hh
-						 opm/models/io/vtkblackoilmicpmodule.hh
+             opm/models/io/vtkblackoilmicpmodule.hh
              opm/models/ncp/ncpmodel.hh
              opm/models/ncp/ncpindices.hh
              opm/models/ncp/ncpextensivequantities.hh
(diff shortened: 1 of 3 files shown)


blattms:
jenkins build this opm-simulators=4185 please

blattms:
Yeah, what do know.
Properties with the same name have to registered with the exact same arguments no matter whether the template parameter is different. That is a bit of a pity as 
- we need to used the same number of iterations both in models an in simulators (not a big deal, we can simply used 20 for both)
- The bigger problem is the usage string (models: "The maximum number of Newton iterations per time step" and simulators: "The maximum number of Newton iterations per time step used by flow")

"by flow" is of coarse a bit redundant information, but if we skip that we would need to skip it everywhere for consistency. Which in particular means changing all these in simulators(NonlinearSolverEbos) and probably more;
```
                EWOMS_REGISTER_PARAM(TypeTag, Scalar, NewtonMaxRelax, "The maximum relaxation factor of a Newton iteration used by flow");
                EWOMS_REGISTER_PARAM(TypeTag, int, NewtonMaxIterations, "The maximum number of Newton iterations per time step used by flow");
                EWOMS_REGISTER_PARAM(TypeTag, int, NewtonMinIterations, "The minimum number of Newton iterations per time step used by flow");
                EWOMS_REGISTER_PARAM(TypeTag, std::string, NewtonRelaxationType, "The type of relaxation used by flow's Newton method");
```

Comments?

blattms:
Last commit in opm-simulators accomplishes the  latter (removing references of flow for Newton).

blattms:
jenkins build this opm-simulators=4185 please

atgeirr:
> Last commit in opm-simulators accomplishes the latter (removing references of flow for Newton).

I am in favour of doing that.

blattms:
Then we should probably merge this and the other one.

blattms:
jenkins build this opm-simulators=4185 please

bska (approved):
Thanks a lot.  This looks good to me now and I'll merge this and its downstream companion int master now.

PR #746, opened 2022-10-21
fix bug in boundary conditions for energy by adding the scaling factor

totto82:
Since this is a bug fix I add a release flag. 

Case with free boundary on the right and an injector injecting 10 C on the left. 

Current master
image

With PR
image


totto82:
jenkins build this please

PR #748, opened 2022-10-27
Add interface for updating stored parameters in linearizers.

atgeirr:
Implement for the TpfaLinearizer to update transmissibilities. The regular linearizer stores no such parameters and has an empty implementations.

Necessary to handle correctly cases with e.g. MULTFLT in the SCHEDULE section. Downstream PR adds call to this method.

### DIFF
--- a/opm/models/discretization/common/fvbaselinearizer.hh
+++ b/opm/models/discretization/common/fvbaselinearizer.hh
@@ -282,6 +282,11 @@ public:
         return linearizationType_;
     };
 
+    void updateDiscretizationParameters()
+    {
+        // This linearizer stores no such parameters.
+    }
+
     /*!
      * \brief Returns the map of constraint degrees of freedom.
      *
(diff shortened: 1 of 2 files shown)


atgeirr:
jenkins build this opm-simulators=4207 please

blattms (approved):
Approved assuming that this fixes the model

bska:
I tested this and the downstream companion PR on the field case that originally revealed the problem.  I can confirm that this PR indeed fixes the problem.

atgeirr:
The only test failure is the one added in opm-simulators, as it has no reference data. Will do a data update and merge when green.

atgeirr:
jenkins build this opm-simulators=4207 update_data please

atgeirr:
jenkins build this opm-simulators=4207 opm-tests=838 please

bska:
jenkins build this opm-simulators=4207 opm-tests=839 please

bska:
jenkins build this please

PR #755, opened 2022-11-14
Fix test_quadrature with dune-alugrid

akva2:
Related to https://github.com/OPM/opm-common/pull/3219


akva2:

blattms:
I guess the fix was dereferencing a unique_ptr for DUNE >=2.7 and using that after throwing away the temporary unique_ptr (as this deallocated the the grid at the very same moment)?

akva2:
yes we need to hold on to the unique_ptr or no refs and object immediately died.

PR #758, opened 2022-11-17
FLOWS and FLORES support

daavid00:
See https://github.com/OPM/opm-simulators/pull/4275 for rationale.

### DIFF
--- a/opm/models/blackoil/blackoillocalresidualtpfa.hh
+++ b/opm/models/blackoil/blackoillocalresidualtpfa.hh
@@ -203,6 +203,7 @@ public:
      * read from the extensive quantities of the element context.
      */
     static void computeFlux(RateVector& flux,
+                            RateVector& darcy,
                             const Problem& problem,
                             const unsigned globalIndexIn,
                             const unsigned globalIndexEx,
@@ -213,6 +214,7 @@ public:
                             const FaceDir::DirEnum facedir)
     {
         flux = 0.0;
+        darcy = 0.0;
         Scalar Vin = problem.model().dofTotalVolume(globalIndexIn);
         Scalar Vex = problem.model().dofTotalVolume(globalIndexEx);
 
@@ -236,6 +238,7 @@ public:
         Scalar distZ = zIn - zEx; // NB could be precalculated
 
         calculateFluxes_(flux,
+                         darcy,
                          intQuantsIn,
                          intQuantsEx,
                          Vin,
@@ -260,6 +263,7 @@ public:
         assert(timeIdx == 0);
 
         flux = 0.0;
+        RateVector darcy = 0.0;
         // need for dary flux calculation
         const auto& problem = elemCtx.problem();
         const auto& stencil = elemCtx.stencil(timeIdx);
@@ -304,6 +308,7 @@ public:
         Scalar distZ = zIn - zEx;
 
         calculateFluxes_(flux,
+                         darcy,
                          intQuantsIn,
                          intQuantsEx,
                          Vin,
@@ -318,6 +323,7 @@ public:
     }
 
     static void calculateFluxes_(RateVector& flux,
+                                 RateVector& darcy,
                                  const IntensiveQuantities& intQuantsIn,
                                  const IntensiveQuantities& intQuantsEx,
                                  const Scalar& Vin,
@@ -371,6 +377,7 @@ public:
                     darcyFlux = pressureDifference *
                        (Toolbox::value(up.mobility(phaseIdx, facedir)) * Toolbox::value(transMult) * (-trans / faceArea));
             }
+            darcy[phaseIdx] = darcyFlux.value() * faceArea; // For the FLORES fluxes 
 
             unsigned pvtRegionIdx = up.pvtRegionIndex();
             // if (upIdx == globalFocusDofIdx){
--- a/opm/models/discretization/common/fvbaselinearizer.hh
+++ b/opm/models/discretization/common/fvbaselinearizer.hh
@@ -32,6 +32,7 @@
 #include "linearizationtype.hh"
 
 #include <opm/common/Exceptions.hpp>
+#include <opm/grid/utility/SparseTable.hpp>
 
 #include <opm/models/parallel/gridcommhandles.hh>
 #include <opm/models/parallel/threadmanager.hh>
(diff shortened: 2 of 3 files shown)


opm/models/blackoil/blackoillocalresidualtpfa.hh
```diff
@@ -371,6 +377,7 @@ public:
                     darcyFlux = pressureDifference *
                        (Toolbox::value(up.mobility(phaseIdx, facedir)) * Toolbox::value(transMult) * (-trans / faceArea));
             }
+            darcy[phaseIdx]=darcyFlux.value(); // For the FLORES fluxes
```
atgeirr, line 380: Nitpick: space around `=`.

opm/models/discretization/common/fvbaselinearizer.hh
```diff
@@ -295,6 +296,22 @@ public:
     const std::map<unsigned, Constraints>& constraintsMap() const
     { return constraintsMap_; }
 
+    /*!
+     * \brief Return constant reference to the flowsInfo.
+     *
+     * (This object it has been only implemented for the tpfalinearizer.)
```
atgeirr, line 302: Nitpick: remove 'it'.

opm/models/discretization/common/tpfalinearizer.hh
```diff
@@ -266,6 +266,26 @@ public:
...
+     *
+     * (This object is only non-empty if the FLOWS keyword is true.)
+     */
+    const auto& getFlowsInfo() const{
+
+        return flowsInfo_;
+    }   
+
+    /*!
+     * \brief Return constant reference to the flresInfo.
```
atgeirr, line 280: Since the idea is to support FLORES, the data member should be called `floresInfo_` instead of `flresInfo_`, same with the method name.

opm/models/discretization/common/tpfalinearizer.hh
```diff
@@ -428,6 +507,8 @@ private:
         const bool well_local = true;
         resetSystem_();
         unsigned numCells = model_().numTotalDof();
+        const bool& enableFlows = simulator_().problem().eclWriter()->eclOutputModule().hasFlows();
+        const bool& enableFlres = simulator_().problem().eclWriter()->eclOutputModule().hasFlres();
```
atgeirr, line 511: Similar situation with names here, it would be better to use `enableFlores` and `hasFlores()`.

opm/models/discretization/common/tpfalinearizer.hh
```diff
@@ -450,15 +532,27 @@ private:
...
+                       adres, darcy, problem_(), globI, globJ, intQuantsIn, intQuantsEx,
                            nbInfo.trans, nbInfo.faceArea, nbInfo.faceDirection);
                 adres *= nbInfo.faceArea;
+                if (enableFlows) {
+                    for (unsigned phaseIdx = 0; phaseIdx < numEq; ++ phaseIdx) {
+                        flowsInfo_[globI][loc].flow[phaseIdx] = adres[phaseIdx].value();
+                    }
+                }
+                if (enableFlres) {
+                    darcy *= nbInfo.faceArea;
```
atgeirr, line 551: Why is this not already multiplied in? It looks like something that could easily cause errors. Would it not be better if the `darcy` objects were the fluxes, not the velocities? In any case, changing the name from `darcy` to `darcyFlux` or `darcyVelocity` (depending on your choice here) would make it clearer, and reduce the risk of future errors.

opm/models/discretization/common/tpfalinearizer.hh
```diff
@@ -405,6 +428,62 @@ private:
...
+        }
+        const auto& model = model_();
+        const auto& nncOutput = simulator_().problem().eclWriter()->getOutputNnc();
+        Stencil stencil(gridView_(), model_().dofMapper());
+        unsigned numCells = model.numTotalDof();
+        std::vector<FlowInfo> loc_flinfo;
+        unsigned int nncId = 0;
+        VectorBlock flow(0.0);
+
+        if (anyFlows)
```
atgeirr, line 448: Please add braces to these if statements. Their lack makes it look like the for statement below them only has a single statement at first glance.

opm/models/discretization/common/tpfalinearizer.hh
```diff
@@ -405,6 +428,62 @@ private:
...
+                loc_flinfo.resize(stencil.numDof() - 1);
+                for (unsigned dofIdx = 0; dofIdx < stencil.numDof(); ++dofIdx) {
+                    unsigned neighborIdx = stencil.globalSpaceIndex(dofIdx);
+                    if (dofIdx > 0) {
+                        const auto scvfIdx = dofIdx - 1;
+                        const auto& scvf = stencil.interiorFace(scvfIdx);
+                        int faceId = scvf.dirId();
+                        const int cartMyIdx = simulator_().vanguard().cartesianIndex(myIdx);
+                        const int cartNeighborIdx = simulator_().vanguard().cartesianIndex(neighborIdx);
+                        for (unsigned nncIdx = 0; nncIdx < nncOutput.size(); ++nncIdx) {
```
atgeirr, line 466: This is potentially very slow. Sometimes an entire case uses only NNCs for connections, since that allows us to run on arbitrary graphs with full flexibility. In such cases, this will be O(n^2). I suggest creating a structure that supports fast lookup. This can be for example `std::unordered_multimap<int, std::pair<int, int>>` from first cartesian index of each NNC connection to  {second cartesian index, the nncIdx index of this connection in the nncOutput array}, but there are many possibilities here.

atgeirr (commented):
I think the overall thinking and structure here is good, and I am in favour of merging it once some issues are adressed.

Also: is there any impact on performance when not using this feature?

daavid00:
> I think the overall thinking and structure here is good, and I am in favour of merging it once some issues are adressed.
> 
> Also: is there any impact on performance when not using this feature?

Many thanks @atgeirr for your comments, I have addressed them. Here are the results of running the benchmark:

https://github.com/OPM/opm-common/pull/3224#issuecomment-1404157035

atgeirr:
Thanks for updating, I am now happy with the changes here. I think the opm-simulators part still needs to be looked at before merging, though.

bska:
Would you mind rebasing this PR onto the top of the current master sources please?  That way we can remove the merge commit in the middle of this commit series.

daavid00:
> Would you mind rebasing this PR onto the top of the current master sources please? That way we can remove the merge commit in the middle of this commit series.

Now they (models and simulators) are rebase, now @totto82 will run a Jenkins to double check everything is fine, many thanks.

totto82:
jenkins build this opm-simulators=4275 please

bska (approved):
Thanks a lot for the updates. I'll merge into master.

PR #761, opened 2022-11-30
Allow updateStoredTransmissibilities() before createMatrix_().

atgeirr:
A bug was triggered by a case placing MULTFLT at the start of the SCHEDULE. Then the required structures would be empty and the simulator bombed. With the change, we increase robustness and address this.


atgeirr:
jenkins build this please

bska:
jenkins build this please

atgeirr:
Self-merging this small fix.

PR #763, opened 2022-12-02
Support dissolved gas in water

totto82:
With GASINWAT enabled gas in allowed to dissolved into water. 
Mutual dissolution in water and oil is not supported. 
The idea there is to provide a function/table that map rs -> rsw

### DIFF
--- a/opm/models/blackoil/blackoildiffusionmodule.hh
+++ b/opm/models/blackoil/blackoildiffusionmodule.hh
@@ -123,7 +123,12 @@ public:
             }
 
             // no diffusion in water for blackoil models
-            if (FluidSystem::waterPhaseIdx == phaseIdx) {
+            if (!FluidSystem::enableDissolvedGasInWater() && FluidSystem::waterPhaseIdx == phaseIdx) {
+                continue;
+            }
+
+            // no diffusion in gas phase in water + gas system. 
+            if (FluidSystem::gasPhaseIdx == phaseIdx && !FluidSystem::phaseIsActive(FluidSystem::oilPhaseIdx)) {
                 continue;
             }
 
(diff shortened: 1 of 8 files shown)


totto82:
Depends on https://github.com/OPM/opm-common/pull/3217 and OPM/opm-material#543

opm/models/blackoil/blackoilprimaryvariables.hh
```diff
@@ -127,6 +127,7 @@ public:
     enum class WaterMeaning {
         Sw,  // water saturation
         Rvw, // vaporized water
+        Rsw, // dissolved gas in water
```
atgeirr, line 130: I am not sure about putting this here. Is it possible to have a 3-phase situation? What would then the gas variable be? I assume gas can either dissolve in water or oil, not both?

opm/models/blackoil/blackoilprimaryvariables.hh
```diff
@@ -603,6 +619,33 @@ public:
                 }
                 break;
             }
+            case WaterMeaning::Rsw:
+            {
+                // Gas phase not present. The hydrocarbon gas phase
+                // appears as soon as more of the water component is present in the water phase
```
atgeirr, line 625: You mean gas component I assume.

opm/models/blackoil/blackoilprimaryvariables.hh
```diff
@@ -578,6 +575,25 @@ public:
                     changed = true;
                     break;
                 }
+                if(sg < -eps && sw > eps && FluidSystem::enableDissolvedGasInWater()) {
```
atgeirr, line 578: These sections could do with some comments explaining what is going on.

atgeirr (commented):
I have some questions, but mostly looks fine.

opm/models/blackoil/blackoilprimaryvariables.hh
totto82: Then we should use Rs as primary variable and Rsw should be a secondary variable dependent on Rs.

opm/models/blackoil/blackoildiffusionmodule.hh
```diff
@@ -147,6 +147,11 @@ public:
                 convFactor = fraction(pvtRegionIndex) / (1.0 + rvAvg*fraction(pvtRegionIndex));
                 diffR = fluidStateI.Rv() - Toolbox::value(fluidStateJ.Rv());
             }
+            if (FluidSystem::enableDissolvedGasInWater() && phaseIdx == FluidSystem::waterPhaseIdx) {
```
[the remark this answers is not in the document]
totto82: I never tested the diffusion part. I will do that and fix.

opm/models/blackoil/blackoilprimaryvariables.hh
```diff
@@ -349,7 +349,8 @@ public:
             primaryVarsMeaningPressure_ = PressureMeaning::Pg;
         } else if (FluidSystem::phaseIsActive(oilPhaseIdx)) {
             primaryVarsMeaningPressure_ = PressureMeaning::Po;
-        } else if (FluidSystem::phaseIsActive(gasPhaseIdx)) {
+        //} else if (FluidSystem::phaseIsActive(gasPhaseIdx)) {
+        } else if (FluidSystem::phaseIsActive(gasPhaseIdx)  && gasPresent) {
```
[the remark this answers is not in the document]
totto82: This will indeed change the 2 phase, gas-water case, with only water present to use the water pressure. I will add a check for enableGasInWater to make sure that we dont change anything else in the PR. We may want to re-think which pressure we would like to solve for in the "onlyWater" special case. Maybe water?

opm/models/blackoil/blackoilprimaryvariables.hh
atgeirr: Not sure I follow you, how can you easily calculate Rsw from Rs? I guess thermodynamically this is well-defined, but looks to me like you requires flash-like calculations to decide how the gas is distributed when dissolved in two different phases.

totto82: Yes, you will need "flash-like" calculations. But we can do it the standard black-oil way. i.e. provide it as a table. So for the 3-phase case with Sg = 0, rs will be the primary variable together with Sw and Po, and rsw = Rsw(pressure, temperature?, rs) So rsw will only be a primary-variable in the gas-water case and hence it belongs in the "WaterMeaning"

opm/models/blackoil/blackoildiffusionmodule.hh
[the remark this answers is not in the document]
totto82: I have now also enabled diffusion in the gas in water model (not only some drive-by adding of code).

opm/models/blackoil/blackoildiffusionmodule.hh
```diff
@@ -147,6 +147,11 @@ public:
...
+                Evaluation rsAvg = (fluidStateI.Rsw() + Toolbox::value(fluidStateJ.Rsw())) / 2;
+                convFactor = 1.0 / (fraction(pvtRegionIndex) + rsAvg);
```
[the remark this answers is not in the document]
totto82: fixed

opm/models/blackoil/blackoilprimaryvariables.hh
totto82: yes, updated

opm/models/blackoil/blackoilprimaryvariables.hh
atgeirr: Thanks, this makes sense. Is this table approach already implemented?

atgeirr (approved):

PR #766, opened 2022-12-13
remove namespace qualifiers from symbols in Opm namespace

akva2:
Some more of these..


bska (approved):

PR #770, opened 2022-12-15
Matrix direct access

hnil:
Add direct access to matrix elements in tpfalinearizer. Seems to give a speedup of 10% for this part. off assembly. Together with not using local well assembly more speedup is seen.

Edit: By direct access to matrix elements, we mean that instead of using `m[row][col]` to access the matrix block to be modified (via the `addToBlock(row, col)` call), we store pointers to the relevant block. The `m[row][col]` requires a search on the relevant row to find the correct block matching the requested column number.

### DIFF
--- a/opm/models/discretization/common/tpfalinearizer.hh
+++ b/opm/models/discretization/common/tpfalinearizer.hh
@@ -389,9 +389,9 @@ private:
         jacobian_->reserve(sparsityPattern);
         for (unsigned globI = 0; globI < numCells; globI++) {
             const auto& nbInfos = neighborInfo_[globI];
-            diagMatAddress_[globI] = jacobian_->blockAddress(globI,globI);           
+            diagMatAddress_[globI] = jacobian_->blockAddress(globI, globI);
             for (auto& nbInfo : nbInfos) {
-                nbInfo.matAddress = jacobian_->blockAddress(nbInfo.neighbor,globI);
+                nbInfo.matBlockAddress = jacobian_->blockAddress(nbInfo.neighbor, globI);
             }
         }
     }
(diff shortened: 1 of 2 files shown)


opm/models/discretization/common/tpfalinearizer.hh
```diff
@@ -429,33 +436,35 @@ private:
...
+                //     throw std::logic_error("Missing updated intensive quantities for cell " + std::to_string(globJ) + " when assembling fluxes for cell " + std::to_string(globI));
+                // }
                 const IntensiveQuantities& intQuantsEx = *intQuantsExP;
                 LocalResidual::computeFlux(
                        adres, problem_(), globI, globJ, intQuantsIn, intQuantsEx,
                            nbInfo.trans, nbInfo.faceArea, nbInfo.faceDirection);
                 adres *= nbInfo.faceArea;
                 setResAndJacobi(res, bMat, adres);
                 residual_[globI] += res;
-                jacobian_->addToBlock(globI, globI, bMat);
```
atgeirr, line 456: Remove the old lines instead of commenting them.

opm/models/discretization/common/tpfalinearizer.hh
```diff
@@ -546,8 +558,10 @@ private:
         double trans;
         double faceArea;
         FaceDir::DirEnum faceDirection;
+        MatrixBlock* matAddress;
```
atgeirr, line 561: Prefer a name for the variable that includes "block",  "matBlockAddress".

atgeirr (approved):
Very nice, just nitpicks.

atgeirr:
jenkins build this please

blattms:
Please improve the description of this PR such that others not in the loop have a chance to understand what is done here. When I hear direct access, I hardly imagine what is done here (pointers to the diagonal etc.). Also what is "local well assembly"?

Please keep in mind that there are people writing documentation and making good descriptions of PRs  saves others a lot more time that it costs.

atgeirr:
> Also what is "local well assembly"?

I agree it is not a good term. See #732. This refers to the notion of inserting the well source terms at the same time as we assemble the rest of the terms for a cell (which is what we do now). The alternative is to instead iterate over the wells separately, inserting each well's source terms then.

atgeirr:
> Please improve the description

I will amend it.

blattms:
Do we really refer to arbitrary diagonals somewhere or only for the current row/column? In that case we could omit setting up an array with all diagonals and simply search once for the diagonal in the loop.

atgeirr:
> Do we really refer to arbitrary diagonals somewhere or only for the current row/column? In that case we could omit setting up an array with all diagonals and simply search once for the diagonal in the loop.

That could save a little memory, however the boundary condition loop does random access. So will a revised version of OPM/opm-simulators#4159 for the wells. A forthcoming change to this assembly routine will also use random access, to handle the (i, j) and (j, i) fluxes simultaneously.  This indicates we should keep the array.

Also, even if there was just a single loop, finding the diagonal element carries a cost that is likely greater than the array lookup. Some variants of the CRS matrix format puts the diagonal element first for this reason, instead of using a sorted order.

In conclusion, I will merge this as is, but remain open to ways to make it more sophisticated.

PR #772, opened 2022-12-15
BUGFIX storage term for timeIdx 1 with DRSDT

totto82:
The bug was introduced with the new tpfa linarizer and effect all cases with DRSDT /DRSTCON/DRVDT 

I make it draft since it does not go well along with the current storageCache feature which seems to only be partly used by the new linearizer. 

Fixes issue reported in https://github.com/OPM/opm-simulators/issues/4318

### DIFF
--- a/opm/models/discretization/common/fvbasediscretization.hh
+++ b/opm/models/discretization/common/fvbasediscretization.hh
@@ -699,10 +699,10 @@ public:
             !intensiveQuantityCacheUpToDate_[timeIdx][globalIdx])
             return 0;
 
-        if (timeIdx > 0 && enableStorageCache_)
-            // with the storage cache enabled, only the intensive quantities for the most
-            // recent time step are cached!
-            return 0;
+        //if (timeIdx > 0 && enableStorageCache_)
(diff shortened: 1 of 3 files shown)


totto82:
I think this issue should be fixed before the release. @atgeirr @hnil

hnil:
I do not have full overview of the features. But I think it probably will be best to enforce cached and update them at end of time step. In principle eclproblem do not allow calculating IQ at anything else than timeIdx=0. I made  separate change reintroducing all but have_stached in the tpfa linearizer which make.

totto82:
I guess you refer to https://github.com/OPM/opm-models/pull/797 So your suggestion is to merge #797 instead. Did you try #797 for the case reported in https://github.com/OPM/opm-simulators/issues/4318?

totto82:
We have a workaround (use the old linearizer) so no. I will remove the tag. I hoped that adding the tag would boost some activity in fixing this. I will remote the tag.

atgeirr:
Closing in favour of #806 and #808.

PR #773, opened 2022-12-19
Fix Todd-Longstaff model for density. 

totto82:
Only modify density (i.e) the effective gravity contribution and not the volumes itself.

### DIFF
--- a/opm/models/blackoil/blackoilsolventmodules.hh
+++ b/opm/models/blackoil/blackoilsolventmodules.hh
@@ -999,36 +999,35 @@ private:
         const Evaluation bg = fs.invB(gasPhaseIdx);
         const Evaluation bs = solventInverseFormationVolumeFactor();
 
-        // Set the effective invB factors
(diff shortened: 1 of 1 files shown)


totto82:
Based on testing against test case 2.1 in "A benchmark study on problems related to CO2 storage
in geologic formations"  https://link.springer.com/article/10.1007/s10596-009-9146-x

totto82:
jenkins build this please

bska:
I haven't looked at this in any kind of detail, but I'm guessing that the base branch is too old. The build failure references version 2022.10 and that suggests to me that you may need to rebase this onto current master.

totto82:
Any review on this? I think the current implementation is wrong. With this fix we reproduce similar results as other simulators in the benchmark study.

totto82:
Since nobody volunteered to review this I will self-merge tomorrow after updating the data. This only affects the solvent model and fixes a bug for the effective density calculations.

totto82:

PR #776, opened 2023-01-04
Add well source terms to the residual separately.

atgeirr:
The existing code adds the source terms cell by cell, and when a perforated cell is reached, all wells and perforations are iterated over to find the correct term to add. So with K total perforations, we do O(K^2) work. with the current PR, the well source terms will be treated separately, iterating over all wells and perforations just once, so O(K) work.

The difference for the assembly performance is modest: ~3% improvement observed (with SPE9, which has a fairly high number of wells for its size). But we see that there are more parts here that can be improved, building on this. We think similarly changing the approach for aquifers may give a similar boost for example.

We add a command line option `--separate-sparse-source-terms` to control the new feature (default is true, activating it). However, actually changing it is not available in the main "flow" executable at this point, since it was added in the TpfaLinearizer class, and the pre-TypeTag used for initial command line parsing uses the default linearizer. For the "flow_blackoil" executable, you can use the option to turn off the feature for comparisons. We may want to use the TpfaLinearizer for the pre-TypeTag, but that would add an option with no effect in the legacy assembly cases. Open to suggestions for the best approach here (or how to get rid of the TypeTag-dependency of the command line options).

There is a downstream branch in opm-simulators that must be combined with this.

### DIFF
--- a/opm/models/blackoil/blackoillocalresidualtpfa.hh
+++ b/opm/models/blackoil/blackoillocalresidualtpfa.hh
@@ -562,6 +562,24 @@ public:
             source[Indices::contiEnergyEqIdx] *= getPropValue<TypeTag, Properties::BlackOilEnergyScalingFactor>();
     }
 
+    static void computeSourceDense(RateVector& source,
+                                   const Problem& problem,
+                                   unsigned globalSpaceIdex,
+                                   unsigned timeIdx)
+    {
+        source = 0.0;
+        problem.addToSourceDense(source, globalSpaceIdex, timeIdx);
+
+        // deal with MICP (if present)
+        // deal with micp (if present)
+        static_assert(!enableMICP, "Relevant addSource() method must be implemented for this module before enabling.");
+        // MICPModule::addSource(source, elemCtx, dofIdx, timeIdx);
+
+        // scale the source term of the energy equation
+        if (enableEnergy)
+            source[Indices::contiEnergyEqIdx] *= getPropValue<TypeTag, Properties::BlackOilEnergyScalingFactor>();
+    }
+
     /*!
      * \copydoc FvBaseLocalResidual::computeSource
      */
--- a/opm/models/discretization/common/fvbasediscretization.hh
+++ b/opm/models/discretization/common/fvbasediscretization.hh
@@ -523,6 +523,7 @@ public:
         IntensiveQuantities::registerParameters();
         ExtensiveQuantities::registerParameters();
         NewtonMethod::registerParameters();
+        Linearizer::registerParameters();
 
         // register runtime parameters of the output modules
         VtkPrimaryVarsModule<TypeTag>::registerParameters();
--- a/opm/models/discretization/common/tpfalinearizer.hh
+++ b/opm/models/discretization/common/tpfalinearizer.hh
@@ -532,11 +532,7 @@ public:
 private:
     void linearize_()
     {
-<<<<<<< HEAD
         OPM_TIMEBLOCK(linearize);
-        const bool well_local = true;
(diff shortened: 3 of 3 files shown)


atgeirr:
jenkins build this opm-simulators=4347 please

atgeirr:
jenkins build this opm-simulators=4347 please

OPMUSER:
@atgeirr Okay I'll bite :-)

Two things make my head hurt, multi-segment wells and OPM Flow's numerical controls. And now the latter is getting more complicated - even more pain for my head.

At the moment we have two linearization routines, the new (TpfaLinearizer) and the original. However, some models only work with the original (solvent, polymer, etc.). I would have thought it would be more sensible to fully implement the TpfaLinearizer for all models first, before adding additional functionality. Then one can use the existing functionality to control the feature. Once tested, we can then remove the original, or perhaps make it a compile option or  a hidden command like option. To be clear, I'm not against adding additional numerical options, just that we should complete the ones we have started first.

As OPM Flow moves towards more being used in a "production environment", it is important to ensure consistency not only in results but also in usage. REs don't really care about the numerical controls unless the run has numerical issues. So imagine an REs anguish when investigating numerical issues having to delve into OPM Flow's numerical controls.

atgeirr:
> At the moment we have two linearization routines, the new (TpfaLinearizer) and the original. However, some models only work with the original (solvent, polymer, etc.). I would have thought it would be more sensible to fully implement the TpfaLinearizer for all models first, before adding additional functionality. Then one can use the existing functionality to control the feature. Once tested, we can then remove the original, or perhaps make it a compile option or a hidden command like option. To be clear, I'm not against adding additional numerical options, just that we should complete the ones we have started first.

This is a very good point. I believe it comes down to two questions:
 - Priority: do we want to spend the effort to move all models to the new linearizer? This is nontrivial for some of the remaining modules, so if we want to be completely on the new linearizer it will take a bit of work.
 - The old/legacy linearization approach is more general, so some features people might want (i.e. new modules with new fluid behaviours) may be simpler to do with that one. It is also possible to use other discretizations than TPFA with it, although we have never used that for OPM Flow. So it has value, even if we were to refactor all the modules currently in Flow. Maybe primarily as a prototyping tool, but I dislike the idea of losing any existing capability!

OPMUSER:
> Priority: do we want to spend the effort to move all models to the new linearizer? This is nontrivial for some of the remaining modules, so if we want to be completely on the new linearizer it will take a bit of work.

I think this is worth the effort, as I think it will save time in the end, but there again I have no skin in the game :-)

> The old/legacy linearization approach is more general, so some features people might want (i.e. new modules with new fluid behaviours) may be simpler to do with that one. It is also possible to use other discretizations than TPFA with it, although we have never used that for OPM Flow. So it has value, even if we were to refactor all the modules currently in Flow. Maybe primarily as a prototyping tool, but I dislike the idea of losing any existing capability! 

Okay, if we wish to keep the old/legacy linearization just place the option as a hidden command line variable, so that the REs don't play with it and break things - they excel at breaking things.

Others should voice an opinion

atgeirr:
jenkins build this opm-simulators=4347 please

atgeirr:
I have now reversed the default for the new parameter, so everything should continue as before by default. The parameter is (as before) only visible from the flow_blackoil executable, which hides it from random users but allows experimentation and building on this going forward.

I think this should be mergeable now.

hnil (approved):
All good to me.

bska:
jenkins build this please

bska:
> All good to me.

Very good.  The build check is green too, so I'll merge into master.

PR #779, opened 2023-02-01
BlackoilPrimaryVariables: add serialization support

akva2:
For restart.

PR #780, opened 2023-02-01
FvBaseDiscretization: add serialization support

akva2:
Sits on top of https://github.com/OPM/opm-models/pull/779

For restart.

### DIFF
--- a/opm/models/discretization/common/fvbasediscretization.hh
+++ b/opm/models/discretization/common/fvbasediscretization.hh
@@ -71,6 +71,7 @@
 #include <dune/fem/misc/capabilities.hh>
 #endif
 
+#include <algorithm>
 #include <limits>
 #include <list>
 #include <stdexcept>
@@ -379,6 +380,7 @@ class FvBaseDiscretization
 
     using LocalEvalBlockVector = typename LocalResidual::LocalEvalBlockVector;
 
+public:
     class BlockVectorWrapper
     {
     protected:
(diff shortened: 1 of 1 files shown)


akva2:
jenkins build this opm-common=3383 opm-simulators=4422 please

bska:
> Sits on top of #779

That PR was merged.  Is this PR still a work in progress?

akva2:
No, just wasn't pertient yet since the need is a bit down the chain in simulators still.

bska:
jenkins build this please

opm/models/discretization/common/fvbasediscretization.hh
```diff
@@ -388,12 +389,44 @@ class FvBaseDiscretization
...
+
+            return result;
+        }
+
         SolutionVector& blockVector()
         { return blockVector_; }
         const SolutionVector& blockVector() const
         { return blockVector_; }
+
+        bool operator==(const BlockVectorWrapper& wrapper)
```
bska, line 409: The equality operator should typically be marked `const`.

opm/models/discretization/common/fvbasediscretization.hh
```diff
@@ -1869,6 +1902,27 @@ public:
...
+        if (this->solution_.size() != rhs.solution_.size())
+            return false;
+
+        for (size_t i = 0; i < this->solution_.size(); ++i) {
+            if (*this->solution_[i] != *rhs.solution_[i]) {
+                return false;
+            }
+        }
+
+        return true;
```
bska, line 1923: I **suppose** you could write this in terms of [`std::equal()`](https://en.cppreference.com/w/cpp/algorithm/equal)
```C++
return std::equal(this->solution_.begin(), this->solution_.end(),
                  rhs.solution_.begin(), rhs.solution_.end(),
    [](const auto& x, const auto& y)
    {
        return *x == *y;
    });
```
Whether or not that's more readable is probably in the eyes of the beholder.

bska (commented):
Looks good for the most part.  One small nit and one suggestion for you to consider.  Otherwise this is ready for merging.

opm/models/discretization/common/fvbasediscretization.hh
akva2: i think it is more readable, one quick glance and you can see what's going on. changed both operators to use it.

bska (approved):
Thanks a lot for the updates.  I appreciate using `std::equal()` in both equality operators.  I too think that's more readable than a raw loop.  I'll merge into master once the build check succeeds.

bska:
> I'll merge into master once the build check succeeds.

Build check is successful.  I'll merge into master.

PR #782, opened 2023-02-09
Simulator: add serialization support

akva2:
Sits on top of #780

### DIFF
--- a/opm/models/utils/simulator.hh
+++ b/opm/models/utils/simulator.hh
@@ -970,6 +970,20 @@ public:
         restarter.deserializeSectionEnd();
     }
 
+    template<class Serializer>
+    void serializeOp(Serializer& serializer)
+    {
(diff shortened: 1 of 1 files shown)


bska (commented):
This is probably okay, but in previous PRs of this kind there has usually been `operator==()` and a `serializeOp()` member functions.  Do we need those here?

akva2:
not really. this is, basically,  the top level of the serialization. a test here is the same as running a full simulator, and that is more the job for a regression test (which are eventually added further down the chain).

bska (commented):
>> Do we need [`operator==()` and `serializeOp()`] here?
>
> not really. [...] a test here is the same as running a full simulator, and that is more the job for a regression test

Understood.  In that case I'll merge into master.

bska:
jenkins build this please

PR #787, opened 2023-02-17
Ensure serialization of the solution object also with dune-fem.

atgeirr:
Not tested to work yet, but at least it compiles.


atgeirr:
jenkins build this please

bska (approved):
I'll merge this to restore the build, but keep the possibility open that we may have to revisit this at a later time.

PR #790, opened 2023-03-01
include TimingMacros.hpp where used

akva2:
Not required to build but was requested in review.
Downstream of https://github.com/OPM/opm-common/pull/3418


bska (approved):

PR #791, opened 2023-03-02
- changes to be able to override methods in blacoilmodel

hnil:
- chaches to update full intensive quantities including extxrution factor without elementctx

### DIFF
--- a/opm/models/discretization/common/fvbaseintensivequantities.hh
+++ b/opm/models/discretization/common/fvbaseintensivequantities.hh
@@ -80,14 +80,15 @@ public:
      */
     void checkDefined() const
     { }
-
+protected:
+    Scalar extrusionFactor_;
 private:
     const Implementation& asImp_() const
     { return *static_cast<const Implementation*>(this); }
     Implementation& asImp_()
     { return *static_cast<Implementation*>(this); }
 
-    Scalar extrusionFactor_;
+    
 };
 
 } // namespace Opm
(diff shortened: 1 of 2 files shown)


bska:
jenkins build this please

opm/models/discretization/common/fvbaseintensivequantities.hh
```diff
@@ -80,14 +80,15 @@ public:
      */
     void checkDefined() const
     { }
-
+protected:
+    Scalar extrusionFactor_;
```
atgeirr, line 84: Why did you need to make this protected? The rest of the changes I believe are good. (But @akva2 should confirm.)

akva2: as this is ebos serialization and not "my" stuff, i don't have much to say as such. but looks like the standard upcasting stuff so it is probably fine.

bska: > Why did you need to make this protected?

The [description](https://github.com/OPM/opm-models/pull/791#issue-1606296298) says that this is needed to update the full set of intensive quantities, **including the extrusion factor**, in a situation that does not have an element context.  For what it's worth, I'm very much in favour of bringing this change in.

bska (approved):
I think this looks good and certainly generalises the current implementation.  I'll merge this unless there's a **really** good reason to avoid the extrusion factor being promoted from `private` to `protected` status.

hnil:
the extrusionFactor_ think is since to be able ti initializ this in a subclass. This is usefull to be able to have a valid object even if one do not use elemtctx

atgeirr:
Initializing from a subclass is a good reason, thanks. Merging!

PR #792, opened 2023-03-09
Fix simulators using quad precision

akva2:
Downstream of https://github.com/OPM/opm-common/pull/3435


blattms (approved):
Approved. Don't understand the code removed, though

akva2:
i dunno, it works fine with the default solvers now so i figured it was better to not do it any longer. looks like long outdated stuff, it works just fine with dune 2.8 atleast.

PR #800, opened 2023-04-17
fix gaswater case with only water

totto82:
The only-water special case only make sense for 3p cases where it is used to avoid singular matrices not of 2p gas-water case.


totto82:
jenkins build this please

totto82:
jenkins build this please

totto82:
jenkins build this please

PR #805, opened 2023-05-12
Extend foam model with solvent or water

totto82:
Depends on https://github.com/OPM/opm-common/pull/3523

### DIFF
--- a/opm/models/blackoil/blackoilfoammodules.hh
+++ b/opm/models/blackoil/blackoilfoammodules.hh
@@ -75,12 +75,15 @@ class BlackOilFoamModule
     static constexpr unsigned foamConcentrationIdx = Indices::foamConcentrationIdx;
     static constexpr unsigned contiFoamEqIdx = Indices::contiFoamEqIdx;
     static constexpr unsigned gasPhaseIdx = FluidSystem::gasPhaseIdx;
+    static constexpr unsigned waterPhaseIdx = FluidSystem::waterPhaseIdx;
 
     static constexpr unsigned enableFoam = enableFoamV;
     static constexpr bool enableVtkOutput = getPropValue<TypeTag, Properties::EnableVtkOutput>();
 
     static constexpr unsigned numEq = getPropValue<TypeTag, Properties::NumEq>();
     static constexpr unsigned numPhases = FluidSystem::numPhases;
+
+    enum { enableSolvent = getPropValue<TypeTag, Properties::EnableSolvent>() };
 
 public:
 #if HAVE_ECL_INPUT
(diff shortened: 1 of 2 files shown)


opm/models/blackoil/blackoilfoammodules.hh
```diff
@@ -475,8 +526,18 @@ public:
...
 
-        // adjust gas mobility
-        asImp_().mobility_[gasPhaseIdx] *= mobilityReductionFactor;
+        // adjust mobility
+        if (FoamModule::transportPhase() == Phase::WATER) {
+            asImp_().mobility_[waterPhaseIdx] *= mobilityReductionFactor;
+        } else if (FoamModule::transportPhase() == Phase::GAS) {
+            asImp_().mobility_[gasPhaseIdx] *= mobilityReductionFactor;
+        } else if (FoamModule::transportPhase() == Phase::SOLVENT) {
+            if constexpr (enableSolvent) {
```
GitPaean, line 535: should we assert `enableSolvent == true` or throw if it is false. 

If we specify the transportPhase is SOLVENT, we should not let code continue running if `enableSolvent` is `false`, right?

opm/models/blackoil/blackoilfoammodules.hh
```diff
@@ -475,8 +526,18 @@ public:
             mobilityReductionFactor = gasMobilityMultiplier.eval(foamConcentration_, /* extrapolate = */ true);
         }
 
-        // adjust gas mobility
-        asImp_().mobility_[gasPhaseIdx] *= mobilityReductionFactor;
+        // adjust mobility
+        if (FoamModule::transportPhase() == Phase::WATER) {
```
GitPaean, line 530: better to use `switch` here.

opm/models/blackoil/blackoilfoammodules.hh
```diff
@@ -280,23 +291,57 @@ public:
...
-            if (upIdx == inIdx) {
-                flux[contiFoamEqIdx] =
-                    extQuants.volumeFlux(gasPhaseIdx)
-                    *up.fluidState().invB(gasPhaseIdx)
-                    *up.foamConcentration();
+
+            // The effect of the mobility reduction factor is
+            // incorporated in the mobility for the relevant phase,
+            // so fluxes do not need modification here.
+            if (params_.transport_phase_ == Phase::WATER) {
```
GitPaean, line 299: better to use `switch` here also, which is more readable.

GitPaean:
The code looks fine to me. Do we have any example for water as the transport phase for foam or is it a future planned combination?

totto82:
Thanks for the review. I have updated the code accordingly. 

> Do we have any example for water as the transport phase for foam or is it a future planned combination?

No, I am mostly interested in the SOLVENT part, but I though it made sense to extend the framework to also include water while working on it.

GitPaean:
jenkins build this please

GitPaean:
Thanks for the update. I am getting this PR in now.

PR #806, opened 2023-05-16
Enable turning off storage cache.

atgeirr:
Replaces #797

With this, there are still some issues, such as seeing slighty different runs with or without storage cache (on Norne, not observed on simpler cases), but they should be handled separately.

The PR #772 should be modified a bit after this, adding the storage cache shifting makes it possible to simplify the code a little.


atgeirr:
jenkins build this opm-simulators=4662 please

atgeirr:
jenkins build this opm-simulators=4662 please

bska:
jenkins build this opm-simulators=4662 opm-tests=975 please

bska:
> update_data please

What part of this patch series requires updating the reference solutions?

bska:
> This fixes a bug for DRSDTCON, reported in [OPM/opm-simulators#4318](https://github.com/OPM/opm-simulators/issues/4318). I am not 100% sure why the other test case is modified, but the behaviour has been checked, and I believe it is fine.

Thanks for the clarifying.  I'm happy with the current version and the new reference solutions have been installed on the CI system.  I'll merge into master.

PR #808, opened 2023-05-31
Storage cache refinements

atgeirr:
This makes the old and new linearizers behave in the same way, and the code is a bit simpler and clearer. Builds on and replaces #772 

Some changes here are directly #772 so I made @totto82 the author, but the commits are new here and not cherry-picked from the other branch.

### DIFF
--- a/opm/models/discretization/common/fvbaselocalresidual.hh
+++ b/opm/models/discretization/common/fvbaselocalresidual.hh
@@ -535,40 +535,30 @@ protected:
             if (elemCtx.enableStorageCache()) {
                 const auto& model = elemCtx.model();
                 unsigned globalDofIdx = elemCtx.globalSpaceIndex(dofIdx, /*timeIdx=*/0);
-                if (model.newtonMethod().numIterations() == 0 &&
-                    !elemCtx.haveStashedIntensiveQuantities())
-                {
-                    if (!elemCtx.problem().recycleFirstIterationStorage()) {
-                        // we re-calculate the storage term for the solution of the
-                        // previous time step from scratch instead of using the one of
-                        // the first iteration of the current time step.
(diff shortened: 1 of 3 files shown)


atgeirr:
jenkins build this please

opm/models/discretization/common/fvbaselocalresidual.hh
```diff
@@ -535,25 +535,26 @@ protected:
...
-                         Valgrind::CheckDefined(tmp2);
-                         model.updateCachedStorage(globalDofIdx, /*timeIdx=*/1, tmp2);
-                     }
-                 } else {
-                     model.updateCachedStorage(globalDofIdx, /*timeIdx=*/0, tmp2);
-                 }
+                for (unsigned eqIdx = 0; eqIdx < numEq; ++eqIdx) {
+                    tmp2[eqIdx] = Toolbox::value(tmp[eqIdx]);
+                }
+                if (!elemCtx.haveStashedIntensiveQuantities()) {
```
bska, line 541: Huh, interesting.  I'd not come across that function before.  Its [documentation](https://github.com/OPM/opm-models/blob/fb2a4030f57a4e19b71a7555776a582abad3cb5f/opm/models/discretization/common/fvbaseelementcontext.hh#L456-L461) is a little hard to comprehend though, since it states that the function

> returns `true` if NO intensive quantities are stashed

Is that accurate?

atgeirr: I have also not really come across it. The mechanism is used to stash away data for a cell before perturbing the element context for finite difference derivative evaluations, and then restoring it afterwards, so its removal from the logic made such cases fail.

The doc is wrong...

atgeirr:
jenkins build this please

PR #811, opened 2023-06-14
ThreadManager: Allow querying number of threads from OpenMP in init.

blattms:
We actually do set the number of OpenMP threads based on OMP_NUM_THREADS or command line option ThreadsPerProcess for OpenMP already in opm-simulators and these settings should just be used here, too. Hence the new flag for skipping the command line parameter and use the number of threads that OpenMP has available.

Querying it here results in different behavior as for FlowMainEbos.hpp: In opm-simulators variable OMP_NUM_THREADS takes precedence over the command line parameter, but here it would only be used if the command line option was not used at all.

Companion PR in opm-simulators is coming.

### DIFF
--- a/opm/models/parallel/threadmanager.hh
+++ b/opm/models/parallel/threadmanager.hh
@@ -64,34 +64,46 @@ public:
                              "('-1' means 'automatic')");
     }
 
-    static void init()
+    /*!
+     * \brief Initialize number of threads used thread manager.
+     *
+     * \param queryCommandLineParameter if set to true we will query ThreadsPerProcess
+     *        and if set (disregard the environment variable OPM_NUM_THREADS).
+     *        If false we will assume that the number of OpenMP threads is already set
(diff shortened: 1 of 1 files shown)


atgeirr:
This and the opm-simulators PR both look good, merging.

opm/models/parallel/threadmanager.hh
```diff
@@ -64,34 +64,46 @@ public:
...
 #ifdef _OPENMP
-        // actually limit the number of threads and get the number of threads which are
-        // used in the end.
-        if (numThreads_ > 0)
-            omp_set_num_threads(numThreads_);
+            // actually limit the number of threads and get the number of threads which are
+            // used in the end.
+            if (numThreads_ > 0)
+                omp_set_num_threads(numThreads_);
+        }
```
atgeirr, line 106: I failed to notice this during review, but the right brace here (inside #ifdef) matched a left brace outside #idfef. I will fix.

PR #803, opened 2023-04-26
Fixes to the special case with almost only water

totto82:
For the almost only water case the primary variables are p, sg and sw.

1) Use sw>=1-1e-6 as threshold for defining almost only water 
2) Add sg to sw if sg<0 to make the oil saturation consistent even though sg is set to zero 
3) Chop the water saturation at 1.01 and not 1.0 in order to improve convergence 
4) Restrict update of rs/rv and rsw/rvw by the saturation scaling factor from the Appleyard-chopping.

PR #816, opened 2023-08-07
add parameter to set maximum water saturation. Default is 1.0

totto82:
This is the second part of

https://github.com/OPM/opm-models/pull/803

which adds a option to set the maximum water saturation. For some cases using a value slightly above 1.0 (for instance 1.05) improves the convergence significantly. Default is 1.0 so this PR should not change default results.


totto82:
jenkins build this please

totto82:
jenkins build this please

totto82:
jenkins build this please

opm/models/blackoil/blackoilprimaryvariables.hh
```diff
@@ -546,16 +544,19 @@ public:
         // keep track if any meaning has changed
         bool changed = false;
 
-        // special case cells with almost only water
-        // use both saturations (if the phase is enabled)
+        // special case for cells with almost only water
+        //thresholdWaterFilledCell = std::min(thresholdWaterFilledCell, 1.0 - eps);
```
atgeirr, line 548: Remove this?

atgeirr (approved):
I think this is acceptable. I would be very nervous about a future PR changing the default, especially for MaxWaterSaturation! However, I acknowledge that this makes it much easier to test various combinations and in turn perhaps find a better solution.

@hnil What do you think?

totto82:
jenkins build this please

totto82:
Thanks for approving. Will merge.

PR #817, opened 2023-08-07
fix almost filled with water to 1.0-1e-6

totto82:
This is the third part of https://github.com/OPM/opm-models/pull/803

### DIFF
--- a/opm/models/blackoil/blackoilprimaryvariables.hh
+++ b/opm/models/blackoil/blackoilprimaryvariables.hh
@@ -495,8 +495,6 @@ public:
      */
     bool adaptPrimaryVariables(const Problem& problem, unsigned globalDofIdx, Scalar eps = 0.0)
     {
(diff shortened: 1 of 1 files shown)


totto82:
jenkins build this please

hnil:
Probably put it in the tag system to be changed. Maybe we should add a type DefaultTolerances where one could get forexample (if it do not exist):
using NumTol = GetPropertyType<TypeTag, DefaultTolerances>;
static constexpr myeps = NumTol::eps;

Reasons: 
- be able to change them
- avoid putting it as hidden defaut input
- collect numbers used for numerical hacks ++ one place
- avoid all the boiler plate assosiated with input parameter i.e. tag definitions, init, register.

totto82:
benchmark please

totto82:
> Probably put it in the tag system to be changed.

Good point. Especially since it seem to affect lot of cases.

totto82:
benchmark please

PR #818, opened 2023-08-08
-- added missing include

hnil:
-- add definition of BCType


bska:
jenkins build this please

akva2:
this needs to be conditional on HAVE_ECL_INPUT (i seriously dislike those defines, but...)

atgeirr:
> this needs to be conditional on HAVE_ECL_INPUT (i seriously dislike those defines, but...)

I believe the whole BC treatment depends on this, so that means TpfaLinearizer also must be guarded by HAVE_ECL_INPUT?

akva2:
yes, the whole thing. we deal with it separately.

PR #820, opened 2023-08-08
Try to detect more exceptions during parallel simulator setup.

blattms:
We have experienced some deadlocks lately because of programming errors. This change will not result in a clean exit for those but rather an abort by MPI because of non-matching communications.

Also we cannot forsee whether people will add more exceptions in those parts that were not protected. We better play it safe here.

### DIFF
--- a/opm/models/utils/simulator.hh
+++ b/opm/models/utils/simulator.hh
@@ -154,23 +154,38 @@ public:
 
         int exceptionThrown = 0;
         std::string what;
-        try
-        { vanguard_.reset(new Vanguard(*this)); }
-        catch (const std::exception& e) {
+
+        auto catchAction =
+            [&exceptionThrown, &what, comm](const std::exception& e,
+                                            bool doPrint) {
             exceptionThrown = 1;
             what = e.what();
             if (comm.size() > 1) {
                 what += " (on rank " + std::to_string(comm.rank()) + ")";
             }
-            if (verbose_)
+            if (doPrint)
                 std::cerr << "Rank " << comm.rank() << " threw an exception: " << e.what() << std::endl;
-        }
+        };
 
-        if (comm.max(exceptionThrown)) {
-            auto all_what = gatherStrings(what);
-            assert(!all_what.empty());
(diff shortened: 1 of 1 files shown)


blattms:

akva2:
++warnings

atgeirr:
> ++warnings

Could we somehow get a yellow button or other hint if there are new warnings? I find I often forget to check this.

atgeirr:
What about filtering out warnings that do not come from a file with "opm" in the path somewhere?

akva2:
jenkins build this please

akva2:
jenkins build this please

akva2:
there u go.

blattms:
Warnings should be gone now.

opm/models/utils/simulator.hh
```diff
@@ -154,94 +154,86 @@ public:
...
                 what += " (on rank " + std::to_string(comm.rank()) + ")";
             }
-            if (verbose_)
+            if (doPrint)
                 std::cerr << "Rank " << comm.rank() << " threw an exception: " << e.what() << std::endl;
-        }
+        };
+
+        auto checkParallelException =
+            [exceptionThrown, what, comm](const std::string& prefix)
```
bska, line 171: If you're taking `what` and `exceptionThrown` by value here you'll get a copy of whatever value was there at the creation of `checkParallelException`.  Is that what you intended?  If not, I think you need to take by reference instead.

bska (commented):
Thanks a lot for the cleanup.  The code looks a lot more regular now.  I only notice one potential issue in the `checkParallelException()` lambda, so maybe you can comment on the reasons for writing it this way?

opm/models/utils/simulator.hh
blattms: Very good catch. That was close. :sweat: :flushed: 

I wanted to prevent the ability to change those guys in the lambda. Of course I overlooked that their clones are created here and not when the lambda is called. Hence it would never have seen the updated values below. Those guys have become parameters now.

blattms:
jenkins build this please

bska (approved):
Thanks a lot for the updates.  This looks good to me now and I'll merge into master.

PR #815, opened 2023-08-04
Dunefem

hnil:
-- changes to make HAVE_FEM only in tagstem.
-- This make it possible to set tags so that one can run without dune fem stuff or without in same code.
Usage:
- make standard opm-flow same always, but be able have executables with possibilility of refeinement or other use of dune-fem.

PR #821, opened 2023-08-10
move the dynamic refinement part related to dune fem to separate class

akva2:
introduced base discretization in tagsystem to be able to change between dunefem adaptation or not

This is a rebased, fixed and improved version of #815

### DIFF
--- a/opm/models/discretization/common/fvbasediscretizationfemadapt.hh
+++ b/opm/models/discretization/common/fvbasediscretizationfemadapt.hh
@@ -1,0 +1,174 @@
+// -*- mode: C++; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*-
+// vi: set et ts=4 sw=4 sts=4:
+/*
+  This file is part of the Open Porous Media project (OPM).
+
+  OPM is free software: you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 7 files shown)


akva2:
jenkins build this opm-simulators=4789 please

opm/models/discretization/common/fvbasediscretizationfemadapt.hh
```diff
@@ -0,0 +1,174 @@
...
+ * \file
+ *
+ * \copydoc Opm::FvBaseDiscretization
+ */
+#ifndef EWOMS_FV_BASE_DISCRETIZATION_FEMADAPT_HH
+#define EWOMS_FV_BASE_DISCRETIZATION_FEMADAPT_HH
+
+#include <opm/models/discretization/common/fvbasediscretization.hh>
+
+#include <dune/fem/space/common/adaptationmanager.hh>
```
atgeirr, line 33: Since we include dune-fem parts here: should we add some guarding to ensure HAVE_DUNE_FEM is set?

atgeirr (approved):
I have one minor issue but I am fine with merging as is.

opm/models/discretization/common/fvbasediscretizationfemadapt.hh
akva2: this whole file requires dune-fem so guard has to be further up. guarding here is pointless.

PR #823, opened 2023-08-25
Avoid collective operations in single-domain calls.

atgeirr:
This fix is necessary for using the NLDD method in MPI parallel runs. There are also fixes coming in opm-simulators but they are independent of this so it can be considered separately.

We must avoid collective operations (here: doing comm.sum() to detect exceptions on any process) in single-domain contexts, as those will not be synchronous across processes (similar to the situation for single-well contexts). Moving the logic to a different function preserves existing behaviour for regular Newton.

### DIFF
--- a/opm/models/blackoil/blackoilnewtonmethod.hh
+++ b/opm/models/blackoil/blackoilnewtonmethod.hh
@@ -239,33 +239,17 @@ public:
                  const GlobalEqVector& currentResidual,
                  const DofIndices& dofIndices)
     {
-        const auto& comm = this->simulator_.gridView().comm();
-
-        int succeeded;
-        try {
-            auto zero = solutionUpdate[0];
-            zero = 0.0;
-            for (auto dofIdx : dofIndices) {
-                if (solutionUpdate[dofIdx] == zero) {
-                    continue;
-                }
-                updatePrimaryVariables_(dofIdx,
-                                        nextSolution[dofIdx],
-                                        currentSolution[dofIdx],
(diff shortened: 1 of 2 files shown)


atgeirr:
jenkins build this please

opm/models/blackoil/blackoilnewtonmethod.hh
```diff
@@ -239,33 +239,18 @@ public:
...
-                if (solutionUpdate[dofIdx] == zero) {
-                    continue;
-                }
-                updatePrimaryVariables_(dofIdx,
-                                        nextSolution[dofIdx],
-                                        currentSolution[dofIdx],
-                                        solutionUpdate[dofIdx],
-                                        currentResidual[dofIdx]);
+        auto zero = solutionUpdate[0];
+        zero = 0.0;
```
bska, line 243: I realise you're "just" moving some code here, but that's weird.  Could you write this as
```C++
const auto zero = 0 * solutionUpdate[0];
```
if "all" you need is an appropriately typed zero value?

atgeirr: Amended.

atgeirr:
jenkins build this please

bska (approved):
Thanks a lot for the updates.  I think this makes sense and the PR looks good to me.  I'll merge into master.  The actual change will, as demonstrated by the build check, only affect those who run NLDD in parallel and that's a very small number of people at the moment...

PR #822, opened 2023-08-24
TPFA version for energy

hnil:
-- added need functions for beeing able to use TPFA linearizer for energy simulation

PR #824, opened 2023-08-25
Diffusion temp

hnil:
add utils for tpfa for diffusion. This builds on the energy tpfa pullrequest (#822).


GitPaean:
jenkins build this opm-simulators=4816 please

GitPaean:
With the default compilation options, with 4 processes running for the SPE11 c case with 47900 active cells, the performance summary is as follows, (left is this PR and right is the master branch)

image

PR #828, opened 2023-09-18
Use temperature from depth in primaryvariable switching

totto82:
This bug could lead to oscillation between primary variables for non constant temperature cases.


totto82:
jenkins build this please

totto82:
The results looks qualitatively the same as expected. The bug introduced some oscillations in the model that seems to be improved with the fix. This also slightly change the time stepping. i.e we should make an update of data and get this fix in.   

image

atgeirr (approved):
This is good, should be merged.

atgeirr:

totto82:
jenkins build this opm-tests=1051 please

bska:
The new reference solutions have been installed on the CI system.  Merging this now to bring the overall system back to a consistent state.

PR #829, opened 2023-09-21
- changed assert to throw

hnil:
-  changed asserts to throw
-  added checks on positivity of physical quantities 
-Saturation is not checked

### DIFF
--- a/opm/models/blackoil/blackoilintensivequantities.hh
+++ b/opm/models/blackoil/blackoilintensivequantities.hh
@@ -39,6 +39,7 @@
 #include "blackoilmicpmodules.hh"
 
 #include <opm/common/TimingMacros.hpp>
+#include <opm/common/Exceptions.hpp>
 #include <opm/common/OpmLog/OpmLog.hpp>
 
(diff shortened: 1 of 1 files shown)


opm/models/blackoil/blackoilintensivequantities.hh
```diff
@@ -472,22 +473,50 @@ public:
 
         // update the diffusion specific quantities of the intensive quantities
         DiffusionIntensiveQuantities::update_(fluidState_, paramCache, elemCtx, dofIdx, timeIdx);
-
-#ifndef NDEBUG
-        // some safety checks in debug mode
+#ifndef NOT_CHECK_STATE
```
blattms, line 476: Why the new name (that nobody knows)?

hnil: Becase NDEBUG is on in production code.

GitPaean: > Becase NDEBUG is on in production code.

I think you meant NDEBUG is **off** in production code with default `-DWITH_NDEBUG=0`
```
-O3 -DNDEBUG -mtune=native -UNDEBUG
```

PR #830, opened 2023-09-22
CLANG: Temporal fix to BC

daavid00:
After updating Xcode (15.0), when running a case with BC the simulation stops by:

``` 
Starting time step 0, stepsize 1 days, at day 0/73, date = 01-Jan-2025
Restart file written for report step 0/1, date = 01-Jan-2025 00:00:00
zsh: trace trap  /Users/dmar/Github/opm/build/opm-simulators/bin/flow_gasoil BC_FREE_TEST.DATA
```
It seems the insideIntQuants is modified after the function calculateBoundaryGradients_. This commit fixes that. I will look for a proper way to fix this, but if someone knows the solution, it will be very appreciated. 

[BC_FREE_TEST.DATA.zip](https://github.com/OPM/opm-models/files/12702188/BC_FREE_TEST.DATA.zip)

### DIFF
--- a/opm/models/blackoil/blackoillocalresidualtpfa.hh
+++ b/opm/models/blackoil/blackoillocalresidualtpfa.hh
@@ -487,7 +487,7 @@ public:
     static void computeBoundaryFluxFree(const Problem& problem,
                                         RateVector& bdyFlux,
                                         const BoundaryConditionData& bdyInfo,
-                                        const IntensiveQuantities& insideIntQuants,
+                                        const IntensiveQuantities& insideIntQuantsHack,
                                         unsigned globalSpaceIdx)
     {
         OPM_TIMEBLOCK_LOCAL(computeBoundaryFluxFree);
@@ -495,6 +495,7 @@ public:
         std::array<short, numPhases> dnIdx;
         RateVector volumeFlux;
         RateVector pressureDifference;
+        const auto insideIntQuants = insideIntQuantsHack;
         ExtensiveQuantities::calculateBoundaryGradients_(problem,
                                                          globalSpaceIdx,
                                                          insideIntQuants,


hnil:
Not sure but the outside code still use model().cachedIntenisveQuantities(... in stead of model().intensiveQuantities...

atgeirr:
Also on mac, with clang. The provided case runs through with no problems, both release and debug mode compiles. I cannot reproduce this.

Version of clang is reported as:
```
Apple clang version 14.0.3 (clang-1403.0.22.14.1)
Target: x86_64-apple-darwin22.5.0
```
I am currently on macOS 13.4.1 (c).

daavid00:
Thanks @hnil and @atgeirr for the reply. It seems then it is a problem with the new version of clang (sorry for not specifying this before).

```
Apple clang version 15.0.0 (clang-1500.0.40.1)
Target: x86_64-apple-darwin22.6.0
```

atgeirr:
> it is a problem with the new version of clang

While this is perhaps likely, it could also be that we have a bug that leads to overwritten memory, manifesting only with that compiler (by random chance). When you say...

> insideIntQuants is modified after the function calculateBoundaryGradients_

...I assume you mean that before the call to calculateBoundaryGradients_ it is ok, but not after the call?

I looked at that function and could not find any obvious bug that could lead to writing out-of-bounds. The proposed change in this PR should not be merged, if you have no other workaround you may want to install a slightly older clang. That would also give us a clearer indication if the compiler is indeed the critical variable here.

daavid00:
Yes, I looked at it last week with @totto82 in debug mode and before the calculateBoundaryGradients_ it is ok, and after the call the insideIntQuants is cleared. I totally agree this should not be merged now, but at least it is a walk around for other mac users that have update Xcode to 15.0 (some colleagues are using it now).

PR #833, opened 2023-10-06
fix issue when BCPROP is not set initially

totto82:
In the current code BCPROP is ignored if it is not set in the first time step. This PR make it possible to set BCPROP later in the schedule.

### DIFF
--- a/opm/models/discretization/common/tpfalinearizer.hh
+++ b/opm/models/discretization/common/tpfalinearizer.hh
@@ -466,16 +466,19 @@ private:
                     }
                 }
                 neighborInfo_.appendRow(loc_nbinfo.begin(), loc_nbinfo.end());
-                for (unsigned bfIndex = 0; bfIndex < stencil.numBoundaryFaces(); ++bfIndex) {
-                    const auto& bf = stencil.boundaryFace(bfIndex);
-                    const int dir_id = bf.dirId();
-                    const auto [type, massrateAD] = problem_().boundaryCondition(myIdx, dir_id);
-                    // Strip the unnecessary (and zero anyway) derivatives off massrate.
-                    VectorBlock massrate(0.0);
-                    for (size_t ii = 0; ii < massrate.size(); ++ii) {
-                        massrate[ii] = massrateAD[ii].value();
-                    }
-                    if (type != BCType::NONE) {
+                if (problem_().nonTrivialBoundaryConditions()) {
+                    for (unsigned bfIndex = 0; bfIndex < stencil.numBoundaryFaces(); ++bfIndex) {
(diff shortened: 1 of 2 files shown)


totto82:
jenkins build this opm-simulators=4912 please

totto82:
jenkins build this opm-simulators=4912 please

totto82:
jenkins build this opm-simulators=4912 please

totto82:
The test failure is expected. The current code does not impose the correct bc for the test case. This fixes a bug, but since there is simple workaround (just adding the BCPROP for the initial step), I dont think it needs to be part of the release. Hopefully it can be merged when the release is done.

totto82:
jenkins build this opm-simulators=4912 please

totto82:
jenkins build this opm-simulators=4912 please

opm/models/discretization/common/tpfalinearizer.hh
```diff
@@ -469,23 +469,24 @@ private:
...
-                                                     exFluidState.pvtRegionIndex(),
-                                                     bfIndex,
-                                                     bf.area(),
-                                                     bf.integrationPos()[dimWorld - 1],
-                                                     exFluidState};
-                        boundaryInfo_.push_back({myIdx, dir_id, bcdata});
-                    }
+                    const auto& exFluidState = problem_().boundaryFluidState(myIdx, dir_id);
+                    BoundaryConditionData bcdata{type,
+                                                    massrate,
```
atgeirr, line 483: Indent.

atgeirr (changes requested):
While I think this is correct, it will likely cause some performance degradation. With this we will always, in every assemble, iterate over all boundary faces even if there are no boundary conditions.

We should make sure that we only build these structures for the boundary faces that will at some point receive a condition, unless you think that is not possible. Maybe we should distinguish between "no condition now" and "no condition ever" in the type enum?

I will review the opm-simulators PR now with a view to this.

totto82:
Thanks for the review. I added a if(nonTrivialBoundaryConditions). That should take care of the performance overhead if you dont have boundary conditions.

totto82:
jenkins build this opm-simulators=4912 please

atgeirr (approved):

totto82:
jenkins build this opm-simulators=4912 update_data please

totto82:
jenkins build this opm-simulators=4912 opm-tests=1089 please

totto82:
jenkins build this opm-simulators=4912 please

totto82:
jenkins build this opm-simulators=4912 update_data please

totto82:
jenkins build this opm-simulators=4912 opm-tests=1089 please

PR #834, opened 2023-10-09
removing flashproperties.hh from multiple files

GitPaean:
it does not looks like it is used.

At least, [flashboundaryratevector.hh](https://github.com/OPM/opm-models/compare/master...GitPaean:removing_flashproperties?expand=1#diff-46406dee878bf82a0f5e5ec542173608fc1c30fc7cd6a25c7e486b6c645601fd)  can reused by other flash solvers directly after removing this header files.  Others might need some adjustments before it can reused directly. 

There might be some design here, but they can be added back anytime when needed.  At least, with its current form, they are NOT used in anyway.

### DIFF
--- a/opm/models/flash/flashboundaryratevector.hh
+++ b/opm/models/flash/flashboundaryratevector.hh
@@ -28,8 +28,6 @@
 #ifndef EWOMS_FLASH_BOUNDARY_RATE_VECTOR_HH
 #define EWOMS_FLASH_BOUNDARY_RATE_VECTOR_HH
 
-#include "flashproperties.hh"
-
 #include <opm/models/common/energymodule.hh>
 #include <opm/material/common/Valgrind.hpp>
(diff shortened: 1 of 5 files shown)


GitPaean:
jenkins build this please

bska:
> There might be some design here, but they can be added back anytime when needed. At least, with its current form, they are used in anyway.

I wouldn't worry too much about it.  No-one has touched this code in anger since commit ec918e30b.  Whatever the initial intention might have been, the code has evolved in a different direction since then.

GitPaean:
> I wouldn't worry too much about it. No-one has touched this code in anger since commit https://github.com/OPM/opm-models/commit/ec918e30b3d1943a8d5e794396fd1f8607c2bd79.

Thanks for the comment. I will self-merge since it is relatively trivial.

PR #836, opened 2023-10-09
Fix to write VTKS when THERMAL is on deck

daavid00:
If the flag `--enable-vtk-output=true` is added to cases with `THERMAL` on the RUNSPEC section (e.g., [CO2STORE_ENERGY.DATA](https://github.com/OPM/opm-tests/blob/master/co2store/CO2STORE_ENERGY.DATA)), then the simulation throws 
```
Report step  0/2 at day 0/18250, date = 01-Jan-2025
Writing visualization results for the current time step.
Assertion failed: (phaseIsActive(phaseIdx)), function canonicalToActivePhaseIdx, file BlackOilFluidSystem.cpp, line 372.
```
This commit fixes this.


blattms:
jenkins build this please

PR #839, opened 2023-10-12
Included missing header causing compile errors with gcc-13

blattms:
compilation error is about missing type Opm::Phase


blattms:
jenkins build this please

PR #840, opened 2023-10-12
Depend on opm-common as a DUNE module

blattms:
We do use a lot of headers from it and opm-grid (which also pulls it in) is optional.


blattms:
jenkins build this please

PR #841, opened 2023-10-16
Fix BC Free for 2p problems

totto82:
This fixes issue when using BCPROP with type FREE and DIRICHLET in a 2p run.

### DIFF
--- a/opm/models/blackoil/blackoillocalresidualtpfa.hh
+++ b/opm/models/blackoil/blackoillocalresidualtpfa.hh
@@ -493,8 +493,9 @@ public:
         OPM_TIMEBLOCK_LOCAL(computeBoundaryFluxFree);
         std::array<short, numPhases> upIdx;
         std::array<short, numPhases> dnIdx;
-        RateVector volumeFlux;
-        RateVector pressureDifference;
+        std::array<Evaluation, numPhases> volumeFlux;
+        std::array<Evaluation, numPhases> pressureDifference;
+
         ExtensiveQuantities::calculateBoundaryGradients_(problem,
                                                          globalSpaceIdx,
                                                          insideIntQuants,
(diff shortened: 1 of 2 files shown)


opm/models/blackoil/blackoillocalresidualtpfa.hh
```diff
@@ -493,8 +493,9 @@ public:
         OPM_TIMEBLOCK_LOCAL(computeBoundaryFluxFree);
         std::array<short, numPhases> upIdx;
         std::array<short, numPhases> dnIdx;
-        RateVector volumeFlux;
-        RateVector pressureDifference;
+        Evaluation volumeFlux[numPhases];
+        Evaluation pressureDifference[numPhases];
```
bska, line 497: Why built-in arrays?  I can't think of a reason we would prefer this to `std::array<>`.

bska (commented):
I want both @atgeirr and @hnil to have a look at the changes to the TPFA lineariser code.  The only thing that stands out to me though, is using built-in arrays over `std::array<>` in the residual impliementation.

opm/models/blackoil/blackoillocalresidualtpfa.hh
totto82: I just copied what was used in eclfluxmodule.hh. I can change to `std:.array<Evaluation, numPhases>`

opm/models/blackoil/blackoillocalresidualtpfa.hh
```diff
@@ -574,8 +575,8 @@ public:
                                                                              inIdx,
                                                                              alpha,
                                                                              bdyInfo.exFluidState);
+            EnergyModule::addHeatFlux(bdyFlux, heatFlux);
```
atgeirr, line 578: I think this change is correct, but then the heatFlux variable should also be created inside the `if constexpr(enableEnergy) {}` block?

atgeirr (approved):
I agree with the fix, just a minor adjustment asked for.

opm/models/blackoil/blackoillocalresidualtpfa.hh
totto82: Good point. Will fix.

totto82:
jenkins build this please

atgeirr:
Thanks, merging!

PR #838, opened 2023-10-12
[WIP] a simulation test based on PTFlash

GitPaean:
various refactoring might be needed before it can go in. 

creating the PR now to help the reviewing of the upstream PR. 

The commits will be squashed in a later stage.

PR #844, opened 2023-10-18
removing compositional related from blackoilmodel.hh

GitPaean:
it is working in progress.

It aims at removing compostional related routines from blackoil related modules. There is a VtkBlackoil already.

### DIFF
--- a/opm/models/blackoil/blackoilmodel.hh
+++ b/opm/models/blackoil/blackoilmodel.hh
@@ -50,7 +50,6 @@
 #include "blackoilmicpmodules.hh"
 
 #include <opm/models/common/multiphasebasemodel.hh>
-#include <opm/models/io/vtkcompositionmodule.hh>
 #include <opm/models/io/vtkblackoilmodule.hh>
 #include "blackoildiffusionmodule.hh"
 #include <opm/models/io/vtkdiffusionmodule.hh>
@@ -72,8 +71,7 @@ namespace Opm::Properties {
 
 namespace TTag {
 //! The type tag for the black-oil problems
-struct BlackOilModel { using InheritsFrom = std::tuple<VtkComposition,
-                                                       VtkBlackOilEnergy,
+struct BlackOilModel { using InheritsFrom = std::tuple<VtkBlackOilEnergy,
                                                        VtkDiffusion,
                                                        VtkBlackOilPolymer,
                                                        VtkBlackOilSolvent,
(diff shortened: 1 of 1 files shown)


GitPaean:
jenkins build this please

GitPaean:
jenkins build this please opm-simulators=4936 please

totto82:
You can compute many of the output you typically have in a compositional simulation from the blackoil model. Some of this conversion is implemented in the BlackoilFluidSystem, some just outputs trivial numbers. It may be useful to look at molar fractions etc even in a blackoil simulations if you want to compare it with compositional simulations. Why do you want to remove it?

GitPaean:
With the current form, when developing https://github.com/OPM/opm-models/pull/838, when we added more output to the VtKCompostional, since VtkCompostional is included in the BlackoilModel, then we have to update the BlackoilFluidState when we add more output to the VtkCompostional.  It might be a bad idea to keep filling the BlackoilFluidState with compostional related stuff, which causes confusion and maintenance burden. 

Then I think it might be a good idea to make sure the VtkCompostional and VtkBlackoil be more separated.  If not many compositional properties from Blackoil are needed to compare against the compositional results, maybe we can just put these few functions also in VtkBlackoil for now, and not include VtkCompostional in the BlackoilModel.  Even for outputting purpose, those are  still  serving and based on blackoil simulations. It is okay to have them in the VtkBlackoil. It is also introduce some maintenance burden, but hopefully more manageable. At least, not when working on compositional simulation and have to take care of the blackoil related.  

How do you think?

GitPaean:
Thanks.  In the future, with some refactoring, we can do refactoring to reduce possible duplication.

GitPaean:
@totto82 , are you okay with simply removing the including of the vtkcompostional from the blackoilmodel, or you prefer we can add the missing function to be vtkblackoil before removing?

I believe in the future, we can have another vtkmodule to handle the common one between compostional and blackoil models.

GitPaean:
If we do not think `vtkcompositionmodule.hh` is for compsitional modeling. We think it is more composition information.  Maybe we should keep vtkcmositionmodule while have a vtkcompostionalmodule.hh (or different name) to handle the compositional simulation related.

totto82:
> are you okay with simply removing the including of the vtkcompostional from the blackoilmodel

Yes. 

> we should keep vtkcmositionmodule while have a vtkcompostionalmodule.hh (or different name) to handle the compositional simulation related.

I am find with this solution as well. 

I suggest you do what you think is easiest in terms of getting something to work for the compositional runs. As I said the usage of compositional output from the blackoil model is probably not used by anybody. And if somebody needs it in the future it should be trivial to add i back.

GitPaean:
Thanks for the discussion. It is in the files that I am not familiar with and some discussion is very helpful. 

At the end, I decided to introduce a new module vtkptflashmodule to avoid touching vtkcompostionmodule. 

In the future, there might be more refactoring. 

I will close this PR now.

PR #848, opened 2023-11-07
[fix] Skip manual error checking and rely on the error default handler aborting

blattms:
According to MPI standard the ERROR field of MPI_Status might not be initialized unless for operations that return multiple statuses, see  Section 3.7.5 of the standard. In older OpenMPI versions (<=4.0.x) we were lucky that ERROR was initialized to 0 always. This is not the case for 4.1.y at least. See open-mpi/ompi#12049.

Therefore we skip the additional manual that has no effect with the default error handler and rely on the error handler aborting the program.

Fixed running the tests on Debian bookworm

### DIFF
--- a/opm/models/parallel/mpibuffer.hh
+++ b/opm/models/parallel/mpibuffer.hh
@@ -111,14 +111,19 @@ public:
     void receive([[maybe_unused]] unsigned peerRank)
     {
 #if HAVE_MPI
-        MPI_Recv(data_,
-                 static_cast<int>(mpiDataSize_),
-                 mpiDataType_,
-                 static_cast<int>(peerRank),
-                 0, // tag
-                 MPI_COMM_WORLD,
-                 &mpiStatus_);
-        assert(!mpiStatus_.MPI_ERROR);
+        // Use return code for error detection
+        // According to MPI standard the ERROR field
+        // might not be initialized unless for operations
+        // that return multiple statuses, see  Section 3.7.5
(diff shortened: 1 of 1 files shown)


blattms:
jenkins build this please

opm/models/parallel/mpibuffer.hh
```diff
@@ -111,14 +111,19 @@ public:
...
+        // that return multiple statuses, see  Section 3.7.5
+        // of the standard
+        [[maybe_unused]] auto result = MPI_Recv(data_,
+                                                static_cast<int>(mpiDataSize_),
+                                                mpiDataType_,
+                                                static_cast<int>(peerRank),
+                                                0, // tag
+                                                MPI_COMM_WORLD,
+                                                MPI_STATUS_IGNORE);
+        assert(!result);
```
bska, line 126: I realise you're "just" changing the source of the condition, but is there a way we could avoid using `assert()` for this purpose?

blattms: Strip it. It has no effect unless some one changes the MPI error handler. Checking the in return values is actually discouraged as it is rather inefficient.

blattms: assert is gone.

blattms:
jenkins build this please

bska (approved):
Thanks a lot for the updates.  This is a much improved version in my opinion and I'll merge into master.

PR #849, opened 2023-11-07
Renaming enableEvaporation to enableVapwat

daavid00:


totto82:
jenkins build this opm-simulators=4973 please

bska:
> jenkins build this opm-simulators=4973 please

That will probably fail.  I think you need OPM/opm-common#3750 too.

PR #858, opened 2024-01-04
use arithmetic averaged compaction transmissibility multipliers

totto82:


totto82:
jenkins build this opm-simulators=5088 please

opm/models/blackoil/blackoillocalresidualtpfa.hh
```diff
@@ -356,7 +356,8 @@ public:
 
             const IntensiveQuantities& up = (upIdx == interiorDofIdx) ? intQuantsIn : intQuantsEx;
             unsigned globalUpIndex = (upIdx == interiorDofIdx) ? globalIndexIn : globalIndexEx;
-            const Evaluation& transMult = up.rockCompTransMultiplier();
+            // Use arithmetic average (more accurate with harmonic, but that requires recomputing the transmissbility)
+            const Evaluation transMult = (intQuantsIn.rockCompTransMultiplier() + Toolbox::value(intQuantsEx.rockCompTransMultiplier()))/2;
```
atgeirr, line 360: I think this is correct. Looking at it, I think that it actually fixes a bug, too: `up.rockCompTransMultiplier()` will have derivatives no matter what, and that is wrong if `up` is `intQuantsEx`! This code is a bit fragile: we should ensure `intQuantsEx` do not have derivatives by passing a derivative-less object. I have an idea for how to make it efficient.

atgeirr (approved):

opm/models/blackoil/blackoillocalresidualtpfa.hh
totto82: There is a check below that assures that derivatives are only kept if in == up. So I dont think the original code is wrong in terms of the derivative.

atgeirr: You are right. However if we made sure that intQuantsEx did not have derivatives the else branch at line ~368 would not be necessary at all, so I think we should consider that for simplification.

totto82: Ok. I see your point and will update.

totto82:
jenkins build this opm-simulators=5088 please

totto82:
jenkins build this opm-simulators=5088 update_data please

totto82:
jenkins build this opm-simulators=5088 opm-tests=1118 please

PR #859, opened 2024-01-11
ignore dispersion in gas phase

totto82:
Dispersion in gas phase leads to convergence issues for the SPE11 caseb and casec (and is not tested). Temporary remove it and add a warning.


totto82:
jenkins build this opm-simulators=5101 please

totto82:
jenkins build this opm-simulators=5101 please

PR #860, opened 2024-01-11
Support for dynamic scaling of capillary pressure due to salt precipitation

daavid00:
With https://github.com/OPM/opm-common/pull/3857, this PR adds support for dynamically scaling the capillary pressure as a result of porosity changes in salt precipitation. This is achieved by adding a multiplier using tables, as it is done for the permeability reduction (while the pcfact tables are set for each satnum region, the current implementation of permfact is defined via the pvtnum regions; this PR could changed that to satnum as well if it is of interest). An example of this functionality can be found in https://github.com/OPM/opm-tests/pull/1117. The Figure on the left corresponds to the resulting PC surface for that example and on the right the comparison for the salt precipitation in the bottom cell with and without this feature. 
SALTPREC_PCFACT

### DIFF
--- a/opm/models/blackoil/blackoilintensivequantities.hh
+++ b/opm/models/blackoil/blackoilintensivequantities.hh
@@ -130,6 +130,7 @@ class BlackOilIntensiveQuantities
     using DispersionIntensiveQuantities = BlackOilDispersionIntensiveQuantities<TypeTag, enableDispersion>;
 
     using DirectionalMobilityPtr = Opm::Utility::CopyablePtr<DirectionalMobility<TypeTag, Evaluation>>;
+    using BrineModule = BlackOilBrineModule<TypeTag>;
 
 
 public:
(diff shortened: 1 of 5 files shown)


opm/models/blackoil/blackoilintensivequantities.hh
```diff
@@ -257,23 +258,34 @@ public:
         MaterialLaw::capillaryPressures(pC, materialParams, fluidState_);
         problem.updateRelperms(mobility_, dirMob_, fluidState_, globalSpaceIdx);
 
+        // dynamic capillary pressure 
+        if (BrineModule::hasPcfactTables() && priVars.primaryVarsMeaningBrine() == PrimaryVariables::BrineMeaning::Sp) {
+            const Evaluation Sp = priVars.makeEvaluation(Indices::saltConcentrationIdx, timeIdx);
+            const Evaluation porosityFactor  = min(1.0 - Sp, 1.0); //phi/phi_0
+            const auto& pcfactTable = BrineModule::pcfactTable(pvtRegionIndex());
+            cpFactor_ = pcfactTable.eval(porosityFactor, /*extrapolation=*/true);
+        }
```
totto82, line 267: Do you need to store the cpFactor? I suggest scaling the pC directly here by applying cpFactor to pC.

daavid00: Done, thanks :).

totto82:
Thanks for the contribution. The PR is now ready for merging. Please consider adding the test case for regression.

PR #862, opened 2024-01-12
added max and minpressure

hnil:
-- add max and min pressure.
This can ensure that one limit evaluations outside tables.
Probably one should change defaults on min temperature to 0 C.

### DIFF
--- a/opm/models/blackoil/blackoilnewtonmethod.hh
+++ b/opm/models/blackoil/blackoilnewtonmethod.hh
@@ -56,6 +56,10 @@ struct TemperatureMax { using type = UndefinedProperty; };
 template<class TypeTag, class MyTypeTag>
 struct TemperatureMin { using type = UndefinedProperty; };
 template<class TypeTag, class MyTypeTag>
+struct PressureMax { using type = UndefinedProperty; };
+template<class TypeTag, class MyTypeTag>
+struct PressureMin { using type = UndefinedProperty; };
+template<class TypeTag, class MyTypeTag>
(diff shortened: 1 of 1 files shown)


opm/models/blackoil/blackoilnewtonmethod.hh
```diff
@@ -98,6 +102,18 @@ struct TemperatureMin<TypeTag, TTag::NewtonMethod>
...
+struct PressureMax<TypeTag, TTag::NewtonMethod>
+{
+    using type = GetPropType<TypeTag, Scalar>;
+    static constexpr type value = 1e99; //Kelvin
+};
+template<class TypeTag>
+struct PressureMin<TypeTag, TTag::NewtonMethod>
+{
+    using type = GetPropType<TypeTag, Scalar>;
+    static constexpr type value = -1e99; //Kelvin
```
GitPaean, line 114: `Kelvin` should be `Pascal`?

totto82:
jenkins build this please

opm/models/blackoil/blackoilnewtonmethod.hh
totto82: I understand the reasoning of setting PressureMin to -1e99 (i.e. default is no clamping), but negative pressure doesn't look right.  Better to use 0?

atgeirr: I think using 0 might lead to test failures and behaviour changes in unexpected places. I suggest merging as is, and then changing the default separately.

PR #863, opened 2024-01-16
avoid divition by zero for case without conductivity

totto82:
For numerical aquifers the thermal half transmissibiltes are zero which results in infinite update. We may think about adding the option for setting these from deck. As a first approach we fix the code to allow for zero values.

### DIFF
--- a/opm/models/blackoil/blackoilenergymodules.hh
+++ b/opm/models/blackoil/blackoilenergymodules.hh
@@ -567,13 +567,13 @@ public:
             exLambda = decay<Scalar>(exIq.totalThermalConductivity());
 
         Evaluation H;
-        if (inLambda > 0.0 && exLambda > 0.0) {
+        const Evaluation& inH = inLambda*inAlpha;
+        const Evaluation& exH = exLambda*outAlpha;
+        if (inH > 0 && exH > 0) {
             // compute the "thermal transmissibility". In contrast to the normal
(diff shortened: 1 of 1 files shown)


GitPaean:
good. I was about to create the same PR.

opm/models/blackoil/blackoilenergymodules.hh
```diff
@@ -567,7 +567,7 @@ public:
             exLambda = decay<Scalar>(exIq.totalThermalConductivity());
 
         Evaluation H;
-        if (inLambda > 0.0 && exLambda > 0.0) {
+        if (inAlpha*inLambda > 0 && outAlpha*exLambda > 0) {
```
GitPaean, line 570: I think it is good to calculate the following before the `if` condition.

totto82:
> It is not for only the numerical aquifers, it looks the problem the same with other NNCs.

That's right. We could think about extending NCC to allow for thermal conductivity.

totto82:
jenkins build this please

opm/models/blackoil/blackoilenergymodules.hh
totto82: Good idea. I will update.

GitPaean: Good. Please update then I think the PR is ready.

totto82:
jenkins build this please

PR #864, opened 2024-01-24
Construct the matrices in the AMG hierarchy with deterministic indices if possible.

lisajulia:
The function 'setUseFixedOrder' is called if it is defined and the matrices in the AMG hierarchy are constructed with deterministic indices. If its not defined yet, it is not called and the matrices in the AMG hierarchy are constructed with non-deterministic indices.

### DIFF
--- a/opm/simulators/linalg/parallelamgbackend.hh
+++ b/opm/simulators/linalg/parallelamgbackend.hh
@@ -255,6 +255,17 @@ protected:
     }
 #endif
 
+    template <typename C>
+    auto setUseFixedOrder(C criterion, bool booleanValue) -> decltype(criterion.setUseFixedOrder(booleanValue))
+    {
+        return criterion.setUseFixedOrder(booleanValue); // Set flag to ensure that the matrices in the AMG hierarchy are constructed with deterministic indices.
+    }
+    template <typename C>
+    void setUseFixedOrder(C, ...)
+    {
+        // do nothing, since the function setUseFixedOrder does not exist yet
+    }
+
     void setupAmg_()
     {
         if (amg_)
@@ -291,6 +302,7 @@ protected:
         // coarsenCriterion.setAccumulate(Dune::Amg::noAccu);
         coarsenCriterion.setAccumulate(Dune::Amg::atOnceAccu);
         coarsenCriterion.setSkipIsolated(false);
+        setUseFixedOrder(coarsenCriterion, true); // If possible, set flag to ensure that the matrices in the AMG hierarchy are constructed with deterministic indices.
 
 // instantiate the AMG preconditioner
 #if HAVE_MPI
--- a/opm/simulators/linalg/parallelamgbackend.hh
+++ b/opm/simulators/linalg/parallelamgbackend.hh
@@ -255,6 +255,7 @@ protected:
     }
 #endif
 
+    // trailing return type with decltype used for detecting existence of setUseFixedOrder member function by overloading the setUseFixedOrder function
     template <typename C>
     auto setUseFixedOrder(C criterion, bool booleanValue) -> decltype(criterion.setUseFixedOrder(booleanValue))
     {


opm/simulators/linalg/parallelamgbackend.hh
```diff
@@ -255,6 +255,17 @@ protected:
     }
 #endif
 
+    template <typename C>
+    auto setUseFixedOrder(C criterion, bool booleanValue) -> decltype(criterion.setUseFixedOrder(booleanValue))
```
atgeirr, line 259: What type is that return value, when defined? Maybe this function should return that, i.e.
```
    return criterion.setUseFixedOrder(booleanValue);
```
to avoid compiler warnings.

blattms: I guess it is void. If there is no return then the result of the last operation is returned. That could raise a warning.

So I guess you have a point and there should either be a return statement or we simply use void.

@lisajulia You are a modern c++ expert :muscle: and I can learn from you :sunglasses:

blattms:
jenkins build this please

opm/simulators/linalg/parallelamgbackend.hh
```diff
@@ -255,6 +255,17 @@ protected:
...
+    {
+        criterion.setUseFixedOrder(booleanValue); // Set flag to ensure that the matrices in the AMG hierarchy are constructed with deterministic indices.
+    }
+    template <typename C>
+    void setUseFixedOrder(C criterion, ...)
```
blattms, line 264: This triggers a warning with [jenkins..](https://ci.opm-project.org/job/opm-models-PR-builder/463/gcc/source.f0dbb1f7-d6f6-4485-be54-77a790478ab9/#264) Please remove the variable name:
```
void setUseFixedOrder(C, ...)
```

blattms:
@lisajulia Now that you have the power you can call jenkins yourself.

blattms (approved):
Looks great now.

opm/simulators/linalg/parallelamgbackend.hh
bska: Is the trailing return type for sfinae purposes&ndash;e.g., detecting existence of a `setUseFixedOrder()` member function?  If so, is there a reason we're not using the traditional detection idiom along the lines of
```C++
template <typename, class = void>
struct has_setUseFixedOrder : public std::false_type;
template <typename T>
struct has_setUseFixedOrder<
    T, std::void_t<std::declval<T>().setUseFixedOrder(std::declval<bool>())>
> : public std::true_type;
template <typename C>
decltype(auto) setUseFixedOrder(C criterion, bool booleanValue)
{
    return this->setUseFixedOrderImpl(C, booleanValue, typename has_setUseFixedOrder<C>::type{});
}
template <typename C>
decltype(auto) setUseFixedOrderImpl(C criterion, bool booleanValue, std::true_type)
{
    return C.setUseFixedCriterion(booleanValue);
}
template <typename C>
void setUseFixedOrderImpl(C, bool, std::false_type)
{}
```

lisajulia: @bska: Yes, that's why. My approach is a little bit shorter, of course the one you suggest works as well - if you want me to update it, let me know! Thanks!

blattms: Probably me, but apart from being shorter @lisajulia's version is easier to understand. Is there a special reason (compiler support) for proposing the other one?

blattms:
I have one request. For the next version of release notes and manual, it would be good if there is a clear description what is accomplished by this PR. Maybe add something the the convergence rates and/or number of linear iterations differed for identical runs before and now these are reproducable with newer DUNE versions?

lisajulia:
jenkins build this opm-simulators=5111 please

opm/simulators/linalg/parallelamgbackend.hh
bska: > apart from being shorter @lisajulia's version is easier to understand.

Maybe.  The `void_t` approach has the benefit of being the conventional way of implementing the detection idiom whereas the trailing return type does not immediately draw attention to its overload resolution purpose.  It is not impossible that someone would try to replace the trailing return type with `decltype(auto)` while maintaining the code and then everything would break.

lisajulia: @bska: Yes, is that ok? https://github.com/OPM/opm-models/pull/864/commits/ac5853303c9967fb8205984354af0b0c0bd5abdd

lisajulia:
@bska should we run a benchmark here as well and then you can merge this PR?

akva2:
this code isn't used in any of the benchmarks.

blattms:
What benchmark will do is to run the binary flow from opm-simulators with a view models and compare performance to previous runs.

Thanks a lot for the changes. I am merging them now.

blattms:
@akva2 If I now broke the post-build then I owe you  a beer. Should have run checked whether jenkins was run...

PR #868, opened 2024-01-26
Add mass fraction diffusion coefficients

totto82:
Depends on https://github.com/OPM/opm-common/pull/3878

PR #869, opened 2024-01-26
Enables diffusion of water in gas phase

totto82:
Is for convenience build on #868


totto82:
jenkins build this please

PR #870, opened 2024-02-01
Only output warning if dispersion is activated

totto82:


totto82:
jenkins build this please

PR #873, opened 2024-02-06
[bugfix] Do not try to save FLOWS or FLOWRES values for NNCs.

blattms:
If we would try that then we would index vectors with index -1 and possibly overwrite other system information. This was the case for some models resulting in cryptic errror " corrupted size vs. prev_size"

Now we simply do nothing for NNCs.


blattms:
jenkins build this please

blattms:
Of course our manual says that at least FLORES includes flow along NNCs. This was maybe true be chance before if we were lucky. It definitely is not true with this patch.

I still think we should merge this to prevent crashes. We might need to implement proper handling of NNCs, though.

daavid00:
If I remember well, this used to work for NNCs (unfortunately we did not add any tests). In the applications I have been using FLOWS/FLORES lately, I do not have NNCs, so I have not face this use. If this currently leads to crashes, then I also think this should be merge and proper handle the NNCs.

atgeirr:
I agree with the above, avoid the crash now, fix NNCs later.

PR #874, opened 2024-02-09
Fix Flows/Flores for NNCs

totto82:
Restores NNC output for Flows. 

See #873

### DIFF
--- a/opm/models/discretization/common/tpfalinearizer.hh
+++ b/opm/models/discretization/common/tpfalinearizer.hh
@@ -496,7 +496,7 @@ private:
                                                      bf.area(),
                                                      bf.integrationPos()[dimWorld - 1],
                                                      exFluidState};
-                        boundaryInfo_.push_back({myIdx, dir_id, bcdata});
+                        boundaryInfo_.push_back({myIdx, dir_id, bfIndex, bcdata});
                     }
                 }
             }
@@ -663,11 +663,9 @@ public:
             // Flux term.
             {
             OPM_TIMEBLOCK_LOCAL(fluxCalculationForEachCell);
+            short loc = 0;
             for (const auto& nbInfo : nbInfos) {
                 OPM_TIMEBLOCK_LOCAL(fluxCalculationForEachFace);
-                // not for NNCs
-                if (nbInfo.res_nbinfo.dirId < 0)
-                    continue;
                 unsigned globJ = nbInfo.neighbor;
                 assert(globJ != globI);
                 adres = 0.0;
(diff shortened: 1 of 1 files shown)


totto82:
jenkins build this please

opm/models/discretization/common/tpfalinearizer.hh
```diff
@@ -673,30 +674,35 @@ public:
...
             const IntensiveQuantities& insideIntQuants = model_().intensiveQuantities(globI, /*timeIdx*/ 0);
             LocalResidual::computeBoundaryFlux(adres, problem_(), bdyInfo.bcdata, insideIntQuants, globI);
             adres *= bdyInfo.bcdata.faceArea;
-
+            const unsigned bfIndex = bdyInfo.bfIndex;
             if (enableFlows) {
-                for (unsigned phaseIdx = 0; phaseIdx < numEq; ++ phaseIdx) {
-                    flowsInfo_[globI][bdyInfo.dir].flow[phaseIdx] = adres[phaseIdx].value();
+                for (unsigned eqIdx = 0; eqIdx < numEq; ++ eqIdx) {
+                    flowsInfo_[globI][nbInfos.size() + bfIndex].flow[eqIdx] = adres[eqIdx].value();
```
atgeirr, line 705: To check I understood this correctly: this assumes that for every cell, flowsInfo_ stores a number of objects equal to the number of neighbours plus the number of boundaries, and the flows for boundaries come last.

atgeirr (commented):
Looks mostly fine, a few minor issues to address before merging.

opm/models/discretization/common/tpfalinearizer.hh
totto82: Yes. That is correct. That is also why I only resize to (numCells, 6) i.e guessing on 6 neighbors. (your comment above)

atgeirr: It is a reserve not a resize, but apart from that I still do not understand. A typical grid with ~1000 cells will have ~6000 neighbours, so reserving just space for 6 in the data array will ensure that the data array gets reallocated many times. How does this relate to the boundaries being added as well? If anything, that would increase the required space.

opm/models/discretization/common/tpfalinearizer.hh
```diff
@@ -565,13 +565,13 @@ private:
         }
 
         if (anyFlows) {
-            flowsInfo_.reserve(numCells, 6 * numCells);
+            flowsInfo_.reserve(numCells, 6);
```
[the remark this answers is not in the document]
totto82: Thanks. I misread the SparseTable class. Will revert.

totto82:
jenkins build this please

totto82:
jenkins build this please

PR #881, opened 2024-03-21
changing including guard name for two diffusionmodule

GitPaean:
they used to use the same EWOMS_DIFFUSION_MODULE_HH

It looks like `#pragma once` might be better when coming to this.


akva2 (approved):
Good catch, merge when green.

bska (approved):

PR #883, opened 2024-04-03
Remove unused EWOMS_RESET_PARAMS macro

akva2:
This macro is unused, it adds no convenience in any case and less macros is always good.


bska (approved):
Very good.  I'll merge into master.

PR #884, opened 2024-04-03
remove EWOMS_END_PARAM_REGISTRATION macro

akva2:
it adds no simplicity and only obfuscates

### DIFF
--- a/opm/models/utils/parametersystem.hh
+++ b/opm/models/utils/parametersystem.hh
@@ -87,18 +87,6 @@
 /*!
  * \ingroup Parameter
  *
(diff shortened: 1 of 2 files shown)


akva2:
jenkins build this opm-simulators=5269 please

blattms:
Using the macro seems to indicate that we can now safely assume that all parameters are registered. E.g. to be sure that --help will print everything.

By removing it we cannot assume that at any stage. Might not really be a problem, though.

akva2:
That would be true if i did not replace the macro usage with the direct call instead....

bska (approved):
Thanks.  This looks good to me and I'll merge into master.

PR #889, opened 2024-04-04
fixed: inconsistent parameter setups

akva2:
registered as int and obtained as unsigned, or registered as Scalar and obtained as int


akva2:
jenkins build this please

bska (commented):
This PR chooses to register the parameters as the retrieved type instead of retrieving the parameters as the registered type.  Is there a particular reason for this choice or are both choices equally valid from a technical point of view?

akva2:
Either would be fine apart from the Scalar one that obviously is a cut and paste issue.

bska (approved):
> My reason for [registering the retrieved type is that] that will be the effect of #888 as in this is the way to not change the user code.

Okay, very good.  The problem of a user passing a negative value as the argument to a parameter of type `unsigned int` is (probably) the same before and after this change so I'll merge into master.

akva2:
for the record; yes, when the cast happens differ but result it exactly the same.

PR #891, opened 2024-04-05
Parameters::hideParam

akva2:
pass property struct instead of name of stringified parameter.


akva2:
jenkins build this opm-simulators=5275 please

PR #894, opened 2024-04-15
Parameters::isSet: pass parameter struct instead of stringified

akva2:
Let's get rid of the final method that is string based.

### DIFF
--- a/opm/models/utils/parametersystem.hh
+++ b/opm/models/utils/parametersystem.hh
@@ -880,15 +880,18 @@ public:
      *
      * If the parameter in question has not been registered, this throws an exception.
      */
-    template <class ParamType>
-    static bool isSet(const char* paramName,
-                      bool errorIfNotRegistered = true)
+    template <template<class,class> class Param>
+    static bool isSet(bool errorIfNotRegistered = true)
     {
-
+        const std::string paramName = getPropName<TypeTag,Param>();
 #ifndef NDEBUG
         // make sure that the parameter is used consistently. since
         // this is potentially quite expensive, it is only done if
         // debugging code is not explicitly turned off.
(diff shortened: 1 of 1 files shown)


akva2:
jenkins build this opm-simulators=5282 please

opm/models/utils/parametersystem.hh
```diff
@@ -880,15 +880,18 @@ public:
...
      * If the parameter in question has not been registered, this throws an exception.
      */
-    template <class ParamType>
-    static bool isSet(const char* paramName,
-                      bool errorIfNotRegistered = true)
+    template <template<class,class> class Param>
+    static bool isSet(bool errorIfNotRegistered = true)
     {
-
+        const std::string paramName = getPropName<TypeTag,Param>();
```
bska, line 886: If you're explicitly storing this in a `std::string`, then I suppose we don't need [`canonicalName`](https://github.com/OPM/opm-models/blob/769bb55c688eb09481295d32dbce6b79b9eaaa25/opm/models/utils/parametersystem.hh#L905) anymore and might as well pass `paramName` directly as an argument to [`hasKey()`](https://github.com/OPM/opm-models/blob/769bb55c688eb09481295d32dbce6b79b9eaaa25/opm/models/utils/parametersystem.hh#L908).

bska (approved):
This looks good to me and I'll be happy to merge it in its current form, but I think there's one more simplification we can make here.

opm/models/utils/parametersystem.hh
akva2: indeed.

akva2:
jenkins build this opm-simulators=5282 please

bska (approved):
Thanks a lot for the updates.  This looks good to me and I'll merge into master.

PR #895, opened 2024-04-15
ParameterSystem: some cleanup

akva2:
- remove check_  - no way to access parameters using wrong type any more
- remove Param<TypeTag> wrapper
- replace iterator loops with range based loops

### DIFF
--- a/opm/models/utils/parametersystem.hh
+++ b/opm/models/utils/parametersystem.hh
@@ -352,23 +352,15 @@ inline void getFlattenedKeyList_(std::list<std::string>& dest,
                                  const std::string& prefix)
 {
     // add the keys of the current sub-structure
-    auto keyIt = tree.getValueKeys().begin();
-    const auto& keyEndIt = tree.getValueKeys().end();
-    for (; keyIt != keyEndIt; ++keyIt) {
-        std::string newKey(prefix);
-        newKey += *keyIt;
+    for (const auto& valueKey : tree.getValueKeys()) {
+        std::string newKey(prefix + valueKey);
         dest.push_back(newKey);
     }
 
     // recursively add all substructure keys
-    auto subStructIt = tree.getSubKeys().begin();
-    const auto& subStructEndIt = tree.getSubKeys().end();
-    for (; subStructIt != subStructEndIt; ++subStructIt) {
-        std::string newPrefix(prefix);
-        newPrefix += *subStructIt;
-        newPrefix += ".";
-
-        getFlattenedKeyList_(dest, tree.sub(*subStructIt), newPrefix);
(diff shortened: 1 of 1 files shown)


akva2:
jenkins build this please

opm/models/utils/parametersystem.hh
```diff
@@ -355,23 +352,19 @@ inline void getFlattenedKeyList_(std::list<std::string>& dest,
...
 {
     // add the keys of the current sub-structure
-    auto keyIt = tree.getValueKeys().begin();
-    const auto& keyEndIt = tree.getValueKeys().end();
-    for (; keyIt != keyEndIt; ++keyIt) {
+    for (const auto& valueKey : tree.getValueKeys()) {
         std::string newKey(prefix);
-        newKey += *keyIt;
+        newKey += valueKey;
         dest.push_back(newKey);
```
bska, line 358: I suppose this could now be shortened to
```C++
dest.emplace_back(prefix + valueKey);
```

opm/models/utils/parametersystem.hh
```diff
@@ -355,23 +352,19 @@ inline void getFlattenedKeyList_(std::list<std::string>& dest,
...
 
     // recursively add all substructure keys
-    auto subStructIt = tree.getSubKeys().begin();
-    const auto& subStructEndIt = tree.getSubKeys().end();
-    for (; subStructIt != subStructEndIt; ++subStructIt) {
+    for (const auto& subKey : tree.getSubKeys()) {
         std::string newPrefix(prefix);
-        newPrefix += *subStructIt;
+        newPrefix += subKey;
         newPrefix += ".";
```
bska, line 365: This could, I think, be shortened to
```C++
const std::string newPrefix = prefix + subKey + '.';
```

bska (commented):
There are a couple of instances of
```C++
const std::string paramName = ...;
// ...
std::string canonicalName{paramName};
use(canonicalName);
```
here which I take it we've just inherited from earlier versions.  Now might be a good time to simplifiy those as well.  Other than that, I think we can replace multi-stage string construction with operator `+=` with single constructor calls.

opm/models/utils/parametersystem.hh
akva2: done.

opm/models/utils/parametersystem.hh
akva2: done.

akva2:
and fixed the copy-paste for canonicalName - commits made before previous review.

akva2:
jenkins build this please

bska (approved):
This looks good to me now and I appreciate the simplification.  I'll merge into master.

PR #898, opened 2024-04-26
Avoid call to FaceDir::FromIntersectionIndex() in hot path.

atgeirr:
Call is expensive, might also be good to improve the function in opm-common. (FaceDir.cpp) I leave that to others though.

This fixes the significant preformance regression in flux calculations that was introduced earlier. On the order of 5% total at least.


atgeirr:
Not directly related to OPM/opm-simulators#5318, but I will test both together.

PR #903, opened 2024-07-01
Fix runtime parameters used as properties

akva2:
It makes no sense to use run-time parameters as properties, it only obtains the default values, not possible runtime values.


akva2:
jenkins build this please

bska (approved):
Looks good to me.  Please merge when green.

Issue #871, opened 2024-02-01
Improvement for tasklets.hh


lisajulia:
See: https://github.com/OPM/opm-common/pull/3870#issue-2098549925
When a faulty file is detected, the program still continues because the error is caught here: https://github.com/OPM/opm-models/blob/a89ccecaa38aa552873a7a90f50902eb7327abd1/opm/models/parallel/tasklets.hh#L234

Suggestion for improvement: https://github.com/OPM/opm-common/pull/3870#issuecomment-1911661089

PR #909, opened 2024-07-08
Feature/add failure flag to tasklets

lisajulia:
As suggested by https://github.com/OPM/opm-common/pull/3870#issuecomment-1911661089, adding a failureFlag to the TaskletRunner, will solve https://github.com/OPM/opm-models/issues/871

### DIFF
--- a/opm/models/parallel/tasklets.hh
+++ b/opm/models/parallel/tasklets.hh
@@ -27,6 +27,7 @@
 #ifndef EWOMS_TASKLETS_HH
 #define EWOMS_TASKLETS_HH
 
+#include <atomic>
 #include <stdexcept>
 #include <cassert>
 #include <thread>
@@ -223,6 +224,11 @@ public:
      */
     void dispatch(std::shared_ptr<TaskletInterface> tasklet)
     {
+        if (failureFlag_.load(std::memory_order_relaxed)) {
+            std::cerr << "Failure flag of the TaskletRunner is set. Not dispatching new tasklets.\n";
+            exit(EXIT_FAILURE);
+        }
+
         if (threads_.empty()) {
             // run the tasklet immediately in synchronous mode.
             while (tasklet->referenceCount() > 0) {
@@ -232,9 +238,11 @@ public:
                 }
                 catch (const std::exception& e) {
                     std::cerr << "ERROR: Uncaught std::exception when running tasklet: " << e.what() << ". Trying to continue.\n";
+                    failureFlag_.store(true, std::memory_order_relaxed);
                 }
                 catch (...) {
                     std::cerr << "ERROR: Uncaught exception (general type) when running tasklet. Trying to continue.\n";
+                    failureFlag_.store(true, std::memory_order_relaxed);
                 }
             }
         }
@@ -280,6 +288,16 @@ public:
 
         barrierTasklet->wait();
     }
+private:
+    // Atomic flag that is set to failure if any of the tasklets run by the TaskletRunner fails.
+    // This flag is checked before new tasklets run or get dispatched and in case it is true, the
+    // thread execution will be stopped / no new tasklets will be started and the program will abort.
+    // To set the flag and load the flag, we use std::memory_order_relaxed.
+    // Atomic operations tagged memory_order_relaxed are not synchronization operations; they do not
+    // impose an order among concurrent memory accesses. They guarantee atomicity and modification order
+    // consistency. This is the right choice for the setting here, since it is enough to broadcast failure
+    // before new run or get dispatched.
+    std::atomic<bool> failureFlag_ = false;
 
 protected:
     // main function of the worker thread
@@ -295,6 +313,12 @@ protected:
     void run_()
     {
         while (true) {
+            // Check if failure flag is set
+            if (failureFlag_.load(std::memory_order_relaxed)) {
+                std::cerr << "Failure flag of the TaskletRunner is set. Exiting thread.\n";
+                exit(EXIT_FAILURE);
+            }
+
             // wait until tasklets have been pushed to the queue. first we need to lock
             // mutex for access to taskletQueue_
             std::unique_lock<std::mutex> lock(taskletQueueMutex_);
@@ -330,10 +354,12 @@ protected:
                 tasklet->run();
             }
             catch (const std::exception& e) {
-                std::cerr << "ERROR: Uncaught std::exception when running tasklet: " << e.what() << ". Trying to continue.\n";
+                std::cerr << "ERROR: Uncaught std::exception when running tasklet: " << e.what() << ".\n";
+                failureFlag_.store(true, std::memory_order_relaxed); 
             }
             catch (...) {
-                std::cerr << "ERROR: Uncaught exception when running tasklet. Trying to continue.\n";
+                std::cerr << "ERROR: Uncaught exception when running tasklet.\n";
+                failureFlag_.store(true, std::memory_order_relaxed);
             }
         }
     }
--- a/opm/models/parallel/tasklets.hh
+++ b/opm/models/parallel/tasklets.hh
@@ -199,6 +199,11 @@ public:
         }
     }
 
+    bool failure() const
+    {
+        return this->failureFlag_.load(std::memory_order_relaxed);
+    }
+
     /*!
      * \brief Returns the index of the current worker thread.
      *
@@ -224,11 +229,6 @@ public:
      */
     void dispatch(std::shared_ptr<TaskletInterface> tasklet)
     {
-        if (failureFlag_.load(std::memory_order_relaxed)) {
-            std::cerr << "Failure flag of the TaskletRunner is set. Not dispatching new tasklets.\n";
-            exit(EXIT_FAILURE);
-        }
-
         if (threads_.empty()) {
             // run the tasklet immediately in synchronous mode.
             while (tasklet->referenceCount() > 0) {
@@ -313,11 +313,6 @@ protected:
     void run_()
     {
         while (true) {
-            // Check if failure flag is set
-            if (failureFlag_.load(std::memory_order_relaxed)) {
-                std::cerr << "Failure flag of the TaskletRunner is set. Exiting thread.\n";
-                exit(EXIT_FAILURE);
-            }
 
             // wait until tasklets have been pushed to the queue. first we need to lock
             // mutex for access to taskletQueue_
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -323,6 +323,8 @@ opm_add_test(tutorial1
 
 opm_add_test(test_tasklets
              DRIVER_ARGS --plain)
+opm_add_test(test_tasklets_failure
+             DRIVER_ARGS --plain)
 
 opm_add_test(test_mpiutil
              PROCESSORS 4
--- a/tests/test_tasklets.cc
+++ b/tests/test_tasklets.cc
@@ -35,7 +35,8 @@
 
 std::mutex outputMutex;
 
-Opm::TaskletRunner *runner;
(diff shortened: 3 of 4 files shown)


lisajulia:
jenkins build this please

lisajulia:
jenkins build this please

opm/models/parallel/tasklets.hh
```diff
@@ -223,6 +224,11 @@ public:
      */
     void dispatch(std::shared_ptr<TaskletInterface> tasklet)
     {
+        if (failureFlag_.load(std::memory_order_relaxed)) {
+            std::cerr << "Failure flag of the TaskletRunner is set. Not dispatching new tasklets.\n";
+            exit(EXIT_FAILURE);
```
blattms, line 229: This is much better than continuing.
Still, it rather ungraceful behavior if there is an `exit(EXIT_FAILURE)` statement executed on one process. All processes might print an error MPI error message about another process that died. It might be hard to find the real error message.

Do you think it is possible to throw here and somewhere up the stack make sure that all processes stop more graceful?

lisajulia:
jenkins build this please

opm/models/parallel/tasklets.hh
blattms, line 229: This is much better than continuing.
Still, it might result in rather ungraceful behavior if there is an exit(EXIT_FAILURE) statement executed on one process. All processes might print an MPI error message about another process that died. It might be hard to find the real error message.

Also: Should we refactor to use OpmLog while at it?

Do you think it is possible to throw here and somewhere up the stack make sure that all processes stop more graceful?

lisajulia: I'll check that

lisajulia: Both are possible:

- "exit(EXIT_FAILURE)" version:
```
 Newton its= 3, linearizations= 4 (0.0sec), linear its= 16 (0.1sec)
ERROR: Uncaught std::exception when running tasklet: Can not open EclFile: ../../opm-data/spe1/SPE1CASE1.UNRST.
Failure flag of the TaskletRunner is set. Exiting thread.
--------------------------------------------------------------------------
Primary job  terminated normally, but 1 process returned
a non-zero exit code. Per user-direction, the job has been aborted.
--------------------------------------------------------------------------
--------------------------------------------------------------------------
mpirun detected that one or more processes exited with non-zero status, thus causing
the job to be terminated. The first process to do so was:

  Process name: [[28130,1],0]
  Exit code:    1
--------------------------------------------------------------------------
```

- thow-version:
```
ERROR: Uncaught std::exception when running tasklet: Can not open EclFile: ../../opm-data/spe1/SPE1CASE1.UNRST.
Failure flag of the TaskletRunner is set. Exiting thread.
terminate called after throwing an instance of 'std::runtime_error'
  what():  Failure flag of the TaskletRunner is set. Not dispatching new tasklets.
[samwise:101326] *** Process received signal ***
[samwise:101326] Signal: Aborted (6)
[samwise:101326] Signal code:  (-6)
[samwise:101326] [ 0] /lib/x86_64-linux-gnu/libc.so.6(+0x3c050)[0x7fb74025b050]
[samwise:101326] [ 1] /lib/x86_64-linux-gnu/libc.so.6(+0x8ae2c)[0x7fb7402a9e2c]
[samwise:101326] [ 2] /lib/x86_64-linux-gnu/libc.so.6(gsignal+0x12)[0x7fb74025afb2]
[samwise:101326] [ 3] /lib/x86_64-linux-gnu/libc.so.6(abort+0xd3)[0x7fb740245472]
[samwise:101326] [ 4] /lib/x86_64-linux-gnu/libstdc++.so.6(+0x9d919)[0x7fb74009d919]
[samwise:101326] [ 5] /lib/x86_64-linux-gnu/libstdc++.so.6(+0xa8e1a)[0x7fb7400a8e1a]
[samwise:101326] [ 6] /lib/x86_64-linux-gnu/libstdc++.so.6(+0xa8e85)[0x7fb7400a8e85]
[samwise:101326] [ 7] /lib/x86_64-linux-gnu/libstdc++.so.6(+0xa90d8)[0x7fb7400a90d8]
[samwise:101326] [ 8] ./bin/flow(+0x4169a0)[0x55c9aca949a0]
[samwise:101326] [ 9] ./bin/flow(+0x41689e)[0x55c9aca9489e]
[samwise:101326] [10] ./bin/flow(+0x7422da)[0x55c9acdc02da]
[samwise:101326] [11] ./bin/flow(+0x73add9)[0x55c9acdb8dd9]
[samwise:101326] [12] ./bin/flow(+0x730c61)[0x55c9acdaec61]
[samwise:101326] [13] ./bin/flow(+0x722f12)[0x55c9acda0f12]
[samwise:101326] [14] ./bin/flow(+0x710836)[0x55c9acd8e836]
[samwise:101326] [15] /lib/x86_64-linux-gnu/libstdc++.so.6(+0xd44a3)[0x7fb7400d44a3]
[samwise:101326] [16] /lib/x86_64-linux-gnu/libc.so.6(+0x89134)[0x7fb7402a8134]
[samwise:101326] [17] /lib/x86_64-linux-gnu/libc.so.6(+0x1097dc)[0x7fb7403287dc]
[samwise:101326] *** End of error message ***
--------------------------------------------------------------------------
Primary job  terminated normally, but 1 process returned
a non-zero exit code. Per user-direction, the job has been aborted.
--------------------------------------------------------------------------
--------------------------------------------------------------------------
mpirun noticed that process rank 0 with PID 0 on node samwise exited on signal 6 (Aborted).
--------------------------------------------------------------------------
```

blattms:
Interesting failure for python_basic:
```
Error: [.../opm-common/opm/io/eclipse/EclFile.cpp:245] Could not open file: './SPE1CASE1.UNRST'
ERROR: Uncaught std::exception when running tasklet: [.../opm-common/opm/io/eclipse/EclFile.cpp:245] Could not open file: './SPE1CASE1.UNRST'.
```

Probably this appeared previously and was not noticed. Question is whether this is a fatal failure or something that should be ignored.

Once that is sorted out, I'll merge.

opm/models/parallel/tasklets.hh
lisajulia: Either option is fine for me, let me know what you prefer

lisajulia:
> Interesting failure for python_basic:
> 
> ```
> Error: [.../opm-common/opm/io/eclipse/EclFile.cpp:245] Could not open file: './SPE1CASE1.UNRST'
> ERROR: Uncaught std::exception when running tasklet: [.../opm-common/opm/io/eclipse/EclFile.cpp:245] Could not open file: './SPE1CASE1.UNRST'.
> ```
> 
> Probably this appeared previously and was not noticed. Question is whether this is a fatal failure or something that should be ignored.
> 
> Once that is sorted out, I'll merge.

Yes! Indeed! I'll look into that!

lisajulia:
jenkins build this opm-simulators=5474 please

lisajulia:
jenkins build this opm-simulators=5474 please

lisajulia:
@bska and @blattms: Can you have a look a this? Thanks! :)

bska (commented):
I like the `atomic<bool>` although if we're looking **very** narrowly at the use case in [opm-simulators](https://github.com/OPM/opm-simulators/blob/93e06d218745e1c63e5948f5615b275ff1f8ee9e/opm/simulators/flow/EclGenericWriter_impl.hpp#L602-L616) I don't think actually need it since there's only ever one thread/request active at any one time.  We should keep it nonetheless.

We might however consider a more polling-based interface.  In particular, I would prefer it if the `TaskletRunner` did not call `std::exit()` on the simulator's behalf.  If we had a predicate along the lines of
```C++
bool TaskletRunner::failure() const { return failureFlag_.load(...); }
```
then the client code in `EclGenericWriter::doWriteOutput()` might be rewritten as
```C++
this->taskletRunner_->barrier();
if (this->taskletRunner_->failure()) {
    throw SomeException { ... };
}

auto tasklet = std::make_shared<EclWriteTasklet>(...);
this->taskletRunner_->dispatch(std::move(tasklet));
```
Then we'd guard the calling code in [`EclWriter::writeOutput()`](https://github.com/OPM/opm-simulators/blob/93e06d218745e1c63e5948f5615b275ff1f8ee9e/opm/simulators/flow/EclWriter.hpp#L459-L480)&ndash;a function which is run on all ranks&ndash;by the [`OPM_BEGIN_PARALLEL_TRY_CATCH()`](https://github.com/OPM/opm-simulators/blob/93e06d218745e1c63e5948f5615b275ff1f8ee9e/opm/simulators/utils/DeferredLoggingErrorHelpers.hpp#L158) and [`OPM_END_PARALLEL_TRY_CATCH()`](https://github.com/OPM/opm-simulators/blob/93e06d218745e1c63e5948f5615b275ff1f8ee9e/opm/simulators/utils/DeferredLoggingErrorHelpers.hpp#L192) macros from `DeferredLoggingErrorHelpers.hpp`.  I.e., we'd have something along the lines of
```C++
OPM_BEGIN_PARALLEL_TRY_CATCH()
if (this->collectOnIORank_.isIORank()) { ... }
OPM_END_PARALLEL_TRY_CATCH("File output failure: ", simulator_.gridView().comm())
```

Finally, I'll second @blattms' comment that we should try to move away from `<iostream>` based logging here, but that could/should be follow-up work.

lisajulia:
jenkins build this please

lisajulia:
@bska: Can you have another look here?

bska (commented):
I might personally consider using one of the STL lock guards like `std::unique_lock` or `std::lock_guard` instead of directly calling the `std::mutex::lock()` and `std::mutex::unlock()` member functions, but in such simple test code it's probably a wash.  On the other hand, I don't quite understand why we're using an owning raw pointer for the `TaskletRunner` object in the new unit test.  Would you care to elaborate on the reason for this?

lisajulia:
> I might personally consider using one of the STL lock guards like `std::unique_lock` or `std::lock_guard` instead of directly calling the `std::mutex::lock()` and `std::mutex::unlock()` member functions, but in such simple test code it's probably a wash. On the other hand, I don't quite understand why we're using an owning raw pointer for the `TaskletRunner` object in the new unit test. Would you care to elaborate on the reason for this?

Thanks, I've added comments here: https://github.com/OPM/opm-models/pull/909/commits/cedd34a379c211ab8a2e45577436e3e62e796e10, is that ok now? Otherwise I can also remove the assertion and outputs.

bska (commented):
This is definitely an improvement, but I guess I still don't quite understand why the `runner` object needs to be a raw pointer instead of a smart pointer like `std::unique_ptr<>`.  Does the following not work?
```C++
std::unique_ptr<Opm::TaskletRunner> runner{};
// ...
void execute()
{
    const int numWorkers = 2;
    runner = std::make_unique<Opm::TaskletRunner>(numWorkers);
    BOOST_REQUIRE_LT(runner->workerThreadIndex(), 0);
    BOOST_REQUIRE_EQ(runner->numWorkerThreads(), numWorkers);
    // ...
}

BOOST_AUTO_TEST_SUITE(Tasklets)
BOOST_AUTO_TEST_CASE(TASKLET_FAILURE)
{
    // As before...
}
BOOST_AUTO_TEST_SUITE_END()
```

lisajulia:
> ```c++
>     runner = std::make_unique<Opm::TaskletRunner>(numWorkers);
> ```

Yes, thanks! I've changed that now in https://github.com/OPM/opm-models/pull/909/commits/f1b4182ef636cfd19242da9227613ae2a995e538

lisajulia:
jenkins build this please

bska (approved):
Thanks a lot for the updates.  This looks good to me now and I'll merge into master.

PR #918, opened 2024-08-07
changed: remove the print-properties parameter

akva2:
property introspection is not supported and unlikely to return.

this has not been supported since the property/typetag system was changed from the old macro based implementation, ie, for the last 5 years. since the new system is based on partial specialization of structs, this is unlikely to ever return since we'd need reflection support to implement it.


akva2:
jenkins build this opm-simulators=5514 please

bska (approved):
> since the new system is based on partial specialization of structs, this is unlikely to ever return since we'd need reflection support to implement it.

I'm still hopeful we'll have reflection in C++26 :wink: 

In any case, I agree that we can't currently support this feature so it's better to not create the impression that we do.  I'll merge into master.

PR #919, opened 2024-08-07
Use constexpr in blackoil indices classes

akva2:
I am aware this is functionally equivalent, but it's nice to be explicit about the compile time availability of these values.


akva2:
jenkins build this please

bska (approved):
Very good!  I like this a lot and I agree with the sentiment.  I'll merge into master.

PR #923, opened 2024-08-13
Move more parameters to typetag-free parameter system

akva2:
Next batch. This needs to be merged in concert with the downstream.


bska (approved):
Very good.  I'll merge this and its downstream counterpart into their respective master branches.

PR #922, opened 2024-08-12
Add workaround for libc++ incomplete from_chars().

atgeirr:
Float version not included in current version of libc++(15).

PR #926, opened 2024-08-15
Don't Require Floating Point from_chars() Function

bska:
This PR broadens the scope of commit 2ad332e0b6 (PR https://github.com/OPM/opm-models/pull/922) to apply to all compilers/libraries, not just Clang/libc++, which do not have support for floating-point types in `std::from_chars()`. While hopefully a transient situation, this enables building the parameter system with GCC versions prior to GCC 11.  We expect to require version 11 in the not too distant future, though.  At that point we should revert this commit.

We use a configure-time feature test of the compiler (CMake command `try_compile`) to detect whether or not the compiler supports floating-point overloads of `std::from_chars()` and emit the result to `config.h` as the new preprocessor symbol
```
HAVE_FLOATING_POINT_FROM_CHARS
```
We use `std::strtod()` as the fall-back alternative for floating point conversion if this symbol is defined to false (zero).


bska:
jenkins build this please

atgeirr:
I think this change removes use of from_chars for floats also from the platforms that support it.

bska:
> I think this change removes use of from_chars for floats also from the platforms that support it.

Yes, it absolutely does.  That's intentional, and part of why I said

> While hopefully a transient situation...

bska:
> > I think this change removes use of from_chars for floats also from the platforms that support it.
> 
> Yes, it absolutely does.

I decided to make this a little more user friendly and adaptive.  I've pushed an update which detects support for the floating-point version of `std::from_chars()` at configure time (using CMake's `try_compile()` command) and exports the result of that detection to `config.h` as a new preprocessor symbol,
```
HAVE_FLOATING_POINT_FROM_CHARS
```
We then have an alternative implementation in terms of `std::strtod()` if this symbol is false&ndash;i.e., zero.  On the other hand, when the symbol is true (one) we use `std::from_chars()` for both integer and floating-point conversions as originally implemented in PR #916.

I welcome additional suggestions for

bska:
jenkins build this please

akva2 (approved):
Looks good. I agree this is better than the other alternative, at least we have an ifdef to run into when we forget all about this again..

bska:
> Looks good

>I agree this is better than the other alternative, at least we have an ifdef to run into when we forget all about this again..

Yeah, that fact is what ultimately made me choose this latest approach.  We can always argue about naming (`HAVE_FROM_CHARS_FLOATING_POINT`, for instance) and whether or not the check should be in opm-common instead, but I think we can leave it here as is and move/improve it as the need arises.

CMakeLists.txt
```diff
@@ -79,6 +79,21 @@ macro (config_hook)
...
   opm_need_version_of("dune-istl")
   opm_need_version_of("dune-fem")
+
+  # The parameter system can leverage std::from_chars() for
+  # floating-point types if available.  Detect support for this
+  # feature.
+  try_compile(
+    have_float_from_chars
+    ${CMAKE_BINARY_DIR}
+    ${CMAKE_SOURCE_DIR}/cmake/test/testFloatFromChars.cpp
```
GitPaean, line 89: It appears not working with the super cmake file. It gets the directory wrong. 

```
Processing opm_defaults opm-models
-- Precompiled headers: disabled
-- Build type: Release
-- Could NOT find CppCheck (missing: CPPCHECK_PROGRAM) 
-- Disabling clang-check as CMAKE_EXPORT_COMPILE_COMMANDS is not enabled
-- Could NOT find dune-fem (missing: dune-fem_DIR)
-- Could NOT find dune-polygongrid (missing: dune-polygongrid_DIR)
-- Could NOT find dune-polygongrid (missing: dune-polygongrid_FOUND HAVE_DUNE_POLYGONGRID) 
-- Could NOT find PTScotch (missing: PTSCOTCH_INCLUDE_DIR SCOTCH_LIBRARY PTSCOTCH_LIBRARY PTSCOTCHERR_LIBRARY) 
-- Could NOT find METIS (missing: METIS_LIBRARY METIS_INCLUDE_DIRS) 
-- Could NOT find dune-fem (missing: dune-fem_FOUND HAVE_DUNE_FEM) 
-- Generating debug symbols: -ggdb3
-- Looking for strip utility
-- Looking for strip utility - found
-- Performing Test HAVE_DYNAMIC_BOOST_TEST
-- Performing Test HAVE_DYNAMIC_BOOST_TEST - Success
CMake Error at /home/kaib/OPM-master-test/build/CMakeFiles/CMakeTmp/CMakeLists.txt:17 (add_executable):
  Cannot find source file:

    /home/kaib/OPM-master-test/opm/cmake/test/testFloatFromChars.cpp


CMake Error at /home/kaib/OPM-master-test/build/CMakeFiles/CMakeTmp/CMakeLists.txt:17 (add_executable):
  No SOURCES given to target: cmTC_c69f3


CMake Error at opm-models/CMakeLists.txt:86 (try_compile):
  Failed to generate test project build system.
Call Stack (most recent call first):
  opm-common/cmake/Modules/OpmLibMain.cmake:131 (config_hook)
  opm-models/CMakeLists.txt:123 (include)


-- Configuring incomplete, errors occurred!
```

akva2: fixed in #927 which goes into master in about 1 min..

PR #929, opened 2024-09-02
add deprecation warning to README

akva2:
Ref https://github.com/OPM/opm-simulators/pull/4627


bska (approved):
OPM-Models has been merged into the OPM-Simulators repository. Merging this now to reflect that fact.

